PR #68, opened 2023-02-06
Bump toml_edit version

maciektr:


Cargo.toml
```diff
@@ -49,7 +49,8 @@ snapbox = { version = "0.4.3", features = ["cmd", "path"] }
 tempfile = "3.3.0"
 test-case = "2.2.2"
 thiserror = "1.0.37"
-toml_edit = { version = "0.17.1", features = ["easy", "serde"] }
+toml = "0.7.1"
+toml_edit = { version = "0.18.1", features = ["easy", "serde"] }
```
mkaput, line 53: I see that there's `toml_edit 0.19.2` available. Could we bump to the latest one? https://github.com/toml-rs/toml/blob/main/crates/toml_edit/CHANGELOG.md

mkaput, line 53: The `easy` feature is deprecated: https://github.com/toml-rs/toml/blob/main/crates/toml_edit/CHANGELOG.md#0180---2023-01-23. We should remove usages of it

scarb/tests/e2e/build.rs
```diff
@@ -225,8 +229,12 @@ fn compile_with_invalid_version() {
 error: failed to parse manifest at `[..]/Scarb.toml`
```
mkaput, line 229: Please use `indoc!` macro and its family, it will deal with newlines and indentation.

mkaput (changes requested):

scarb/tests/e2e/build.rs
```diff
@@ -193,14 +193,16 @@ fn compile_with_invalid_empty_name() {
...
+        .stdout_matches(indoc! {"\
+            error: failed to parse manifest at `[..]/Scarb.toml`
+
+            Caused by:
+                TOML parse error at line 3, column 20
+                  |
+                3 |             name = \"\"
+                  |                    ^^
+                empty string cannot be used as package name
+                \n",});
```
mkaput, line 205:
```suggestion
        .stdout_matches(indoc! {r#"
            error: failed to parse manifest at `[..]/Scarb.toml`

            Caused by:
                TOML parse error at line 3, column 20
                  |
                3 |             name = ""
                  |                    ^^
                empty string cannot be used as package name
        "#});
```

Apply this everywhere here. See how other tests do this: https://github.com/software-mansion/scarb/blob/dc08364119a5d19a132cddd566d9025d12e206d6/scarb/tests/e2e/build.rs#L65-L79

mkaput (changes requested):

mkaput (approved):

scarb/tests/e2e/build.rs
maciektr: Resolving after discussing offline.

PR #71, opened 2023-02-10
Pass scarb env vars

### LINKED ISSUE #8: Pass `PATH`, `*_LOG` and other vital envvars to subcommands
1. Write a test which ensures this behaviour (could be a unix-only test which runs a shell script which outputs these envs values)

### LINKED ISSUE #64: Remove trailing slash from path sources IDs Display representation
```
Compiling ci_testing v0.1.0 (D:\a\scarb\scarb\ci_testing\Scarb.toml\)
```

should look like this:

```
Compiling ci_testing v0.1.0 (D:\a\scarb\scarb\ci_testing\Scarb.toml)
```

this happens on all platforms

maciektr:
fix #8, fix #64

### DIFF
--- a/scarb/src/core/config.rs
+++ b/scarb/src/core/config.rs
@@ -27,6 +27,7 @@ pub struct Config {
     // HACK: This should be the lifetime of Config itself, but we cannot express that, so we
     //   put 'static here and transmute in getter function.
     package_cache_lock: OnceCell<AdvisoryLock<'static>>,
+    scarb_log: String,
     offline: bool,
 }
 
@@ -49,6 +50,8 @@ impl Config {
 
         let dirs = Arc::new(dirs);
 
+        let scarb_log = env::var("SCARB_LOG").unwrap_or_default();
+
         Ok(Self {
             manifest_path,
             dirs,
@@ -57,6 +60,7 @@ impl Config {
             ui,
             creation_time,
             package_cache_lock: OnceCell::new(),
+            scarb_log,
             offline: false,
         })
     }
@@ -69,6 +73,10 @@ impl Config {
         self.manifest_path()
             .parent()
             .expect("parent of manifest path must always exist")
+    }
+
+    pub fn scarb_log(&self) -> &str {
+        &self.scarb_log
     }
 
     pub fn dirs(&self) -> &AppDirs {
--- a/scarb/src/ops/subcommands.rs
+++ b/scarb/src/ops/subcommands.rs
@@ -24,6 +24,9 @@ pub fn execute_external_subcommand(cmd: &str, args: &[&OsStr], config: &Config) 
     cmd.args(args);
     cmd.env(SCARB_ENV, config.app_exe()?);
     cmd.env("PATH", config.dirs().path_env());
+    cmd.env("SCARB_CACHE", config.dirs().cache_dir.path_unchecked());
+    cmd.env("SCARB_CONFIG", config.dirs().config_dir.path_unchecked());
+    cmd.env("SCARB_MANIFEST_DIR", config.root());
     exec_replace(&mut cmd)
 }
 
--- a/scarb/src/ops/subcommands.rs
+++ b/scarb/src/ops/subcommands.rs
@@ -27,6 +27,7 @@ pub fn execute_external_subcommand(cmd: &str, args: &[&OsStr], config: &Config) 
     cmd.env("SCARB_CACHE", config.dirs().cache_dir.path_unchecked());
     cmd.env("SCARB_CONFIG", config.dirs().config_dir.path_unchecked());
     cmd.env("SCARB_MANIFEST_DIR", config.root());
+    cmd.env("SCARB_UI_VERBOSITY", config.ui().verbosity().to_string());
     exec_replace(&mut cmd)
 }
 
--- a/scarb/src/ops/subcommands.rs
+++ b/scarb/src/ops/subcommands.rs
@@ -28,6 +28,7 @@ pub fn execute_external_subcommand(cmd: &str, args: &[&OsStr], config: &Config) 
     cmd.env("SCARB_CONFIG", config.dirs().config_dir.path_unchecked());
     cmd.env("SCARB_MANIFEST_DIR", config.root());
     cmd.env("SCARB_UI_VERBOSITY", config.ui().verbosity().to_string());
+    cmd.env("SCARB_LOG", config.scarb_log());
     exec_replace(&mut cmd)
 }
 
--- a/scarb/src/ui/mod.rs
+++ b/scarb/src/ui/mod.rs
@@ -1,4 +1,8 @@
+use anyhow::{bail, Result};
 use clap::ValueEnum;
+use std::env;
+use std::fmt::Display;
+use std::str::FromStr;
 
 pub use machine::*;
 pub use message::*;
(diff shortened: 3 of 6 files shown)


scarb/src/core/config.rs
```diff
@@ -64,6 +64,10 @@ impl Config {
         &self.manifest_path
     }
 
+    pub fn manifest_dir(&self) -> &Utf8Path {
+        self.manifest_path.parent().unwrap()
+    }
```
mkaput, line 69: 1. This duplicates `root` :)
2. And this is ambiguous, because with `manifest_path` you are able to use non-standard-named manifest file, for example `Scarb.dev.toml`

mkaput (changes requested):
Passing only `verbosity` flag value to `SCARB_LOG` is imprecise, because `SCARB_LOG` is much more powerful, for example it allows granular filtering based on log event target ([docs](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html)).

I think what should be passed is a stringified representation of the EnvFilter that is built in [`main`](https://github.com/software-mansion/scarb/blob/abfb9ead54c78cbc6771d47ad08ebb24546e1dae/scarb/src/bin/scarb/main.rs#L23-L26).

Ideally, I would like to infer this from `tracing` rather than requiring passing to Config/UI, but I couldn't find a way to do so.
This representation should also be stored in Config rather than UI, because it's touching primarily logs.

mkaput:
> Passing only `verbosity` flag value to `SCARB_LOG` is imprecise, because `SCARB_LOG` is much more powerful, for example it allows granular filtering based on log event target ([docs](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html)).
> 
> 
> 
> I think what should be passed is a stringified representation of the EnvFilter that is built in [`main`](https://github.com/software-mansion/scarb/blob/abfb9ead54c78cbc6771d47ad08ebb24546e1dae/scarb/src/bin/scarb/main.rs#L23-L26).
> 
> 
> 
> Ideally, I would like to infer this from `tracing` rather than requiring passing to Config/UI, but I couldn't find a way to do so.

I was wrong here. Let's discuss this topic in the office next week

scarb/tests/e2e/subcommand.rs
```diff
@@ -40,6 +40,42 @@ fn subcommand() {
...
+            )
+            for v in "${required[@]}"; do
+                if test -z "${!v}"
+                then
+                    echo "Variable $v is not set!"
+                    exit 1
+                fi
+            done
+            "#
+        ),
```
mkaput, line 68: nitpick, people usually use `{}` with `indoc!`, see examples here: https://docs.rs/indoc/latest/indoc/

```suggestion
        indoc!{r#"
            #!/bin/bash
            required=(
                PATH
                SCARB
                SCARB_CACHE
                SCARB_CONFIG
                SCARB_LOG
                SCARB_MANIFEST_DIR
            )
            for v in "${required[@]}"; do
                if test -z "${!v}"
                then
                    echo "Variable $v is not set!"
                    exit 1
                fi
            done
        "#},
```

scarb/src/ops/subcommands.rs
```diff
@@ -24,6 +24,13 @@ pub fn execute_external_subcommand(cmd: &str, args: &[&OsStr], config: &Config)
     cmd.args(args);
     cmd.env(SCARB_ENV, config.app_exe()?);
     cmd.env("PATH", config.dirs().path_env());
+    cmd.env("SCARB_CACHE", config.dirs().cache_dir.path_unchecked());
+    cmd.env("SCARB_CONFIG", config.dirs().config_dir.path_unchecked());
+    cmd.env("SCARB_MANIFEST_DIR", config.root());
+    cmd.env(
+        "SCARB_LOG",
+        config.ui().verbosity().env_filter().to_string(),
```
mkaput, line 32: bug: this doesn't work

```
SCARB_LOG=scarb=trace,salsa=info scarb -vvvv xyz
```

The value of `SCARB_LOG` should be passed verbatim, but UI verbosity should still be sent somehow. I think it would be best to introduce `SCARB_UI_VERBOSITY` environment variable and pass it separately.

1. Serialization of verbosity: implement `FromStr for Verbosity` and `fmt::Display for Verbosity`
2. Add `Verbosity::from_env_var(env_var_name: &str)` and read the environment variable there. Be analogous to `EnvFilter` API.
3. I would store `SCARB_LOG` value as a `String` field in `Config`. Logs != UI

mkaput (changes requested):

scarb/src/core/config.rs
```diff
@@ -71,6 +75,10 @@ impl Config {
             .expect("parent of manifest path must always exist")
     }
 
+    pub fn scarb_log(&self) -> &String {
```
mkaput, line 78:
```suggestion
    pub fn scarb_log(&self) -> &str {
```

scarb/src/ui/mod.rs
```diff
@@ -37,22 +37,37 @@ impl Default for Verbosity {
...
+    fn from_str(s: &str) -> Result<Self> {
+        if s.is_empty() {
+            bail!("empty string cannot be used as verbosity level");
+        }
+
+        match s {
+            "quiet" => Ok(Verbosity::Quiet),
+            "normal" => Ok(Verbosity::Normal),
+            "verbose" => Ok(Verbosity::Verbose),
+            _ => bail!("invalid verbosity level: {}", s),
```
mkaput, line 62: clippy should yell with this:

```suggestion
            _ => bail!("invalid verbosity level: {s}"),
```

mkaput, line 62: FYI

```suggestion
        match s {
            "quiet" => Ok(Verbosity::Quiet),
            "normal" => Ok(Verbosity::Normal),
            "verbose" => Ok(Verbosity::Verbose),
            "" => bail!("empty string cannot be used as verbosity level"),
            _ => bail!("invalid verbosity level: {}", s),
```

scarb/tests/e2e/subcommand.rs
```diff
@@ -76,6 +77,40 @@ fn env_variables_are_passed() {
...
 
+#[test]
+#[cfg(unix)]
+fn env_scarb_log_is_passed_verbatim() {
+    let t = TempDir::new().unwrap();
+    write_script(
+        "env",
+        indoc! {
+            r#"
+            #!/bin/bash
```
mkaput, line 88: This is more portable as far as I know, not every OS keeps bash in `/bin` (I think NixOS could be one example)

```suggestion
            #!/usr/bin/env bash
```

mkaput (changes requested):

mkaput:
Please rebase, so that each commit resolves one topic

scarb/src/ui/mod.rs
```diff
@@ -119,11 +154,35 @@ impl Ui {
...
+
+    #[test]
+    fn verbosity_from_env_var() {
+        use Verbosity::*;
+        env::set_var("SOME_ENV_VAR", "quiet");
+        assert_eq!(Verbosity::from_env_var("SOME_ENV_VAR").unwrap(), Quiet);
+        env::set_var("SOME_ENV_VAR", "verbose");
+        assert_eq!(Verbosity::from_env_var("SOME_ENV_VAR").unwrap(), Verbose);
+
+        assert_true!(matches!(
```
mkaput, line 183:
```suggestion
        assert!(matches!(
```

this will work :D

mkaput (approved):
👏🏻

PR #95, opened 2023-02-22
Allow listing installed subcommands

### LINKED ISSUE #9: Implement `list-commands` command
It should do the same thing the `cargo --list` does. Perhaps we should also replace `commands` command with `--list` argument?

maciektr:
fix #9

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -13,7 +13,6 @@ use tracing_log::AsTrace;
 use scarb::core::PackageName;
 use scarb::manifest_editor::DepId;
 use scarb::metadata::MetadataVersion;
-use scarb::ops::CommandInfo;
 use scarb::ui;
 use scarb::ui::OutputFormat;
 use scarb::version;
@@ -92,16 +91,16 @@ impl ScarbArgs {
         }
     }
 
-    pub fn get_subcommands() -> BTreeMap<String, CommandInfo> {
-        let mut commands = BTreeMap::new();
-        let command = Self::command();
-        for sub in command.get_subcommands() {
-            let name = sub.get_name().to_string();
-            let about = sub.get_about().map(|s| s.to_string());
-            let info = CommandInfo::BuiltIn { about };
-            commands.insert(name, info);
-        }
-        commands
+    pub fn get_builtin_subcommands() -> BTreeMap<String, Option<String>> {
+        Self::command()
+            .get_subcommands()
+            .into_iter()
+            .map(|sub| {
+                let name = sub.get_name().to_string();
+                let about = sub.get_about().map(|s| s.to_string());
+                (name, about)
+            })
+            .collect()
     }
 }
 
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -2,16 +2,18 @@
 
 //! CLI arguments datastructures.
 
+use std::collections::BTreeMap;
 use std::ffi::OsString;
 
 use camino::Utf8PathBuf;
-use clap::{Parser, Subcommand};
+use clap::{CommandFactory, Parser, Subcommand};
 use tracing::level_filters::LevelFilter;
 use tracing_log::AsTrace;
 
 use scarb::core::PackageName;
 use scarb::manifest_editor::DepId;
 use scarb::metadata::MetadataVersion;
+use scarb::ops::CommandInfo;
 use scarb::ui;
 use scarb::ui::OutputFormat;
 use scarb::version;
@@ -88,6 +90,18 @@ impl ScarbArgs {
         } else {
             ui::Verbosity::Quiet
         }
+    }
+
+    pub fn get_subcommands() -> BTreeMap<String, CommandInfo> {
+        let mut commands = BTreeMap::new();
+        let command = Self::command();
+        for sub in command.get_subcommands() {
+            let name = sub.get_name().to_string();
+            let about = sub.get_about().map(|s| s.to_string());
+            let info = CommandInfo::BuiltIn { about };
+            commands.insert(name, info);
+        }
+        commands
     }
 }
 
--- a/scarb/src/bin/scarb/commands/commands.rs
+++ b/scarb/src/bin/scarb/commands/commands.rs
@@ -1,14 +1,140 @@
 use anyhow::Result;
+use serde::{Serialize, Serializer};
+use std::collections::BTreeMap;
+use std::path::PathBuf;
+use std::{env, fmt, fs};
 
 use crate::args::ScarbArgs;
 use scarb::core::Config;
-use scarb::ops;
+use scarb::process::is_executable;
+use scarb::ui::Message;
+use scarb::EXTERNAL_CMD_PREFIX;
+
+#[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Serialize)]
+enum CommandInfo {
+    BuiltIn { about: Option<String> },
+    External { path: PathBuf },
+}
+
+impl fmt::Display for CommandInfo {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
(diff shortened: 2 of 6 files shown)


scarb/src/bin/scarb/args.rs
```diff
@@ -49,6 +51,18 @@ impl Args {
             OutputFormat::default()
         }
     }
+
+    pub fn get_subcommands() -> BTreeMap<String, CommandInfo> {
```
mkaput, line 55:
```suggestion
    pub fn get_builtin_commands() -> BTreeMap<String, CommandInfo> {
```

scarb/src/ops/subcommands.rs
```diff
@@ -1,14 +1,87 @@
...
 use std::process::Command;
+use std::{env, fs};
 
 use anyhow::{bail, Result};
 
 use crate::core::Config;
 use crate::process::{exec_replace, is_executable};
 use crate::SCARB_ENV;
 
+pub const EXTERNAL_CMD_PREFIX: &str = "scarb-";
```
mkaput, line 14: hmm this makes this constant available as `scarb::ops::EXTERNAL_CMD_PREFIX` which doesn't sound right

I think you could move this to `src/lib.rs` along other constatnts

scarb/src/ops/subcommands.rs
```diff
@@ -1,14 +1,87 @@
...
+        }
+        Ok(())
+    }
+}
+
+#[tracing::instrument(level = "debug", skip(config, builtins))]
+pub fn list_commands(
+    config: &Config,
+    builtins: &mut BTreeMap<String, CommandInfo>,
+) -> CommandsList {
```
mkaput, line 50: I think we might not want to provide such types/functions in scarb as a library? They seem too coupled with the way Scarb's CLI works...

I think it would be better to move this to the `bin` part

scarb/tests/e2e/subcommand.rs
```diff
@@ -40,6 +40,78 @@ fn subcommand() {
         .stdout_eq("Hello beautiful world\n");
 }
 
+#[test]
+#[cfg_attr(
+    not(target_family = "unix"),
+    ignore = "This test should write a Rust code, because currently it only assumes Unix."
+)]
+fn list_commands_builtin_only() {
```
mkaput, line 48: these tests are going to be painful to maintain I think... each new command/slight change in docs would require to change this test

Wdyt about taking an unit-test-like approach and simply verify that `list-commands` includes hello, on CommandInfo-level? without ever looking into builtins

mkaput (changes requested):

scarb/src/bin/scarb/commands/commands.rs
```diff
@@ -1,9 +1,159 @@
...
+        text
+    }
+
+    fn structured<S: Serializer>(self, ser: S) -> Result<S::Ok, S::Error> {
+        self.commands.serialize(ser)
+    }
+}
+
+#[tracing::instrument(level = "debug", skip(config, builtins))]
+pub fn list_commands(
```
mkaput, line 60:
```suggestion
fn list_commands(
```

1. no need to instrument this because this is wrapped by `run` pretty swiftly
2. no need to pub it either because this is going only to be used here I guess?

scarb/src/bin/scarb/commands/commands.rs
```diff
@@ -1,9 +1,159 @@
...
+) -> CommandsList {
+    let prefix = EXTERNAL_CMD_PREFIX;
+    let suffix = env::consts::EXE_SUFFIX;
+
+    let mut commands = BTreeMap::new();
+    for dir in config.dirs().path_dirs.iter() {
+        let entries = match fs::read_dir(dir) {
+            Ok(entries) => entries,
+            _ => continue,
+        };
```
mkaput, line 72:
```suggestion
        let Ok(entries) = fs::read_dir(dir) else { continue };
```

scarb/src/bin/scarb/commands/commands.rs
```diff
@@ -1,9 +1,159 @@
...
+        for entry in entries.filter_map(|e| e.ok()) {
+            let path = entry.path();
+            let filename = match path.file_name().and_then(|s| s.to_str()) {
+                Some(filename) => filename,
+                _ => continue,
+            };
```
mkaput, line 78: same here, `let ... else`

scarb/src/bin/scarb/commands/commands.rs
```diff
@@ -1,9 +1,159 @@
...
+impl Message for CommandInfo {
+    fn text(self) -> String {
+        self.to_string()
+    }
+
+    fn structured<S: Serializer>(self, ser: S) -> Result<S::Ok, S::Error> {
+        self.serialize(ser)
+    }
+}
+
```
mkaput, line 39:
```suggestion
```

dead code?

scarb/src/bin/scarb/commands/commands.rs
```diff
@@ -1,9 +1,159 @@
...
+use std::{env, fs};
 
+use crate::args::ScarbArgs;
 use scarb::core::Config;
+use scarb::process::is_executable;
+use scarb::ui::Message;
+use scarb::EXTERNAL_CMD_PREFIX;
+
+#[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Serialize)]
+pub enum CommandInfo {
```
mkaput, line 15:
```suggestion
enum CommandInfo {
```

not used outside this file?

scarb/src/bin/scarb/commands/commands.rs
```diff
@@ -1,9 +1,159 @@
...
+    BuiltIn { about: Option<String> },
+    External { path: PathBuf },
+}
+
+impl Display for CommandInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
mkaput, line 21:
```suggestion
use std::fmt;

impl fmt::Display for CommandInfo {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```

nitpick: let's be consistent here with rest of code

mkaput (changes requested):

scarb/src/bin/scarb/commands/commands.rs
```diff
@@ -1,9 +1,139 @@
...
+    fn cmd_list() {
+        let t = TempDir::new().unwrap();
+
+        let cache_dir = TempDir::new().unwrap();
+        let config_dir = TempDir::new().unwrap();
+        let manifest = t.child("Scarb.toml");
+        let path_dir = t.child("bin");
+        let sub_cmd = path_dir.child("scarb-hello");
+        sub_cmd.touch().unwrap();
+        set_permissions(sub_cmd.path(), std::fs::Permissions::from_mode(0o755)).unwrap();
```
mkaput, line 116: You can move `make_executable` to `scarb::process` and use it here and thus this test will work on Windows

scarb/src/bin/scarb/commands/commands.rs
```diff
@@ -1,9 +1,139 @@
...
+
+        let config = Config::builder(Utf8Path::from_path(manifest.path()).unwrap().to_path_buf())
```
mkaput, line 118: `fsx` module has some helper traits for conversion to Utf8 paths

scarb/src/bin/scarb/commands/commands.rs
```diff
@@ -1,9 +1,139 @@
...
+        }
+        text
+    }
+
+    fn structured<S: Serializer>(self, ser: S) -> Result<S::Ok, S::Error> {
+        self.commands.serialize(ser)
+    }
+}
+
+fn list_commands(config: &Config, builtins: &mut BTreeMap<String, Option<String>>) -> CommandsList {
```
mkaput, line 48: hmm why is this `&mut` ?

mkaput (changes requested):

mkaput (approved):

PR #114, opened 2023-02-27
Add compilation units to metadata

maciektr:
cc #40

### DIFF
--- a/scarb/src/metadata/mod.rs
+++ b/scarb/src/metadata/mod.rs
@@ -1,20 +1,21 @@
 // NOTE: All collections must have stable sorting in order to provide reproducible outputs!
 
-use std::collections::BTreeMap;
-use std::path::PathBuf;
-
 use anyhow::{bail, Result};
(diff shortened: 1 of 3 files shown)


scarb/src/metadata/mod.rs
```diff
@@ -78,6 +81,14 @@ pub struct TargetMetadata {
     pub params: BTreeMap<String, Value>,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct CompilationUnitMetadata {
+    pub package: PackageMetadata,
+    pub target: TargetMetadata,
+    pub components: Vec<PackageMetadata>,
```
mkaput, line 88:
```suggestion
    pub package: PackageId,
    pub target: TargetMetadata,
    pub components: Vec<PackageId>,
```

I think, given that PackageMetadata for all packages is already printed, why duplicate this here?

mkaput (changes requested):

mkaput (approved):

PR #122, opened 2023-02-27
Add `scarb rm`

### LINKED ISSUE #93: Add scarb rm command
Like `cargo rm` for Scarb. 

- Add `RemoveDependency` `Op`, just like `Add` one is done in `manifest_editor`
- Add `rm` command
- Try to preserve Cargo's behaviour as much as possible

kariy:
fix #93 

Adds `remove` command with `rm` as alias. Currently, all it does is just removing the specified dependencies from the dependency table of a manifest file. It doesn't do any workspace clean up like `cargo rm` as there's no support for proper workspace yet as well as other types of dependency (e.g., dev, build).

I tried to refer as much as possible to `cargo`, so if there's anything lacking please let me know.

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -111,6 +111,11 @@ pub enum Command {
     // External should go last.
     /// Add dependencies to a Scarb.toml manifest file.
     Add(AddArgs),
+
+    /// Remove dependencies from a manifest file
+    #[command(alias = "rm")]
+    Remove(RemoveArgs),
+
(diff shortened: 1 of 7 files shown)


mkaput (changes requested):
This looks 🔥 My only concern is about use of `Option` in `RemoveDependency`, please address that and we're good to merge this to be released in next release ❤️

mkaput:
Also FYI: I will probably squash this PR, we used to combine related tests with commits that introduce tested logic :)

PR #124, opened 2023-02-28
Initialize Git repository on `scarb new` & `scarb init`

### LINKED ISSUE #80: Initialize Git repository in scarb new/init
### Problem

`scarb new` and `scarb init` currently create `.gitignore` file, but it would be also nice to let it automatically call `git init` for the user.

### Proposed Solution

Add new function `init_vcs` in `new` ops file and run it just before `write_vcs_ignore`. This should be on-by-default behaviour, with opt out via `--no-vcs` CLI flag.

### Notes

We do not want to bother yet with non-Git VCS systems, like Cargo for example is. Nobody uses anything else than Git in Cairo ecosystem as far as I am aware.

kariy:
Fix #80 

Refactor `write_vcs_ignore` a bit to allow some flexibility **in case** we want to support other version control systems in the future.

### DIFF
--- a/scarb/tests/e2e/new_and_init.rs
+++ b/scarb/tests/e2e/new_and_init.rs
@@ -25,6 +25,39 @@ fn new_simple() {
     assert!(t.child("Scarb.toml").is_file());
     assert!(t.child("src/lib.cairo").is_file());
     assert!(t.child(".gitignore").is_file());
+    assert!(t.child(".git").is_dir());
+
+    let toml_manifest = TomlManifest::read_from_path(t.child("Scarb.toml").utf8_path()).unwrap();
+    assert_eq!(toml_manifest.package.unwrap().name.as_str(), "hello");
+
+    Scarb::quick_snapbox()
+        .arg("build")
+        .current_dir(&t)
+        .assert()
+        .success();
+
+    t.child("target/release/hello.sierra")
+        .assert(predicates::str::is_empty().not());
+}
+
+#[test]
+fn new_simple_without_vcs() {
+    let pt = assert_fs::TempDir::new().unwrap();
+
+    Scarb::quick_snapbox()
+        .arg("new")
+        .arg("hello")
+        .arg("--no-vcs")
+        .current_dir(&pt)
+        .assert()
+        .success();
+
+    let t = pt.child("hello");
+    assert!(t.is_dir());
+    assert!(t.child("Scarb.toml").is_file());
+    assert!(t.child("src/lib.cairo").is_file());
+    assert!(!t.child(".gitignore").exists());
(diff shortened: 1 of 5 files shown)


scarb/tests/e2e/new_and_init.rs
```diff
@@ -25,6 +25,39 @@ fn new_simple() {
...
+        .current_dir(&t)
+        .assert()
+        .success();
+
+    t.child("target/release/hello.sierra")
+        .assert(predicates::str::is_empty().not());
+}
+
+#[test]
+fn new_simple_without_vcs() {
```
mkaput, line 44: Hmm, building Cairo code is the biggest factor of test suite execution time currently, so I would welcome avoiding it whenever possible. I think this test could limit itself just to ensure there's no `.git` and `.gitignore` and assume `new_simple` does the rest of the job. Wdyt?

This is not blocking. I can do this myself after merging this. I'd love to get this released today :)

scarb/tests/e2e/new_and_init.rs
```diff
@@ -55,6 +88,39 @@ fn init_simple() {
...
+        .current_dir(&t)
+        .assert()
+        .success();
+
+    t.child("target/release/hello.sierra")
+        .assert(predicates::str::is_empty().not());
+}
+
+#[test]
+fn init_simple_without_vcs() {
```
mkaput, line 107: Same here

mkaput (approved):
♥️♥️♥️♥️♥️

scarb/tests/e2e/new_and_init.rs
mkaput: https://github.com/software-mansion/scarb/commit/3421a411ff6554a6692113da47f22f5894d77933

kariy: I agree.

kariy: Just a random suggestion but do you think it's possible if we avoid building at all in all the tests under `new_and_init`? Just have something like `check_can_build( )` where it checks that the created/initialized package has everything needed to build successfully?

mkaput: This approach has a problem that it's easy to forget to update it in case something changes :(

mkaput: We might consider implementing `scarb check` though, but this requires investigating if skipping Sierra codegen speeds up anything

mkaput: #137 I don't know if I'll open this for external contributors though, seems to have a potential for invasive changes which we prefer to keep in-house

Issue #13, opened 2023-01-03
Support namespaced package names


mkaput:
```toml
[dependencies]
"org/package" = "1.0.0"
```

On both cases, only package word as package name should be fed to the compiler. The org part is purely for registry purposes and the resolver will reject the following manifest with duplicated package conflict error:
```toml
[dependencies]
"argent/account" = "1.0.0"
"openzeppelin/account" = "6.0.0"
```

Reason: Support publishing forks to the registry. Suppose package `company/X` has a critical bug, but the maintainer is not responsive. Community can create a fix, push it to a fork and publish this it as `community/X`. Users will only have to change their manifests, instead of having to also rename all references to `X` in their codebases.

Note: I would rather avoid Javaesque/Apple domain-styled organization names (i.e. prefer starkware instead of co.starkware etc.) and in order to enforce this I plan to ban names like x.y.z

PR #143, opened 2023-03-02
Sort `[dependencies]` section on `scarb add`

### LINKED ISSUE #128: Sort `[dependencies]` section in `scarb add`
https://github.com/software-mansion/scarb/blob/765aaad27a3529607d33576fb288d5286efe4953/scarb/src/manifest_editor/mod.rs#L59

kariy:
Fix #128 

Will sort if the `[dependencies]` section is already sorted otherwise leave it as is.

### DIFF
--- a/scarb/src/manifest_editor/add.rs
+++ b/scarb/src/manifest_editor/add.rs
@@ -52,8 +52,19 @@ impl Op for AddDependency {
 
         let dep = Dep::resolve(*self, ctx)?;
 
+        let was_sorted = tab.as_table_like().map_or(true, |dep_table| {
+            let values = dep_table.get_values();
+            is_sorted(values.iter().map(|(key, _)| key[0]))
+        });
+
         let dep_key = dep.toml_key().to_string();
         dep.upsert(tab.as_table_like_mut().unwrap().entry(&dep_key));
+
+        if was_sorted {
+            if let Some(table) = tab.as_table_like_mut() {
+                table.sort_values();
+            }
+        }
 
         if let Some(t) = tab.as_inline_table_mut() {
             t.fmt()
@@ -221,4 +232,21 @@ fn path_value(manifest_path: &Utf8Path, abs_path: &Utf8Path) -> String {
         .expect("Both paths should be absolute at this point.")
         .as_str()
         .replace('\\', "/")
+}
+
+// Based on Iterator::is_sorted from nightly std; remove in favor of that when stabilized.
+fn is_sorted(mut item: impl Iterator<Item = impl PartialOrd>) -> bool {
+    let mut last = match item.next() {
+        Some(e) => e,
+        None => return true,
+    };
+
+    for current in item {
+        if current < last {
+            return false;
+        }
+        last = current;
+    }
+
+    true
 }
--- a/scarb/src/manifest_editor/add.rs
+++ b/scarb/src/manifest_editor/add.rs
@@ -52,10 +52,10 @@ impl Op for AddDependency {
 
         let dep = Dep::resolve(*self, ctx)?;
 
-        let was_sorted = tab.as_table_like().map_or(true, |dep_table| {
-            let values = dep_table.get_values();
+        let was_sorted = {
+            let values = tab.as_table_like().unwrap().get_values();
             is_sorted(values.iter().map(|(key, _)| key[0]))
-        });
+        };
 
         let dep_key = dep.toml_key().to_string();
         dep.upsert(tab.as_table_like_mut().unwrap().entry(&dep_key));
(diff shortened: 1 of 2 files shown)


scarb/src/manifest_editor/add.rs
```diff
@@ -52,9 +52,20 @@ impl Op for AddDependency {
 
         let dep = Dep::resolve(*self, ctx)?;
 
+        let was_sorted = tab.as_table_like().map_or(true, |dep_table| {
```
mkaput, line 55: `get_table_mut` ensures `tab` is `TableLike`, so I think you could just `.unwrap()` here instead of doing `map_or`

Anyway, I think setting this to `false` in case it is not a table would be more correct -- we wouldn't want to touch this at all for sorting purpose

scarb/tests/e2e/add.rs
```diff
@@ -350,3 +350,61 @@ fn overwrite_change_source_from_path_to_git() {
...
+        .args(["add", "cat@2.0.0"])
+        .input(indoc! {r#"
+            [package]
+            name = "hello"
+            version = "1.0.0"
+
+            [dependencies]
+            bar = "1.0.0"
+            dep = "1.0.0"
+            foo = "1.0.0"
```
mkaput, line 366: how would this behave in such scenario, when adding `cat`?

```diff
  [dependencies]
  bar = "1.0.0"
+
  dep = "1.0.0"
  foo = "1.0.0"
```

I'm interested on what happens with the blank line

mkaput (changes requested):
This looks good! I just have two non-critical suggestions :) Thanks a lot for this contribution!

In the future, please leave a comment under an issue that you'd like to tackle it, like you used to do in the past. Thanks to this comment, I can assign you to the issue 😄

kariy:
> This looks good! I just have two non-critical suggestions :) Thanks a lot for this contribution!
> 
> Note: I'll merge this to `main` branch, but this won't land in Scarb 0.1. We don't want to release new features in RC period.
> 
> In the future, please leave a comment under an issue that you'd like to tackle it, like you used to do in the past. Thanks to this comment, I can assign you to the issue 😄

Yeah my bad 😓. I initially wanted to take this along with the other one, but when I opened the PR I realized I forgot to ask for this one first. Won't happen again, sir. 🫡

scarb/src/manifest_editor/add.rs
kariy: This because I wanted to map the option directly, not sure how I can do it more concisely using `unwrap` tho. 

I was thinking because `get_table_mut` already ensure that it's `TableLike`, might as well just set the default to `true`, but yeah I think makes more sense to default to `false`.

scarb/tests/e2e/add.rs
kariy: Ah.. this would just append the entry before the blank line, if it's already sorted. You reckon I should put that in the tests as well?

mkaput:
> Yeah my bad 😓. I initially wanted to take this along with the other one, but when I opened the PR I realized I forgot to ask for this one first. Won't happen again, sir. 🫡

No worries!

scarb/tests/e2e/add.rs
mkaput: Yeah, please definitely add a test, so that we can keep an eye on this behaviour. No need to deal with it explicitly, it's fine what it does!

scarb/src/manifest_editor/add.rs
mkaput:
```rust
let was_sorted = {
    let values = tab.as_table_like().unwrap().get_values();
    is_sorted(values.iter().map(|(key, _)| key[0]))
};
```

scarb/src/manifest_editor/add.rs
```diff
@@ -52,9 +52,20 @@ impl Op for AddDependency {
...
+            let values = dep_table.get_values();
+            is_sorted(values.iter().map(|(key, _)| key[0]))
```
mkaput, line 57: How does this deal with this input? Doesn't this panic?

```table
[dependencies]
foo.bar = "1.0.0"
```

Yes, that's still invalid code, but this **may** become valid with #13. No need to deal with this specially, only I'd like to ensure it's resilient to panics

scarb/src/manifest_editor/add.rs
kariy: Wow I can so dumb sometimes. Alright will change it to that.

scarb/src/manifest_editor/add.rs
kariy: Seem like this has more to do with how the manifest file is being deserialized? specifically the `PackageName` struct

mkaput: Yes, but take into account that here we are not deserializing it

mkaput (approved):
Looks great!

♥️♥️♥️♥️♥️

PR #151, opened 2023-03-03
feat: support user defined core version override

### LINKED ISSUE #90: Allow providing custom `core` version
1. Stabilize `no-core` field in the manifest, i.e. document it properly.
2. If the main package of a compilation unit has `no-core = true`, then make sure the compilation unit has no implicit `core` package in its components, and instead require the main package to provide custom `core` dependency.

tarrencev:
allows top level workspace to override corelib version like:

```
[dependencies]
core = { git = "https://github.com/starkware-libs/cairo.git", version = "1.0.0-alpha.3" }
```

this can also be useful for developers working on contributions to the corelib

resolves #90


mkaput (changes requested):
We have discussed this PR privately, there are some changes to be made but overall I'm open to merge this as an unstable temporary feature

PR #152, opened 2023-03-04
Generalize logic for getting selected package

### LINKED ISSUE #127: Generalize logic for parsing `-p` CLI argument
https://github.com/software-mansion/scarb/blob/bb5c31e7d5f4ea04bb57df2036d485654f3c8c4c/scarb/src/bin/scarb/commands/add.rs#L13-L20

kariy:
Fix #127 

Referring to Cargo, the problem I'm seeing with this if later we want to add flags like `--exclude` / `--all` for commands like `scarb build`, we can just add new fields in `PackagesFilter`, but then we can't disable the individual flags for commands that don't require them and they'd also be shown in the help message. Or maybe that's not something we should worry about atm? Wdyt?

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -10,7 +10,8 @@ use clap::{CommandFactory, Parser, Subcommand};
 use tracing::level_filters::LevelFilter;
 use tracing_log::AsTrace;
 
-use scarb::core::PackageName;
+use anyhow::{anyhow, Result};
+use scarb::core::{Package, PackageName, Workspace};
 use scarb::manifest_editor::DepId;
 use scarb::metadata::MetadataVersion;
 use scarb::ui;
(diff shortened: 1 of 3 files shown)


mkaput:
> Referring to Cargo, the problem I'm seeing with this if later we want to add flags like `--exclude` / `--all` for commands like `scarb build`, we can just add new fields in `PackagesFilter`, but then we can't disable the individual flags for commands that don't require them and they'd also be shown in the help message. Or maybe that's not something we should worry about atm? Wdyt?

No need to worry about this now, there are three reasons:
1. This is CLI interface, we don't have to worry about changing how this is structured as far as CLI input works the same.
2. Seems like premature overengineering to me :)
3. I actually think this can be easily implemented by writing "proxy structs" that will only expose subsets of fields and implement conversion methods to `PackageFilter`

mkaput (approved):
Looks great! Thank you a lot!

♥️♥️♥️♥️♥️

PR #162, opened 2023-03-08
Canonicalize Git URL in `scarb add`

### LINKED ISSUE #108: Canonicalize Git URLs in scarb add
We have a logic for canonicalization of Git URLs, primarily for GitHub. It's done by `CanonicalUrl` struct.

`scarb add` currently pastes Git URL to the manifest verbatim. In order to encourage consistent style of everything, I think it'd be nice to canonicalize these URLs that users are passing.

kariy:
Fix #108

Add trailing slash in tests because they have empty paths.

### DIFF
--- a/scarb/src/manifest_editor/add.rs
+++ b/scarb/src/manifest_editor/add.rs
@@ -3,10 +3,12 @@ use std::mem;
 use anyhow::{anyhow, ensure, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use toml_edit::{value, Document, Entry, InlineTable, Item};
+use url::Url;
 
 use crate::core::{GitReference, PackageName};
 use crate::internal::fsx;
 use crate::internal::fsx::PathBufUtf8Ext;
+use crate::sources::canonical_url::CanonicalUrl;
 
 use super::tomlx::get_table_mut;
 use super::{DepId, Op, OpCtx};
@@ -126,6 +128,13 @@ impl Dep {
                 Rev(rev.into())
             } else {
                 DefaultBranch
+            };
+
+            let git = match Url::parse(&git) {
+                Ok(url) => CanonicalUrl::new(&url)
+                    .map(|git_url| git_url.as_str().to_string())
+                    .unwrap_or(git),
+                Err(_) => git,
             };
 
             Box::new(GitSource {
--- a/scarb/src/sources/git/mod.rs
+++ b/scarb/src/sources/git/mod.rs
@@ -18,7 +18,7 @@ use crate::ui::Status;
 
 use super::PathSource;
 
-mod canonical_url;
+pub mod canonical_url;
 mod client;
 
 pub struct GitSource<'c> {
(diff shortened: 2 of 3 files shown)


scarb/src/manifest_editor/add.rs
```diff
@@ -128,6 +130,17 @@ impl Dep {
                 DefaultBranch
             };
 
+            let git = match CanonicalUrl::new(&Url::parse(&git)?) {
+                Ok(url) => url.as_str().to_string(),
+                Err(_) => {
+                    ctx.opts
+                        .config
+                        .ui()
+                        .warn(format!("dependency ({name}) has an invalid git URL {git}"));
```
mkaput, line 139: Why not simply print error message here (without causes)? That would actually tell more what's making Scarb upset.

scarb/src/manifest_editor/add.rs
```diff
@@ -128,6 +130,17 @@ impl Dep {
                 DefaultBranch
             };
 
+            let git = match CanonicalUrl::new(&Url::parse(&git)?) {
```
mkaput, line 133: Failures to parse the URL will cause this operation to error, while I think we would still rather print just a warning here.

kariy: Hmm.. was considering that but is there a reason not to panic if it cannot be parsed as url at all?

mkaput: My reasoning here is that user might make a typo and I think I wouldn't yell at his face for this at this stage? Rather print some warnings and let him peacefully fix a mistake in the edited file.

mkaput: I'm open to being convinced that I think wrong 🙃

scarb/src/manifest_editor/add.rs
kariy: i was thinking bcs on non `dry-run`, when its trying to resolve the workspace it will output the same error mesage so i was thinking maybe it'd be redundant?

mkaput: Oh that's true! It looks like this code shouldn't print anything at all then, and instead the verification step should do everything.

scarb/src/manifest_editor/add.rs
kariy: i feel like simply giving just warning would make user assumes that its not that big of a problem... and also maybe somewhat counter intuitive to still write to the manifest when we know it will eventually panic? wdyt?

but this one also has same the problem as above... will eventually panic when resolving workspace

mkaput: Scarb already follows a pattern that is prints warnings if it knows that future steps will likely fail. For example if you use non-lib dependency. 

I'm not sure. Generally I'd like the editor to avoid failures as much as possible and only guide what might go wrong

kariy: I see... I didnt realize about that pattern actually.

mkaput (approved):
♥️♥️♥️♥️♥️

mkaput:
Could you please squash all commits?

kariy:
Squashed

PR #176, opened 2023-03-14
Migrate from Smol to Tokio

### LINKED ISSUE #174: Migrate from Smol to Tokio

maciektr:
fix #174


scarb/tests/e2e/flock.rs
```diff
@@ -74,61 +74,62 @@ fn locking_build_artifacts() {
...
+    let barrier = Arc::new(Barrier::new(2));
 
-        s.spawn({
-            let barrier = barrier.clone();
-            move || {
-                barrier.wait();
-                drop(lock);
-            }
-        });
+    config.runtime_handle().spawn({
```
mkaput, line 94:
```suggestion
    tokio::spawn({
```

scarb/tests/e2e/flock.rs
```diff
@@ -74,61 +74,62 @@ fn locking_build_artifacts() {
...
     let t = assert_fs::TempDir::new().unwrap();
     ProjectBuilder::start()
         .name("hello")
         .version("0.1.0")
         .build(&t);
 
     let manifest = t.child("Scarb.toml");
-    let config = Scarb::test_config(manifest);
+    let handle = tokio::runtime::Handle::current();
+    let builder = Scarb::test_config_builder(manifest);
```
mkaput, line 88: you can inline this variable

scarb/src/sources/git/client.rs
```diff
@@ -263,20 +279,21 @@ impl<'d> GitCheckout<'d> {
         with_verbosity_flags(&mut cmd, config);
         cmd.args(["--config", "core.autocrlf=false"]);
         cmd.arg("--recurse-submodules");
-        cmd.arg(db.repo.path());
+        let repo = db.repo().await;
+        cmd.arg(repo.path());
```
mkaput, line 283: You're locking `repo` until end of the function. Please inline it. If you don't store a variable, but use an object directly in the expression, Rust will `drop` it immediately when expression ends.

scarb/src/sources/git/client.rs
mkaput, line 1: Hmm I am thinking about these changes, and more and more I don't like the need for guarding `repo` behind a `Mutex`. It all boils down to the fact that we are using two things here: `gix` and processes, and we only try to asyncify `processes`, because `gix` is and will probably always be sync.

I think we should revert changes here, in `scarb::process` and simply call https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

mkaput (changes requested):

scarb/src/ops/resolve.rs
```diff
@@ -36,22 +36,25 @@ impl WorkspaceResolve {
     fields(root = ws.root().to_string())
 )]
 pub fn resolve_workspace(ws: &Workspace<'_>) -> Result<WorkspaceResolve> {
-    async {
-        let source_map = SourceMap::preloaded(ws.members(), ws.config());
-        let mut registry_cache = RegistryCache::new(source_map);
+    ws.config().runtime_handle().block_on(
```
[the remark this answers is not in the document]
maciektr: Closing, as we have decided to go with `spawn_blocking` instead.

scarb/src/sources/git/client.rs
maciektr: Closing, as we have decided to go with spawn_blocking instead.

mkaput (approved):

mkaput (approved):

mkaput (approved):

PR #211, opened 2023-03-23
Add manifest profiles definition

maciektr:


scarb/src/bin/scarb/args.rs
```diff
@@ -139,6 +140,14 @@ pub enum Command {
     External(Vec<OsString>),
 }
 
+/// Arguments accepted by the `build` command.
+#[derive(Parser, Clone, Debug)]
+pub struct BuildArgs {
+    /// Specify profile to use.
+    #[command(flatten)]
+    pub profile_spec: ProfileSpec,
```
mkaput, line 148: I think I would make this global and just keep it as `-P/--profile`

maciektr:
For now this PR supports only simple profile inheritance, limited to `dev` (by default) and `release` base profiles.

scarb/tests/e2e/profiles.rs
mkaput, line 1: Cairo compiler is quite slow. Could all these tests try to check their results by looking at `scarb metadata` instead?

mkaput, line 1: I like the completeness of these tests 😍

scarb/src/bin/scarb/args.rs
```diff
@@ -292,6 +304,46 @@ pub struct GitRefGroup {
...
+            bail!("cannot specify both `--dev` and `--release`");
+        }
+        if self.release && self.profile.is_some() {
+            bail!("cannot specify both `--release` and `--profile`");
+        }
+        if self.release && self.dev {
+            bail!("cannot specify both `--release` and `--dev`");
+        }
+        Ok(())
+    }
```
mkaput, line 333: Clap has this built-in: https://github.com/software-mansion/scarb/blob/e10047e932fb2c53bceca01d87ba60f9dd240775/scarb/src/bin/scarb/args.rs#L281

scarb/src/bin/scarb/args.rs
```diff
@@ -292,6 +304,46 @@ pub struct GitRefGroup {
...
+        self.verify()?;
+        let profile = if self.release {
+            "release".into()
+        } else {
+            self.profile.clone().unwrap_or("dev".into())
+        };
+
+        ws.current_package()?.fetch_profile(&profile)
+    }
+}
```
mkaput, line 345: Looks like this function tries to do too much. Shouldn't this just return the profile name?

```suggestion
    pub fn determine(&self) -> SmolStr {
        match &self {
            Self { release: true, .. } => "release".into(),
            Self { dev: true, .. } => "dev".into(),
            Self { profile: Some(profile), .. } => profile.clone(),
            _ => "dev".into()
        }
    }
}
```

scarb/src/bin/scarb/args.rs
```diff
@@ -292,6 +304,46 @@ pub struct GitRefGroup {
...
+
+    pub fn fetch<'a>(&self, ws: &'a Workspace<'_>) -> Result<&'a Profile> {
+        self.verify()?;
+        let profile = if self.release {
+            "release".into()
+        } else {
+            self.profile.clone().unwrap_or("dev".into())
+        };
+
+        ws.current_package()?.fetch_profile(&profile)
```
mkaput, line 343: I would like to have `fetch_profile` on entire workspace

scarb/src/bin/scarb/commands/build.rs
```diff
@@ -1,10 +1,12 @@
...
+use crate::args::BuildArgs;
 use scarb::core::Config;
 use scarb::ops;
 
 #[tracing::instrument(skip_all, level = "info")]
-pub fn run(config: &Config) -> Result<()> {
+pub fn run(args: BuildArgs, config: &Config) -> Result<()> {
     let ws = ops::read_workspace(config.manifest_path(), config)?;
-    ops::compile(&ws)
+    let profile = args.profile_spec.fetch(&ws)?;
```
mkaput, line 10: Wouldn't it make sense if `Config` stored `profile_name` in its state?

On top of that, `Workspace` could also expose `profile_name` as a shortcut.

scarb/tests/e2e/flock.rs
```diff
@@ -24,11 +24,10 @@ fn locking_build_artifacts() {
...
 
     thread::scope(|s| {
-        let lock =
-            config
-                .target_dir()
-                .child("release")
-                .open_rw("hello.sierra", "artifact", &config);
+        let lock = config
+            .target_dir()
+            .child("dev")
```
mkaput, line 29:
```suggestion
            .child(config.profile_name())
```

scarb/src/bin/scarb/args.rs
mkaput, line 1: I am more and more convinced that global profile flag would make more sense for us... Seems that this would reduce plumbing, and would enforce all code paths to take profiles into consideration.

scarb/src/core/manifest/mod.rs
```diff
@@ -161,6 +163,25 @@ pub struct ManifestCompilerConfig {
...
+        }
+    }
+}
+
+impl Default for ManifestCompilerConfig {
+    fn default() -> Self {
+        Self::default_dev()
+    }
+}
+
```
mkaput, line 184: Wdyt about:
1. Introducing `DefaultForProfile` trait
2. Skipping `impl Default` for all profile-sensitive types

scarb/src/compiler/compilation_unit.rs
```diff
@@ -24,9 +23,8 @@ pub struct CompilationUnit {
     pub components: Vec<CompilationUnitComponent>,
     /// The profile contains information about *how* the build should be run, including debug
     /// level, etc.
+    // Includes the Cairo compiler configuration parameters to use in this unit.
     pub profile: Profile,
-    /// Cairo compiler configuration parameters to use in this unit.
-    pub compiler_config: ManifestCompilerConfig,
 }
```
mkaput, line 28: Hmm, shouldn't compiler_config be modified somewhere up? I currently think `profile` field here should perhaps only keep profile name?

mkaput (changes requested):
Partial review

scarb/src/ops/resolve.rs
```diff
@@ -63,6 +63,7 @@ pub fn resolve_workspace(ws: &Workspace<'_>) -> Result<WorkspaceResolve> {
 pub fn generate_compilation_units(
     resolve: &WorkspaceResolve,
     ws: &Workspace<'_>,
+    profile: &Profile,
```
mkaput, line 66: why not just `fetch_profile` inside?

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -150,6 +158,43 @@ pub struct TomlCairo {
...
+        ensure!(
+            !name.to_string().starts_with(".."),
+            format!("profile name cannot start with `..` prefix")
+        );
+        let sanitized = sanitize(name.as_str());
+        ensure!(
+            sanitized == name.as_str(),
+            format!("profile name `{name}` is not allowed, consider using `{sanitized}` instead")
+        );
+        Ok(Self(name))
```
mkaput, line 185: 1. Profile construction itself isn't guarding this. I think this should be part of `ProfileType` constructor. Target is validated here only because otherwise it would be possible to override `[lib]` with `[[targets.lib]]`
2. I think we are not strict enough here. Shouldn't profiles be always lowercase ascii alphanumeric, with `-` symbol only accepted?

mkaput (changes requested):

scarb/src/core/package/mod.rs
```diff
@@ -67,6 +69,10 @@ impl Package {
             .ok_or_else(|| anyhow!("package `{self}` has no target `{kind}`"))
     }
 
+    pub fn fetch_profile(&self, profile: &SmolStr) -> Result<Profile> {
+        self.manifest.profiles.fetch_profile(profile)
+    }
```
mkaput, line 74: This operation doesn't make sense. Profile is just a handle. You ask for underlying representation, and get back the handle, which you can easily construct yourself. What is more needed here is `has_profile` operation I think

scarb/src/compiler/profile.rs
```diff
@@ -1,19 +1,172 @@
...
+        }
+    }
+}
+
+pub trait DefaultForProfile {
+    fn default_for_profile(profile: &Profile) -> Self;
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct ProfileBuilder {
```
mkaput, line 83: I don't like this type. It sounds like a bad, leaky abstraction:
1. Leaky, because you import unrelated types into this module and type.
2. Also it looks more Javaesque than Rusty. This is object-oriented rather than data-.
 
I think what would play better here would be a mechanism for _merging_ two objects, which can be done on any `T: Serialize + Deserialize`: you serialize object A (here: default for profile) to TOML Value. Then you merge user globals into it. And finally you merge profile overrides.

mkaput (changes requested):

scarb/src/core/package/mod.rs
```diff
@@ -69,8 +68,8 @@ impl Package {
             .ok_or_else(|| anyhow!("package `{self}` has no target `{kind}`"))
     }
 
-    pub fn fetch_profile(&self, profile: &SmolStr) -> Result<Profile> {
-        self.manifest.profiles.fetch_profile(profile)
+    pub fn has_profile(&self, profile: &Profile) -> Result<bool> {
```
mkaput, line 71:
```suggestion
    pub fn has_profile(&self, profile: &Profile) -> bool {
```

this will never fail I think

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -412,3 +413,24 @@ impl DetailedTomlDependency {
...
     }
 }
+
+/// Merge two `toml::Value` serializable structs.
+pub fn toml_merge<'de, T, S>(target: &mut T, source: &S) -> Result<()>
+where
+    T: Clone + Serialize + Deserialize<'de>,
+    S: Serialize + Deserialize<'de>,
+{
+    let mut params = toml::Value::try_from(target.clone())?;
```
mkaput, line 423:
```suggestion
/// Merge two `toml::Value` serializable structs.
pub fn toml_merge<'de, T, S>(target: &mut T, source: &S) -> Result<()>
where
    T: Serialize + Deserialize<'de>,
    S: Serialize + Deserialize<'de>,
{
    let mut params = toml::Value::try_from(target)?;
```

https://docs.rs/serde/latest/serde/ser/trait.Serialize.html#impl-Serialize-for-%26'a+mut+T

mkaput (changes requested):

scarb/src/core/workspace.rs
```diff
@@ -56,6 +58,36 @@ impl<'c> Workspace<'c> {
...
+            .collect::<Vec<_>>();
+        names.push(Profile::DEV.to_string());
+        names.push(Profile::RELEASE.to_string());
+        names.sort();
+        Ok(names)
+    }
+
+    pub fn compiler_config(&self) -> Result<ManifestCompilerConfig> {
+        Ok(self.current_package()?.manifest.compiler_config.clone())
+    }
```
mkaput, line 90: dead code? this function doesn't also make sense, because different workspace packages may have different compiler configs

mkaput (approved):
🚀

mkaput (approved):

PR #289, opened 2023-04-27
List subcommands from Scarb exe dir

maciektr:
Fix listing subcommands to include Scarb exe directory, as in #277

### DIFF
--- a/scarb/tests/e2e/subcommand.rs
+++ b/scarb/tests/e2e/subcommand.rs
@@ -224,3 +224,34 @@ fn can_find_scarb_directory_scripts_without_path() {
         .success()
         .stdout_eq("Hello beautiful world\n");
 }
+
+#[test]
+fn can_list_scarb_directory_scripts() {
+    let t = assert_fs::TempDir::new().unwrap();
+    write_script(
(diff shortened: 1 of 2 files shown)


scarb/tests/e2e/subcommand.rs
```diff
@@ -224,3 +224,38 @@ fn can_find_scarb_directory_scripts_without_path() {
         .success()
         .stdout_eq("Hello beautiful world\n");
 }
+
+#[test]
+#[cfg_attr(
+    not(target_family = "unix"),
+    ignore = "This test should write a Rust code, because currently it only assumes Unix."
+)]
+fn can_list_scarb_directory_scripts() {
```
mkaput, line 233: This test technically doesn't run the code, so it could be written in a Windows-supporting way. You simply can write any rubbish to the file and make it executable on Unix or add `.exe` suffix on Windows (https://doc.rust-lang.org/stable/std/env/consts/constant.EXE_SUFFIX.html)

mkaput (changes requested):

mkaput (approved):

PR #290, opened 2023-04-28
Add cairo-version to Scarb.toml and comparing it with current cairo version

### LINKED ISSUE #226: Add `cairo-version` field to Scarb.toml
Basically mimic behaviour of `rust-version` in Cargo

4rgorok:
Adding `cairo-version` to `Scarb.toml` and comparing it with current cairo version

Closes #226

### DIFF
--- a/scarb/src/core/manifest/mod.rs
+++ b/scarb/src/core/manifest/mod.rs
@@ -148,6 +148,7 @@ pub struct ManifestMetadata {
     pub repository: Option<String>,
     #[serde(rename = "tool")]
     pub tool_metadata: Option<BTreeMap<SmolStr, Value>>,
+    pub cairo_version: Option<VersionReq>,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq, Hash)]
--- a/scarb/src/ops/resolve.rs
+++ b/scarb/src/ops/resolve.rs
@@ -13,6 +13,7 @@ use crate::core::registry::Registry;
 use crate::core::resolver::Resolve;
 use crate::core::workspace::Workspace;
 use crate::core::Target;
+use crate::internal::to_version::ToVersion;
 use crate::resolver;
 
 pub struct WorkspaceResolve {
@@ -142,6 +143,8 @@ fn generate_cairo_compilation_units(
     assert!(!packages.is_empty());
     assert_eq!(packages[0].id, member.id);
 
+    check_cairo_version_compatibility(&packages, ws)?;
+
     // Print warnings for dependencies that are not usable.
     for pkg in other {
         ws.config().ui().warn(format!(
@@ -199,4 +202,26 @@ fn generate_cairo_compilation_units(
 /// Build a set of `cfg` items to enable while building the compilation unit.
 fn build_cfg_set(target: &Target) -> CfgSet {
     CfgSet::from_iter([Cfg::kv("target", target.kind.clone())])
+}
+
+fn check_cairo_version_compatibility(packages: &[Package], ws: &Workspace<'_>) -> Result<()> {
+    let current_version = crate::version::get().cairo.version.to_version().unwrap();
+    let matching_version = packages.iter().all(|pkg| {
+        match &pkg.manifest.metadata.cairo_version {
+            Some(package_version) if !package_version.matches(&current_version) => {
+                ws.config().ui().error(format!(
+                    "Package {}. Required Cairo version isn't compatible with current version. Should be: {} is: {}",
+                    pkg.id.name, package_version, current_version
+                ));
+                false
+            }
+            _ => true
+        }
+    });
+    if !matching_version {
+        anyhow::bail!(
+            "For each package, the required Cairo version must match the current Cairo version."
+        );
+    }
+    Ok(())
 }
--- a/scarb/tests/e2e/build.rs
+++ b/scarb/tests/e2e/build.rs
@@ -232,6 +232,77 @@ fn compile_with_invalid_version() {
 }
 
 #[test]
+fn compile_with_invalid_cairo_version() {
+    let t = TempDir::new().unwrap();
+    t.child("Scarb.toml")
+        .write_str(
+            r#"
+            [package]
+            name = "foo"
+            version = "0.1.0"
+            cairo-version = "f"
+            "#,
+        )
+        .unwrap();
+    Scarb::quick_snapbox()
+        .arg("build")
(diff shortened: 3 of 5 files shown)


scarb/src/core/manifest/mod.rs
```diff
@@ -148,6 +148,7 @@ pub struct ManifestMetadata {
     pub repository: Option<String>,
     #[serde(rename = "tool")]
     pub tool_metadata: Option<BTreeMap<SmolStr, Value>>,
+    pub cairo_version: Option<Version>,
```
mkaput, line 151: This should be `VersionReq` isn't it? We want to allow users specifying ranges of Cairo versions

```suggestion
    pub cairo_version: Option<VersionReq>,
```

scarb/src/ops/resolve.rs
```diff
@@ -78,10 +89,28 @@ pub fn generate_compilation_units(
                         member.id, pkg.id.name
                     ));
                 }
+
+                let package_version = pkg.manifest.metadata.cairo_version.clone();
+                if package_version.is_some() {
+                    let package_version = package_version.unwrap();
```
maciektr, line 95:
```suggestion
                if let Some(package_version) = package_version  {
```
FYI https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html

scarb/src/ops/resolve.rs
```diff
@@ -67,6 +69,15 @@ pub fn generate_compilation_units(
...
     for member in ws.members() {
+        let current_version = Version::parse(&*get().cairo.version)?;
+        // Remove any pre-release identifiers for easier comparison
+        let untagged_version = semver::Version::new(
+            current_version.major,
+            current_version.minor,
+            current_version.patch,
+        );
+
+        let mut matching_version = true;
```
maciektr, line 80: I would suggest moving the logic out of the filter closure instead, with something among the lines of `let matching_version = packages.iter().all(|pkg| { ... });`

mkaput (changes requested):
Minor code style protips + I think dropping prerelease tag is not okay yet for us.

This looks like it's quite close to be done 🎉

mkaput (commented):
Don't forget to write tests!

scarb/tests/e2e/build.rs
```diff
@@ -231,6 +231,60 @@ fn compile_with_invalid_version() {
...
+                TOML parse error at line 5, column 29
+                  |
+                5 |             cairo-version = "f"
+                  |                             ^^^
+                unexpected character 'f' while parsing major version number
+        "#});
+}
+
+#[test]
+fn compile_with_incompatible_cairo_version() {
```
mkaput, line 265: Could you please also add a test which compiles successfully?

scarb/src/ops/resolve.rs
```diff
@@ -123,6 +96,28 @@ pub fn generate_compilation_units(
...
+                }
+            }
+            is_matching
+        });
+
+        if !matching_version {
+            anyhow::bail!(
+                "For each package, the required Cairo version must match the current Cairo version."
+            );
+        }
```
mkaput, line 119: Consider extracting this to separate function

scarb/src/ops/resolve.rs
```diff
@@ -123,6 +96,28 @@ pub fn generate_compilation_units(
...
+            if let Some(package_version) = package_version  {
+                is_matching = package_version.matches(&current_version);
+                if !is_matching {
+                    ws.config().ui().error(format!(
+                        "Package {}. Required Cairo version isn't compatible with current version. Should be: {} is: {}",
+                        pkg.id.name, package_version.to_string(), current_version.to_string()
+                    ));
+                }
+            }
+            is_matching
```
mkaput, line 112: protip, this would be more idiomatic I think:

```suggestion
            match &pkg.manifest.metadata.cairo_version {
                Some(package_version) if !package_version.matches(&current_version) => {
                    ws.config().ui().error(format!(
                        "Package {}. Required Cairo version isn't compatible with current version. Should be: {} is: {}",
                        pkg.id.name, package_version, current_version
                    ));
                    false
                }
                _ => true
            }
```

note that I have removed `to_string()` calls, because they come from `ToString` trait which is implemented if a type implements `Display`, and `format!` macro can happily consume `Display` types

mkaput (changes requested):
Looks very good! Only the missing test is what I really need here

scarb/src/ops/resolve.rs
```diff
@@ -183,3 +167,25 @@ async fn collect_packages_from_resolve_graph(
 fn build_cfg_set(target: &Target) -> CfgSet {
     CfgSet::from_iter([Cfg::kv("target", target.kind.clone())])
 }
+
+#[tracing::instrument(skip_all, level = "debug")]
```
mkaput, line 171:
```suggestion
```

I don't think we'll need to measure execution times of this function :)

scarb/tests/e2e/support/project_builder.rs
```diff
@@ -51,6 +53,11 @@ impl ProjectBuilder {
         self
     }
 
+    pub fn cairo_version(mut self, cairo_version: impl ToVersion) -> Self {
```
mkaput, line 56: 🔥

scarb/src/ops/resolve.rs
```diff
@@ -183,3 +167,25 @@ async fn collect_packages_from_resolve_graph(
...
+pub fn is_cairo_version_compatible(
```
mkaput, line 172:
```suggestion
fn check_cairo_version_compatibility(
```

this is not a predicate, `is_` naming is misleading here, it'd be slightly more enclosed function if it returned `Result<()>` and bailed inside

scarb/src/ops/resolve.rs
```diff
@@ -183,3 +167,25 @@ async fn collect_packages_from_resolve_graph(
...
+    packages: &Vec<Package>,
```
mkaput, line 173:
```suggestion
    packages: &[Package],
```

passing slices is favourable over passing reference to a vector

mkaput (changes requested):

mkaput (approved):
Please resolve merge conflict and let's see if CI passes. To my eyes this looks good to go, though I see that `cargo fmt` and possibly clippy will complain

mkaput (changes requested):
Please rebase and create sensible commit history on this branch before merging

mkaput (approved):

PR #296, opened 2023-05-02
Clarify information about package names in docs

mkaput:
- Remove package name regex from docs
- Explicitly state that it must not be a Cairo keyword


piotmag769 (commented):
Doesn't the code allow for packages starting with the underscore though?

https://github.com/software-mansion/scarb/blob/main/scarb/src/core/package/name.rs#LL47C8-L47C8

mkaput:
Ah, right, I'll fix docs

PR #302, opened 2023-05-04
Remove invalid sentence that package names cannot start with `_`

mkaput:
https://github.com/software-mansion/scarb/pull/296#pullrequestreview-1412973251

### DIFF
--- a/website/pages/docs/reference/manifest.mdx
+++ b/website/pages/docs/reference/manifest.mdx
@@ -27,7 +27,6 @@ The package name is an identifier used to refer to the package.
 It is used when listed as a dependency in another package, and as the default name of targets.
 
 The name must use only ASCII alphanumeric characters or `_`, and cannot be empty.
-It cannot also start with underscore.
 It also must not be a valid Cairo identifier.
 
 ### `version`
--- a/website/pages/docs/reference/manifest.mdx
+++ b/website/pages/docs/reference/manifest.mdx
@@ -23,11 +23,11 @@ The only required fields are [`name`](#name) and [`version`](#version).
 
 ### `name`
 
-The package name is an identifier used to refer to the package.
+The package name is a valid Cairo identifier used to refer to the package.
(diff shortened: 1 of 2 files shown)


piotmag769 (approved):
It leaves room for weird package names like "___", but I would leave it as it is.
I think it is also worth to add the 4th condition (that the package name cannot be equal to `_`) [there](https://github.com/software-mansion/scarb/blob/main/scarb/src/core/package/name.rs#LL24C38-L24C38) and change the regex right below that to:
```regex
^(?!_$)[a-zA-Z_][0-9a-zA-Z_]*$
```

website/pages/docs/reference/manifest.mdx
```diff
@@ -27,7 +27,6 @@ The package name is an identifier used to refer to the package.
 It is used when listed as a dependency in another package, and as the default name of targets.
 
 The name must use only ASCII alphanumeric characters or `_`, and cannot be empty.
-It cannot also start with underscore.
```
piotmag769, line 30:
```suggestion
It cannot also be equal to `_`.
```

mkaput: `_` is not a valid identifier in Cairo so that's already covered

mkaput:
> It leaves room for weird package names like "___", but I would leave it as it is. I think it is also worth adding the 4th condition (that the package name cannot be equal to `_`) [there](https://github.com/software-mansion/scarb/blob/main/scarb/src/core/package/name.rs#LL24C38-L24C38) and change the regex right below that to:
> 
> ```regexp
> ^(?!_$)[a-zA-Z_][0-9a-zA-Z_]*$
> ```

If `___` is a valid Cairo identifier, then I think we must also permit it as a valid package name. We could impose such restrictions on the package registry level, but the package manager should always keep package names cover Cairo identifier set entirely

piotmag769 (approved):

PR #336, opened 2023-05-17
Document cairo-version field in manifest reference

mkaput:


website/pages/docs/reference/manifest.mdx
```diff
@@ -40,6 +40,19 @@ Scarb bakes in the concept of [Semantic Versioning](https://semver.org/), so mak
    Always increment the minor version if you add any new public structs, traits, fields, types, functions, methods, impls or anything else.
 4. Use version numbers with three numeric parts such as 1.0.0 rather than 1.0.
 
+### `cairo-version`
+
+The `cairo-version` field is an optional key that tells Scarb what version of the Cairo language and compiler your package can be compiled with.
+If the currently selected version of the Scarb compiler is older than the stated version, Scarb will exit with an error, telling the user what version is required.
```
maciektr, line 46: > compiler is older
Does it necessarily need to be older? It seems to me that something like "does not match requirements" would be more precise.

website/pages/docs/reference/manifest.mdx
```diff
@@ -40,6 +40,19 @@ Scarb bakes in the concept of [Semantic Versioning](https://semver.org/), so mak
    Always increment the minor version if you add any new public structs, traits, fields, types, functions, methods, impls or anything else.
 4. Use version numbers with three numeric parts such as 1.0.0 rather than 1.0.
 
+### `cairo-version`
```
maciektr, line 43: I feel like we should emphasize that this field does not affect the version of compiler used by scarb in any way, as I feel like this will be a common misconception among users.

website/pages/docs/reference/manifest.mdx
mkaput: Done.

maciektr (approved):

PR #378, opened 2023-05-25
Upgrade Cairo to `1.1.0`

mkaput:


mkaput:
Due to the issue with Cairo prerelease versioning conflicting with semver we will probably postpone this and wait for Cairo 1.1.0 stable. Topic is being discussed with Starkware.

PR #382, opened 2023-05-29
Scarb Test Runner 🚀

### LINKED ISSUE #222: Cairo test runner extension
You will have `scarb protostar` for running tests with Protostar
You will have `scarb cairo-test` for running tests with cairo-test-runner
You will have `scarb test` which will be an alias for `scarb run test`, i.e. this will run a script named test. Users will be able to select their test runner with:

```toml
[scripts]
test = "scarb protostar"
```

We don’t want to bundle either of these tools with default Scarb installation yet, in the future we lean towards bundling Protostar.

### LINKED ISSUE #389: Write tests in `starknet_hello_world` example

mkaput:
fix #222 
fix #389

### DIFF
--- a/extensions/scarb-cairo-test/Cargo.toml
+++ b/extensions/scarb-cairo-test/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "scarb-cairo-test"
+version.workspace = true
+rust-version.workspace = true
+edition.workspace = true
+publish = false
+
+authors.workspace = true
+
+[dependencies]
+anyhow.workspace = true
+cairo-lang-compiler.workspace = true
+cairo-lang-filesystem.workspace = true
+cairo-lang-starknet.workspace = true
+cairo-lang-test-runner.workspace = true
+clap.workspace = true
+scarb-metadata = { path = "../../scarb-metadata", features = ["packages_filter"] }
+serde_json.workspace = true
+
+[dev-dependencies]
+snapbox.workspace = true
+test-for-each-example = { path = "../../utils/test-for-each-example" }
--- a/extensions/scarb-cairo-test/src/main.rs
+++ b/extensions/scarb-cairo-test/src/main.rs
@@ -1,0 +1,139 @@
+use std::iter;
+use std::sync::Arc;
+
+use anyhow::Result;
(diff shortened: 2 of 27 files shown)


extensions/scarb-cairo-test/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+version.workspace = true
+rust-version.workspace = true
+edition.workspace = true
+publish = false
+
+authors.workspace = true
+
+[dependencies]
+anyhow.workspace = true
+cairo-lang-compiler = { version = "1.1.0-rc0", git = "https://github.com/starkware-libs/cairo.git", rev = "9d6cd56" }
```
maciektr, line 12: Why not `1.1.0` instead? 🤔

mkaput: I would have to use even newer commit here, because at that particular one the version haven't been changed yet :(

mkaput: And this commit if "after" 1.1.0 release

extensions/scarb-cairo-test/tests/examples.rs
```diff
@@ -0,0 +1,14 @@
+use std::path::Path;
+
+use snapbox::cmd::{cargo_bin, Command};
+
+use test_for_each_example::test_for_each_example;
+
+#[test_for_each_example]
```
maciektr, line 7: 🚒 🚒 🔥

maciektr (approved):

extensions/scarb-cairo-test/src/main.rs
```diff
@@ -0,0 +1,139 @@
...
+    #[arg(short, long, default_value = "")]
+    filter: String,
+
+    /// Run ignored and not ignored tests.
+    #[arg(long, default_value_t = false)]
+    include_ignored: bool,
+
+    /// Run only ignored tests.
+    #[arg(long, default_value_t = false)]
+    ignored: bool,
```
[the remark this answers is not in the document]
mkaput: Original test runner has it named like this, and also cargo test. I don't think we want to change it

PR #412, opened 2023-06-13
Update Cairo to 2.0.0-rc2

maciektr:


examples/starknet_hello_world/src/lib.cairo
```diff
@@ -1,58 +1,65 @@
...
-    }
-
-    // Returns the current balance.
-    #[view]
-    fn get_balance() -> felt252 {
-        balance::read()
+    fn replace(ref self: ContractState, a: u128) -> u128 {
+        let value = self.value.read();
+        self.value.write(a);
+        value
```
cptartur, line 24: What about getting the current balance? I think we should still have a method for that, one that takes `self: @ContractState`

examples/starknet_hello_world/src/lib.cairo
```diff
@@ -1,58 +1,65 @@
...
 
     #[constructor]
-    fn constructor(balance_: felt252) {
-        balance::write(balance_);
+    fn constructor(ref self: ContractState, value_: u128) {
+        self.value.write(value_);
     }
 
-    // Increases the balance by the given amount.
     #[external]
```
cptartur, line 20: The blogpost https://community.starknet.io/t/cairo-1-contract-syntax-is-evolving/94794 had something about external now changing to `#[external(v0)]`. Shouldn't this be changed here?

examples/starknet_hello_world/src/lib.cairo
cptartur: I'm pretty sure external functions shouldn't be allowed to return values. Transactions to starknet don't have any return data anyway so not sure how that would work.

examples/starknet_multiple_contracts/src/balance.cairo
```diff
@@ -1,18 +1,19 @@
...
     }
 
     // Returns the current balance.
     #[view]
-    fn get_balance() -> felt252 {
-        balance::read()
+    fn get_balance(ref self: ContractState) -> felt252 {
+        self.balance.read()
     }
 }
```
cptartur, line 19: Shouldn't this also define a starknet::interface?

cptartur, line 19: I think `#[view]` tag is no longer relevant. This method should just accept `self: @ContractState` - immutable snapshot of state instead of mutable reference.

scarb/src/compiler/compilers/starknet_contract.rs
```diff
@@ -5,7 +5,7 @@ use anyhow::{bail, Context, Result};
 use cairo_lang_compiler::db::RootDatabase;
 use cairo_lang_starknet::allowed_libfuncs::{
     validate_compatible_sierra_version, AllowedLibfuncsError, ListSelector,
-    DEFAULT_EXPERIMENTAL_LIBFUNCS_LIST,
+    BUILTIN_EXPERIMENTAL_LIBFUNCS_LIST,
```
cptartur, line 8: I'm assuming this is the same value, just renamed in the cairo2?

cptartur (commented):
I didn't comment on all contracts present in this PR as it would result in duplicate comments.

But my suggestion apply to other contracts as well.

scarb/src/compiler/compilers/starknet_contract.rs
maciektr: It used to be "experimental_v0.1.0", now it is just "experimental", the underlying list has been renamed.

scarb/tests/e2e/build_starknet_contract.rs
```diff
@@ -13,32 +13,68 @@ use crate::support::command::Scarb;
...
+                // Returns the current balance.
+                fn get(self: @T) -> u128;
+                // Increases the balance by the given amount.
+                fn increase(ref self: T, a: u128);
+            }
+        "#}
+    };
+}
+
+const BALANCE_CONTRACT_INTERFACE: &str = BALANCE_CONTRACT_INTERFACE!();
```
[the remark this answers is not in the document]
maciektr: It could be used here directly, but not in `concat!` macro below.

scarb/tests/e2e/build_starknet_contract.rs
```diff
@@ -6,39 +6,66 @@ use assert_fs::prelude::*;
...
+                self.value.write( self.value.read() + a );
+            }
         }
     }
 "#};
 
+static BALANCE_CONTRACT: Lazy<String> = Lazy::new(|| {
+    format!(
+        "{}\n{}",
+        BALANCE_CONTRACT_INTERFACE, BALANCE_CONTRACT_IMPLEMENTATION
```
cptartur, line 57: Nit, these could be used directly in format I think.
```suggestion
        "{BALANCE_CONTRACT_INTERFACE}\n{BALANCE_CONTRACT_IMPLEMENTATION}",
```

cptartur (approved):

PR #416, opened 2023-06-14
Bump cairo to 2.0.0-rc2

tarrencev:


tarrencev:
Hmm I see there is a 2.0.0-rc1 PR already

PR #423, opened 2023-06-18
Compile used contracts

### LINKED ISSUE #406: Verify if user can compile contract from a dependency in the project
Use-case is deploying those contracts using protostar
It should be done by adding something like this to `lib.cairo`
```
use my_contract::ContractA;
```

maciektr:
fix #406 
CC @MaksymilianDemitraszek


extensions/scarb-cairo-test/Cargo.toml
```diff
@@ -16,6 +16,7 @@ clap.workspace = true
 scarb-metadata = { path = "../../scarb-metadata", features = ["packages_filter"] }
 serde_json.workspace = true
 
+cairo-lang-utils.workspace = true
```
mkaput, line 19: this line seems to be put off

scarb/src/compiler/compilers/starknet_contract.rs
```diff
@@ -213,6 +221,69 @@ impl Compiler for StarknetContractCompiler {
...
+            if let ModuleId::Submodule(submodule_id) = module_id {
+                contracts.push(ContractDeclaration { submodule_id });
+            } else {
+                panic!("Contract `{contract_name}` was not found.");
+            }
+        }
+    }
+    contracts
+}
+
```
mkaput, line 286: could we upstream these changes to the compiler itself?

mkaput (changes requested):

scarb/src/compiler/compilers/starknet_contract.rs
maciektr: Sure! Do you think we can upstream whole logic of searching imported modules, or only the `find_module_contracts` function that we can reuse?

mkaput: whole logic. I think `starknet-compile` should also behave like this for consistency

maciektr:
superseded by https://github.com/software-mansion/scarb/pull/458

maciektr:
superseded by https://github.com/software-mansion/scarb/pull/458

PR #426, opened 2023-06-20
Update install script

maciektr:
- Stop `check_proc` and `get_bitness` warnings from popping up
- Use `sh` consistently in docs instead 
- Allow skipping `PATH` modification (e.g. for docker builds)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -26,8 +26,9 @@ The installer for Scarb
 Usage: install.sh [OPTIONS]
 
 Options:
-    -v        Specify Scarb _requested_version to install
-    -h        Print help information
+    -h | --help             Print help information
+    -p | --no-modify-path   Skip PATH variable modification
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -26,8 +26,9 @@ The installer for Scarb
 Usage: install.sh [OPTIONS]
 
 Options:
-    -v        Specify Scarb _requested_version to install
-    -h        Print help information
+    -h | --help             Print help information
+    -p | --no-modify-path   Skip PATH variable modification
+    -v | --version          Specify Scarb _requested_version to install
```
mkaput, line 31:
```suggestion
  -p, --no-modify-path   Skip PATH variable modification
  -h, --help             Print help
  -v, --version          Specify Scarb version to install
```

let's stick with `clap` style:
image

mkaput (approved):

PR #437, opened 2023-06-29
Resolver: add all dependencies to graph

maciektr:


maciektr:
Will be stacked with spr

PR #445, opened 2023-06-30
Load workspaces

maciektr:
**Stack**:
- #508
- #506
- #502
- #500
- #490
- #489
- #465
- #446
- #447
- #445 ⬅

### DIFF
--- a/scarb/src/core/manifest/maybe_workspace.rs
+++ b/scarb/src/core/manifest/maybe_workspace.rs
@@ -1,0 +1,80 @@
+// The following implementation has been copied from the `Cargo` codebase with slight modifications only.
+// The original implementation can be found here:
+// https://github.com/rust-lang/cargo/blob/31eda6f7c360d9911f853b3014e057db61238f3e/src/cargo/util/toml/mod.rs#L1071
+
+use anyhow::{bail, Context, Result};
+use serde::{de, Deserialize, Serialize};
+
+/// This Trait exists to make [`MaybeWorkspace::Workspace`] generic. It makes deserialization of
+/// [`MaybeWorkspace`] much easier, as well as making error messages for
+/// [`MaybeWorkspace::resolve`] much nicer.
+///
+/// Implementors should have a field `workspace` with the type of `bool`. It is used to ensure
+/// `workspace` is not `false` in a `Scarb.toml`.
+pub trait WorkspaceInherit {
+    /// This is the workspace table that is being inherited from.
+    /// For example `[workspace.dependencies]` would be the table "dependencies".
+    fn inherit_toml_table(&self) -> &str;
+
+    /// This is used to output the value of the implementors `workspace` field.
+    fn workspace(&self) -> bool;
+}
+
+/// An enum that allows for inheriting keys from a workspace in a Scarb.toml.
+#[derive(Clone, Debug, Serialize)]
+#[serde(untagged)]
+pub enum MaybeWorkspace<T, W: WorkspaceInherit> {
+    /// The type when inheriting from a workspace.
+    Workspace(W),
+    /// The "defined" type, or the type that that is used when not inheriting from a workspace.
+    Defined(T),
+}
+
+impl<'de, T: Deserialize<'de>, W: WorkspaceInherit + de::Deserialize<'de>> de::Deserialize<'de>
+    for MaybeWorkspace<T, W>
+{
+    fn deserialize<D>(deserializer: D) -> Result<MaybeWorkspace<T, W>, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        let value = serde_value::Value::deserialize(deserializer)?;
+
+        if let Ok(w) = W::deserialize(serde_value::ValueDeserializer::<D::Error>::new(
+            value.clone(),
+        )) {
+            return if w.workspace() {
+                Ok(MaybeWorkspace::Workspace(w))
+            } else {
+                Err(de::Error::custom("`workspace` cannot be false"))
+            };
+        }
+        T::deserialize(serde_value::ValueDeserializer::<D::Error>::new(value))
+            .map(MaybeWorkspace::Defined)
+    }
+}
+
+impl<T, W: WorkspaceInherit> MaybeWorkspace<T, W> {
+    pub fn resolve(self, label: &str, get_ws_inheritable: impl FnOnce() -> Result<T>) -> Result<T> {
+        match self {
+            MaybeWorkspace::Defined(value) => Ok(value),
+            MaybeWorkspace::Workspace(w) => {
+                if !w.workspace() {
+                    bail!("`workspace` cannot be false");
+                }
+                get_ws_inheritable().with_context(|| {
+                    format!(
+                        "error inheriting `{label}` from workspace root manifest's `workspace.{}.{label}`",
+                        w.inherit_toml_table(),
+                    )
+                })
+            }
+        }
+    }
+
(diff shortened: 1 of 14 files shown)


scarb/src/core/manifest/maybe_workspace.rs
```diff
@@ -0,0 +1,76 @@
...
+/// [`MaybeWorkspace::resolve`] much nicer
+///
+/// Implementors should have a field `workspace` with the type of `bool`. It is used to ensure
+/// `workspace` is not `false` in a `Cargo.toml`
+pub trait WorkspaceInherit {
+    /// This is the workspace table that is being inherited from.
+    /// For example `[workspace.dependencies]` would be the table "dependencies"
+    fn inherit_toml_table(&self) -> &str;
+
+    /// This is used to output the value of the implementors `workspace` field
```
mkaput, line 15:
```suggestion
    /// This is used to output the value of the implementors `workspace` field.
```

scarb/src/core/manifest/maybe_workspace.rs
```diff
@@ -0,0 +1,76 @@
...
+
+/// This Trait exists to make [`MaybeWorkspace::Workspace`] generic. It makes deserialization of
+/// [`MaybeWorkspace`] much easier, as well as making error messages for
+/// [`MaybeWorkspace::resolve`] much nicer
+///
+/// Implementors should have a field `workspace` with the type of `bool`. It is used to ensure
+/// `workspace` is not `false` in a `Cargo.toml`
+pub trait WorkspaceInherit {
+    /// This is the workspace table that is being inherited from.
+    /// For example `[workspace.dependencies]` would be the table "dependencies"
```
mkaput, line 12:
```suggestion
    /// For example `[workspace.dependencies]` would be the table "dependencies".
```

mkaput (changes requested):

scarb/src/core/workspace.rs
```diff
@@ -1,31 +1,52 @@
...
-/// **Note:** Currently only single-package workspaces are supported.
-///
 /// A workspace is often created very early on and then threaded through all other functions.
 /// It's typically through this object that the current package is loaded and/or learned about.
 pub struct Workspace<'c> {
     config: &'c Config,
-    package: Package,
+    members: BTreeMap<PackageId, Package>,
+    root_package: Option<PackageId>,
+    manifest_path: Utf8PathBuf,
```
mkaput, line 24: shouldn't we store manifest here?

scarb/src/core/workspace.rs
```diff
@@ -52,12 +75,30 @@ impl<'c> Workspace<'c> {
     /// In this case an error is returned indicating that the operation
     /// must be performed on specific package.
     pub fn current_package(&self) -> Result<&Package> {
-        Ok(&self.package)
+        Ok(self.members.values().next().unwrap())
+    }
+
+    pub fn root_package(&self) -> Option<Package> {
+        self.root_package.and_then(|id| self.package(&id)).cloned()
+    }
```
mkaput, line 83: what does this represent? the name is not self-descriptive unfortunately

scarb/src/core/workspace.rs
```diff
@@ -52,12 +75,30 @@ impl<'c> Workspace<'c> {
...
+
+    pub fn package(&self, id: &PackageId) -> Option<&Package> {
+        self.members.get(id)
+    }
+
+    pub fn fetch_package(&self, id: &PackageId) -> Result<&Package> {
+        self.package(id)
+            .ok_or_else(|| anyhow::anyhow!("package `{}` is not a member of this workspace", id))
```
mkaput, line 91:
```suggestion
            .ok_or_else(|| anyhow!("package `{}` is not a member of this workspace", id))
```

scarb/src/core/workspace.rs
```diff
@@ -1,31 +1,52 @@
...
-    package: Package,
+    members: BTreeMap<PackageId, Package>,
+    root_package: Option<PackageId>,
+    manifest_path: Utf8PathBuf,
 }
 
 impl<'c> Workspace<'c> {
+    pub(crate) fn new(
+        manifest_path: Utf8PathBuf,
+        packages: Vec<Package>,
```
mkaput, line 30: Please document that the first package will be the "current" one

mkaput, line 30:
```suggestion
        packages: &[Package],
```

scarb/src/core/workspace.rs
```diff
@@ -1,31 +1,52 @@
...
+        root_package: Option<PackageId>,
+        config: &'c Config,
+    ) -> Result<Self> {
+        let packages = packages
+            .iter()
+            .map(|p| (p.id, p.clone()))
+            .collect::<Vec<_>>();
```
mkaput, line 37:
```suggestion
            .collect::<BTreeMap<_, _>>();
```

mkaput (changes requested):

scarb/src/core/workspace.rs
mkaput: Oh I didn't know that feature. Yeah, the comment is enough

mkaput (changes requested):

mkaput (changes requested):

scarb/src/ops/workspace.rs
```diff
@@ -29,23 +34,140 @@ pub fn read_workspace_with_source_id<'c>(
...
+                        .map(PathBuf::try_into_utf8)
+                })
+                .collect::<Result<Result<Vec<_>, _>>>()?
+        })
+        .collect::<Result<Vec<_>>>()
+        .map(|v| {
+            v.into_iter()
+                .flatten()
+                // Make sure all files exist.
+                .filter(|p| p.is_file())
```
mkaput, line 164: I wonder if we shouldn't print a warning if we matched a directory that is missing Scarb.toml

Technically, I think we should make users avoid overusing globs in this way? What do you think?

mkaput (approved):

scarb/src/core/workspace.rs
mkaput: bump

scarb/src/ops/workspace.rs
maciektr: Hmm, do you think a warning is enough, or we should just error?

maciektr: Afaik Cargo does emit errors for such cases, but they also have `workspace.exclude` for better control over the search.

mkaput: warning should be enough I think

workspace.exclude is something we can leave in our backlog :)

mkaput (approved):

scarb/src/ops/workspace.rs
mkaput: this change can be done in separate PR, resolving this discussion

scarb/src/core/workspace.rs
maciektr: I feel like `core::manifest::Manifest` is really concerned with being the **package manifest**, defining targets, dependencies etc. We could create some `WorkspaceManifest` representation, although I do not think we need it **yet**. All the values defined for `.workspace = true` manifest inheritance can be merged when deserializing package manifests (like we do as of this pr), without loosing any functionality.

mkaput: Makes sense

mkaput (approved):

maciektr:
✓ Commit merged in pull request [#446](https://github.com/software-mansion/scarb/pull/446)

PR #446, opened 2023-06-30
Add workspace example

maciektr:
commit-id:214ce15e

---

**Stack**:
- #508
- #506
- #502
- #500
- #490
- #489
- #465
- #446 ⬅
- #447
- #445

### DIFF
--- a/examples/hello_workspaces/Scarb.toml
+++ b/examples/hello_workspaces/Scarb.toml
@@ -1,0 +1,24 @@
+[workspace]
+members = [
+    "crates/**",
+]
+
+[workspace.scripts]
+test = "snforge"
+
+[workspace.dependencies]
+starknet = ">=2.1.0-rc0"
+
+[package]
+name = "hello_workspaces"
+version = "0.1.0"
+
+[scripts]
+test.workspace = true
+
+[dependencies]
+starknet.workspace = true
+fibonacci = { path = "crates/fibonacci" }
+addition = { path = "crates/addition" }
+
+[[target.starknet-contract]]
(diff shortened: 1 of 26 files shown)


examples/hello_workspaces/Scarb.toml
```diff
@@ -0,0 +1,22 @@
+[workspace]
+members = [
+    "crates/**",
+]
+
+[workspace.scripts]
+test = "protostar --version"
+
+[workspace.dependencies]
+starknet = ">=2.0.0-rc0"
```
mkaput, line 10: We're post rc phase in 2.0.0 release line, please revert back to standard version specs

examples/hello_workspaces/Scarb.toml
```diff
@@ -0,0 +1,22 @@
+[workspace]
+members = [
+    "crates/**",
```
mkaput, line 3: does Cargo support globs here? 😱

examples/hello_workspaces/Scarb.toml
mkaput, line 1: This example kinda seems to overfocus on specific features instead of giving users most basic glimpse of what is possible. I imagine having this example as a test case rather something I'd show newcomers

mkaput (changes requested):

examples/hello_workspaces/Scarb.toml
maciektr: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

mkaput: first time I see it live actually, interesting. let's support them then if you did the work

PR #447, opened 2023-06-30
Add workspace loading test case

maciektr:
**Stack**:
- #508
- #506
- #502
- #500
- #490
- #489
- #465
- #446
- #447 ⬅
- #445

### DIFF
--- a/scarb/tests/build.rs
+++ b/scarb/tests/build.rs
@@ -9,6 +9,7 @@ use scarb_build_metadata::CAIRO_VERSION;
 use scarb_test_support::command::Scarb;
 use scarb_test_support::fsx::ChildPathEx;
 use scarb_test_support::project_builder::ProjectBuilder;
+use scarb_test_support::workspace_builder::WorkspaceBuilder;
 
 #[test]
 fn compile_simple() {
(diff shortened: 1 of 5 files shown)


scarb/tests/e2e/support/workspace_builder.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    pub fn add_member(mut self, member: impl Into<String>) -> Self {
+        self.members.push(member.into());
+        self
+    }
+
+    pub fn package(mut self, package: ProjectBuilder) -> Self {
+        self.package = Some(package);
+        self
+    }
```
mkaput, line 25: This seems like a kinda rare, single test, usage scenario. Can't we just use `extra(&str)` pattern that we have in ProjectBuilder? It should simplify much of this code as far as I see.

mkaput (changes requested):

scarb/tests/e2e/support/workspace_builder.rs
maciektr: Hmm, I am not sure about this one. 
Root packages can be tricky sometimes, as their location in workspace root can cause some weird edge-cases (e.g. our offline discussion about fmt), so naturally we may want more than one test case. 
Also, eradicating this field would make us to create package sources, pass the package manifest, sometimes also concatenated with other manifest extras on our own, which may be inconvenient at times.

mkaput: ok, convinced

mkaput (approved):

mkaput (approved):

maciektr:
✓ Commit merged in pull request [#446](https://github.com/software-mansion/scarb/pull/446)

PR #459, opened 2023-07-04
Use `target_name` instead of `package_name` for `starknet_artifacts` map file

maciektr:
- Consistent with starknet artifacts naming convention
- Avoids name collisions (target names are guaranteed to be unique across all units)


mkaput (approved):
Good catch! Let's merge this, but we need not forget to describe this change in release notes. It's technically breaking, though we never committed to file names here anyway...

PR #474, opened 2023-07-13
Use mold linker to speed up CI compile times

mkaput:


mkaput:
This seems not to help us :(

PR #513, opened 2023-07-31
Add compile time check for Core library and Cairo compiler versions

### LINKED ISSUE #456: Add compile time (build.rs) checks for core library version matching

MrDenkoV:
Fix #456

### DIFF
--- a/scarb/build.rs
+++ b/scarb/build.rs
@@ -5,7 +5,7 @@ use std::{env, fs, io};
 
 use zip::ZipArchive;
 
-use scarb_build_metadata::CAIRO_COMMIT_REV;
+use scarb_build_metadata::{CAIRO_COMMIT_REV, CAIRO_VERSION};
 
 fn main() {
     download_core(CAIRO_COMMIT_REV);
@@ -81,6 +81,7 @@ fn download_core(rev: &str) {
         }
     }
 
+    check_corelib_version(core_path.join("Scarb.toml"));
     println!("cargo:rustc-env=SCARB_CORE_PATH={}", core_path.display());
 }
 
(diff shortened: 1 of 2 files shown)


scarb/build.rs
```diff
@@ -81,6 +81,34 @@ fn download_core(rev: &str) {
         }
     }
 
+    let ensure_err = "\nUnable to ensure Core library and Cairo Compiler versions match!\
```
mkaput, line 84: this line + usinc `panic!` directly is quite verbose. You can easily go with https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.expect and say _Could not read this_, _Could not get field Y_ etc.

mkaput (changes requested):

scarb/build.rs
```diff
@@ -91,3 +92,20 @@ fn ident(id: &str) -> String {
     }
     ident
 }
+
+fn check_versions(comp_ver: &str, path: PathBuf) {
```
mkaput, line 96:
```suggestion
fn check_corelib_version(core_scarb_toml: PathBuf) {
```

This function is `core`-specific anyway, so no need to overly generalise it

scarb/build.rs
```diff
@@ -91,3 +92,20 @@ fn ident(id: &str) -> String {
...
+    let core_toml: toml::Value =
+        toml::from_str(&fs::read_to_string(path).expect("Could not read Core library Scarb.toml"))
```
mkaput, line 98:
```suggestion
        toml::from_str(&fs::read_to_string(path).expect("Could not read `core` Scarb.toml."))
```

nit: that's how we usually refer to this package in messages :)
nit2: We use "Full sentences with dot." for panics and "lower case without dot" for errors

mkaput (approved):
Looks good! Please resolve my nitpicks and this is good to go! 🔥

scarb/build.rs
```diff
@@ -91,3 +92,21 @@ fn ident(id: &str) -> String {
...
+    let core_ver = core_toml
+        .get("package")
+        .expect("Could not get package section from `core` Scarb.toml.")
+        .get("version")
+        .expect("Could not get version field from `core` Scarb.toml.")
+        .as_str()
+        .unwrap();
+    assert_eq!(
+        core_ver, CAIRO_VERSION,
+        "\n`core` version does not match Cairo Compiler version!"
```
mkaput, line 110:
```suggestion
        "`core` version does not match Cairo compiler version."
```

mkaput (approved):

PR #523, opened 2023-08-01
Make build tests not fail if they run too long

mkaput:
https://github.com/software-mansion/scarb/actions/runs/5697461654/job/15444219731


maciektr (approved):
: joy:

PR #533, opened 2023-08-04
Add fetch to docs

maciektr:


mkaput (changes requested):
I don't think `fetch` is such a command worthy including in guides. They should cover the most important commands like `build`, `add`, `metadata` etc. and minor utility stuff like `fetch`, `manifest-path` is always discoverable via `--help`.

PR #543, opened 2023-08-08
Add Scarb command abstraction, build Scarb project before run

### LINKED ISSUE #526: Automatically call `scarb build` before `scarb cairo-run`
### Problem

People forget to run build before calling cairo-run and thus they end up accidentally running older builds.

### Proposed Solution

Call `scarb build` at the beginning of `cairo-run`. We need to make a nice API for doing so as part of `scarb-metadata` crate I think.

### Notes

_No response_

maciektr:
fix #526

### DIFF
--- a/extensions/scarb-cairo-run/src/main.rs
+++ b/extensions/scarb-cairo-run/src/main.rs
@@ -13,7 +13,7 @@ use clap::Parser;
 use indoc::formatdoc;
 
 use scarb_metadata::packages_filter::PackagesFilter;
-use scarb_metadata::MetadataCommand;
+use scarb_metadata::{MetadataCommand, ScarbCommand};
 
 /// Execute the main function of a package.
 #[derive(Parser, Clone, Debug)]
@@ -38,6 +38,8 @@ fn main() -> Result<()> {
 
     let package = args.packages_filter.match_one(&metadata)?;
     println!("running {} ...", package.name);
+
+    build_project()?;
 
     let filename = format!("{}.sierra", package.name);
     let path = Utf8PathBuf::from(env::var("SCARB_TARGET_DIR")?)
@@ -123,4 +125,8 @@ fn main() -> Result<()> {
     }
 
     Ok(())
+}
+
+fn build_project() -> Result<()> {
+    Ok(ScarbCommand::new().arg("build").run()?)
 }
--- a/scarb-metadata/src/command/scarb_command.rs
+++ b/scarb-metadata/src/command/scarb_command.rs
@@ -1,0 +1,113 @@
+use std::ffi::OsStr;
+use std::io;
+use std::path::PathBuf;
+
+use crate::command::internal_command::InternalScarbCommandBuilder;
+use thiserror::Error;
+
+/// Error thrown while trying to execute `scarb` command.
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum ScarbCommandError {
+    /// Failed to read `scarb` output.
+    #[error("failed to read `scarb` output")]
+    Io(#[from] io::Error),
+    /// Error during execution of `scarb` command.
+    #[error("`scarb metadata` exited with error")]
+    ScarbError,
+}
+
+/// A builder for `scarb` command invocation.
+#[derive(Clone, Debug, Default)]
+pub struct ScarbCommand {
+    inner: InternalScarbCommandBuilder,
+}
+
+impl ScarbCommand {
+    /// Creates a default `scarb` command, which will look for `scarb` in `$PATH` and
+    /// for `Scarb.toml` in the current directory or its ancestors.
+    pub fn new() -> Self {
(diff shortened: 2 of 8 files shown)


scarb-metadata/src/scarb_command.rs
```diff
@@ -0,0 +1,210 @@
...
+        /// Captured standard output if any.
+        stdout: String,
+        /// Captured standard error if any.
+        stderr: String,
+    },
+}
+
+/// A builder for `scarb` command invocation.
+#[derive(Clone, Debug, Default)]
+pub struct ScarbCommandBuilder {
```
mkaput, line 28: I don't think anybody will want to use this to collect Scarb output, neither we should promote doing this. The only command that we expect people to read is `scarb metadata` which is covered by MetadataCommand API already. What we want here is just an ability to shell out to Scarb, always inheriting stdout and stderr.

So I think what we pursue here is something simpler, like:

```rust
ScarbCommand::new("build")
    .arg(...)
    .current_dir(...)
    .run()? -> Result<(), ScarbCommandError>
```

I know there is JSON output option everywhere, but reading it is exotic case enough I'd rather ask people to use `std::command::Command`

mkaput (changes requested):

scarb-metadata/src/scarb_command.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Clears the entire environment map for the child process.
+    pub fn env_clear(&mut self) -> &mut Self {
+        self.inner.env_clear();
+        self
+    }
+
+    /// Runs configured `scarb` command.
+    pub fn run(&self) -> Result<(), ScarbCommandError> {
+        let mut cmd = self.inner.command();
+        let output = cmd.output()?;
```
mkaput, line 107: https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.status

scarb-metadata/src/lib.rs
mkaput, line 1: I think you can group all command modules as children of a single module `scarb_metadata::command`.

this way you would have to play less with `#[cfg(...)]`, and also won't have to use `pub(crate)`, because you could put internal command struct within this module

extensions/scarb-cairo-run/src/main.rs
```diff
@@ -124,3 +128,13 @@ fn main() -> Result<()> {
 
     Ok(())
 }
+
+fn build_project(package_name: &String, profile: &String) -> Result<()> {
+    Ok(ScarbCommand::new()
+        .arg("--profile")
+        .arg(profile)
```
mkaput, line 135:
```suggestion
```

This should be derived via environment

extensions/scarb-cairo-run/src/main.rs
```diff
@@ -124,3 +128,13 @@ fn main() -> Result<()> {
...
+        .arg("build")
+        .arg("--package")
+        .arg(package_name)
```
mkaput, line 138:
```suggestion
```

This should be derived via environment

scarb-metadata/src/command.rs
```diff
@@ -112,36 +104,39 @@ impl MetadataCommand {
...
-        self.inherit_stderr = true;
+        self.inner.inherit_stderr();
         self
     }
 
+    fn scarb_command(&self) -> Command {
+        let mut builder = self.inner.clone();
+        builder.json();
+        builder.args(["metadata", "--format-version"]);
+        builder.arg(VersionPin.numeric().to_string());
```
mkaput, line 133: This will break if somebody adds an arg that is specific to the `metadata` command. It will land before `metadata` command name, while it should be afterwards

mkaput (changes requested):

scarb-metadata/src/command.rs
maciektr: Discussed offline

scarb-metadata/src/command/scarb_command.rs
```diff
@@ -0,0 +1,113 @@
...
+use crate::command::internal_command::InternalScarbCommandBuilder;
+use thiserror::Error;
+
+/// Error thrown while trying to execute `scarb` command.
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum ScarbCommandError {
+    /// Failed to read `scarb` output.
+    #[error("failed to read `scarb` output")]
+    Io(#[from] io::Error),
```
mkaput, line 14: unused?

mkaput (changes requested):

scarb-metadata/src/command/scarb_command.rs
maciektr:
```
104 |         if cmd.status()?.success() {
    |                        ^ the trait `From<std::io::Error>` is not implemented for `ScarbCommandError`
```

mkaput (approved):

Issue #558, opened 2023-08-11
Add `scarb cache path` command


mkaput:
### Problem

Currently, the only way to know where Scarb puts its cache files is to look at [relevant Scarb documentation](https://docs.swmansion.com/scarb/docs/reference/global-directories.html) and manually implement if-else chain based on running platform. For example, that's [what `setup-scarb` action does](https://github.com/software-mansion/setup-scarb/blob/608d5f20f2dd942626f7f618a312536bc7ffeaa2/lib/cache-utils.js#L13-L25). It would be much easier for everyone if global paths could be accessible programmatically from `scarb` CLI.

### Proposed Solution

Add a new command `scarb cache path` (residing next to `scarb cache clean`) which will just `println!` [cache path](https://github.com/software-mansion/scarb/blob/06c4dc21cc0b80274ee8d8374ee5a3b0ed10e17c/scarb/src/core/dirs.rs#L15) from `Config`.

CLI help message for this command: `Print path to Scarb's global cache directory`

### Notes

- For extra reference, have a look at `yarn cache dir`.
- Implementation is very trivial, please add a test which just ensures this command succeeds and outputs a single line.
- [Bonus points for integrating this into `setup-scarb`](https://github.com/software-mansion/setup-scarb/issues/11)

PR #566, opened 2023-08-12
added cache-path command

ametel01:
solves #558

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -115,6 +115,8 @@ impl ScarbArgs {
 pub enum CacheSubcommand {
     /// Remove all cached dependencies.
     Clean,
+    /// Print the path of the cache directory.
+    Path,
 }
 
 /// Subcommand and its arguments.
--- a/scarb/src/bin/scarb/commands/cache_path.rs
+++ b/scarb/src/bin/scarb/commands/cache_path.rs
@@ -1,0 +1,10 @@
+use anyhow::{Ok, Result};
+
+use scarb::{core::Config, ui::ValueMessage};
+
+#[tracing::instrument(skip_all, level = "info")]
+pub fn run(config: &Config) -> Result<()> {
+    let path = config.dirs().cache_dir.path_unchecked();
+    config.ui().print(ValueMessage::new("path", &path));
+    Ok(())
+}
--- a/scarb/src/bin/scarb/commands/mod.rs
+++ b/scarb/src/bin/scarb/commands/mod.rs
@@ -9,6 +9,7 @@ use crate::args::{CacheSubcommand, Command};
 pub mod add;
 pub mod build;
 pub mod cache_clean;
+pub mod cache_path;
 pub mod clean;
 pub mod commands;
 pub mod external;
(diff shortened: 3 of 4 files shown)


scarb/src/bin/scarb/args.rs
```diff
@@ -134,6 +136,8 @@ pub enum Command {
     Cache(CacheSubcommand),
     /// Remove generated artifacts.
     Clean,
+    /// Print cache path.
+    CachePath,
```
maciektr, line 140: Hey! `scarb cache clean` seems enough, I do not think we need an alias for that here.

scarb/src/bin/scarb/commands/cache_path.rs
```diff
@@ -0,0 +1,9 @@
+use anyhow::Result;
+use scarb::core::Config;
+
+use scarb::ops;
+
+#[tracing::instrument(skip_all, level = "info")]
+pub fn run(config: &Config) -> Result<()> {
+    ops::cache_path(config)
```
maciektr, line 8: Similarly to https://github.com/software-mansion/scarb/blob/main/scarb/src/bin/scarb/commands/manifest_path.rs, I do not think we need to move this to `ops`.

scarb/src/ops/cachepath.rs
```diff
@@ -0,0 +1,12 @@
+use anyhow::Result;
+
+use crate::core::Config;
+
+#[tracing::instrument(skip_all, level = "debug")]
+pub fn cache_path(config: &Config) -> Result<()> {
+    let path = config.dirs().cache_dir.path_unchecked();
+    if path.exists() {
```
maciektr, line 8: I think we should output the path even if it is not existing in the filesystem yet.

maciektr (commented):
Would you mind adding a test case for this?

scarb/src/bin/scarb/args.rs
ametel01: So we don't need a `scarb cache-path` cli command?

scarb/src/bin/scarb/args.rs
```diff
@@ -134,7 +136,7 @@ pub enum Command {
     Cache(CacheSubcommand),
     /// Remove generated artifacts.
     Clean,
-    /// List installed commands.
+    /// Print cache path.
```
maciektr, line 139: Can you please revert the change in this line? :D

scarb/src/bin/scarb/commands/cache_path.rs
```diff
@@ -0,0 +1,10 @@
+use anyhow::{Ok, Result};
+
+use scarb::core::Config;
+
+#[tracing::instrument(skip_all, level = "info")]
+pub fn run(config: &Config) -> Result<()> {
+    let path = config.dirs().cache_dir.path_unchecked();
+    println!("{path}");
```
maciektr, line 8: I've noticed one more thing, sorry for not mentioning this before. 

In Scarb we use our own UI abstraction for printing text to users (https://github.com/software-mansion/scarb/blob/main/scarb/src/ui/mod.rs). 
This also allows us to structure all messages to json if needed (with `--json` arg). 

Please print with `config.ui()` here (as below).

```suggestion
    config.ui().print(ValueMessage::new("path", &path));
```

maciektr (commented):
Great stuff! 
Would you mind rebasing into a single commit after all changes?

maciektr (approved):
👏 👏 👏 
Looks great, thank you ❤️ !

PR #570, opened 2023-08-15
Add OZ installation example

martriay:
This PR adds an example of how to properly install the OZ Contracts for Cairo library, while showing how to use the `tag` parameter for dependencies.

This PR is better merged by next week, when the OZ release tag is live.


mkaput (approved):
❤️❤️❤️❤️❤️ 

@martriay can you please ping me when the tag will be ready? We'll merge this and publish immediately 😄

PR #571, opened 2023-08-16
Add support for installing nightlies in install script

### LINKED ISSUE #538: Support nightlies through installation script
Make it possible to install nightlies version using [Scarb installation script](https://docs.swmansion.com/scarb/download#install-via-installation-script)

szymmis:
Resolves #538

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -14,6 +14,7 @@
 set -u
 
 SCARB_REPO="https://github.com/software-mansion/scarb"
+SCARB_NIGHTLIES_REPO="https://github.com/software-mansion/scarb-nightlies"
 XDG_DATA_HOME="${XDG_DATA_HOME:-"${HOME}/.local/share"}"
 INSTALL_ROOT="${XDG_DATA_HOME}/scarb-install"
 LOCAL_BIN="${HOME}/.local/bin"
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -400,12 +399,27 @@ get_architecture() {
...
+    else
+      _ref="tag/v${_requested_version}"
+    fi
+  fi
 
   local _response
 
-  say "retrieving $_requested_version version from ${SCARB_REPO}..."
-  _response=$(ensure curl -Ls -H 'Accept: application/json' "${SCARB_REPO}/releases/${_requested_ref}")
+  say "${_repo}/releases/${_ref}"
```
maciektr, line 420: Just making sure, is this intentional?

mkaput: Yeah, I don't think so

mkaput (commented):
Sanity check: have you tested all of this manually? :)

install.sh
szymmis: It's good to have a second pair of eyes 😅, totally unintentional, thanks for the catch!

install.sh
```diff
@@ -488,4 +526,4 @@ add_local_bin_to_path() {
     "Run 'source ${_profile}' or start a new terminal session to use Scarb."
 }
 
-main "$@" || exit 1
+main "$@" || exit 1
```
mkaput, line 529:
```suggestion
main "$@" || exit 1

```

mkaput (commented):
Please also run `checkbashisms install.sh`. We technically do not target bash here :)

mkaput (approved):

PR #587, opened 2023-08-18
Basic implementation of `scarb-ui` in `scarb cairo-run`

mkaput:
Extra: reordered methods in `Message` trait to group them by medium, so that linters will not yell, as this new order makes more sense.

JSON support to be included in later commits.

### DIFF
--- a/extensions/scarb-cairo-run/src/main.rs
+++ b/extensions/scarb-cairo-run/src/main.rs
@@ -1,9 +1,9 @@
 use std::env;
 use std::fs;
 
-use anyhow::{anyhow, bail, Context, Result};
+use anyhow::{anyhow, bail, ensure, Context, Result};
 use cairo_lang_runner::short_string::as_cairo_short_string;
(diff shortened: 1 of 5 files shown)


extensions/scarb-cairo-run/src/main.rs
```diff
@@ -34,24 +37,29 @@ struct Args {
 
 fn main() -> Result<()> {
     let args: Args = Args::parse();
+
+    let ui = Ui::new(Verbosity::default(), OutputFormat::Text);
```
maciektr, line 41: Does it take `SCARB_UI_VERBOSITY` into consideration?

mkaput: Unfortunately not, it's something I/somebody will have to add later. I'll make an issue about this, as I already forgot 😅

mkaput: https://github.com/software-mansion/scarb/issues/592

maciektr (approved):

Issue #577, opened 2023-08-17
Allow building all contracts from dependency package at once


mkaput:
### Problem

Some projects grew quite sizeable groups of contracts to include in `build-external-contracts` listing in their users. For instance, in Dojo, users have to include this behemoth in their projects:

```toml
[[target.dojo]]
build-external-contracts = [
    "dojo_erc::erc721::components::Balance",
    "dojo_erc::erc721::components::OperatorApproval",
    "dojo_erc::erc721::components::Owner",
    "dojo_erc::erc721::components::TokenApproval",
    "dojo_erc::erc721::erc721::ERC721",
    "dojo_erc::erc721::systems::erc721_approve",
    "dojo_erc::erc721::systems::erc721_burn",
    "dojo_erc::erc721::systems::erc721_mint",
    "dojo_erc::erc721::systems::erc721_set_approval_for_all",
    "dojo_erc::erc721::systems::erc721_transfer_from",
]
```

Missing one of the contracts here will lead to problems.

### Proposed Solution

Allow the `build-external-contracts` field to allow pulling all contracts from a given package/crate at once, in a glob fashion.

The expected syntax would look like this:

```toml
[[target.dojo]]
build-external-contracts = [
    "dojo_erc::*",
]
```

If this happens to be trivial to implement, it would be cool to also allow importing all contracts prefixed by a module. **But this is not mandatory, and just pulling everything from a crate is sufficient.** For instance:

```toml
[[target.dojo]]
build-external-contracts = [
    "dojo_erc::erc721::*",
]
```

### Notes

- The `dojo` target in this example inherits source code from `starknet-contract`. This task is about adding this logic to `starknet-contract`.
- The following parts of code will require amending:
    - https://github.com/software-mansion/scarb/blob/bf927194941f6c0ce62677e7e2ef4f9122489ff6/scarb/src/compiler/compilers/starknet_contract.rs#L69-L92
    - https://github.com/software-mansion/scarb/blob/bf927194941f6c0ce62677e7e2ef4f9122489ff6/scarb/src/compiler/compilers/starknet_contract.rs#L300-L325
    - ➕ I guess somewhere else
- Don't forget about adding tests!

---

Please leave a comment that you are willing to contribute a PR fixing this issue before starting doing so, in order to avoid having two people working on this independently 😄

mkaput:
@tarrencev what do you think about this proposed syntax (with `*` usage)? Scarb team uniformly likes it :)

[...]

PR #606, opened 2023-08-24
Implement wild card contracts selector

ametel01:
solve #577 

only one test added, do we need to test more complex path declarations?

### DIFF
--- a/scarb/src/compiler/compilers/starknet_contract.rs
+++ b/scarb/src/compiler/compilers/starknet_contract.rs
@@ -2,7 +2,7 @@ use std::collections::HashSet;
 use std::fmt::Write;
 use std::iter::zip;
 
-use anyhow::{bail, Context, Result};
+use anyhow::{bail, ensure, Context, Result};
 use cairo_lang_compiler::db::RootDatabase;
 use cairo_lang_filesystem::ids::{CrateId, CrateLongId};
 use cairo_lang_semantic::db::SemanticGroup;
@@ -28,6 +28,7 @@ use crate::internal::serdex::RelativeUtf8PathBuf;
 use crate::internal::stable_hash::short_hash;
 
 const CAIRO_PATH_SEPARATOR: &str = "::";
+const GLOB_PATH_SELECTOR: &str = "*";
 
 // TODO(#111): starknet-contract should be implemented as an extension.
 pub struct StarknetContractCompiler;
@@ -84,6 +85,18 @@ impl ContractSelector {
             .rsplit_once(CAIRO_PATH_SEPARATOR)
             .unwrap_or((self.0.as_str(), ""));
         parts.1.to_string()
+    }
+
+    fn is_wildcard(&self) -> bool {
+        self.0.ends_with(GLOB_PATH_SELECTOR)
+    }
+
+    fn partial_path(&self) -> String {
+        let parts = self
+            .0
+            .split_once(GLOB_PATH_SELECTOR)
+            .unwrap_or((self.0.as_str(), ""));
+        parts.0.to_string()
     }
 
     fn full_path(&self) -> String {
@@ -203,6 +216,12 @@ impl Compiler for StarknetContractCompiler {
             props.build_external_contracts.clone(),
         )?;
 
+        if let Some(external_contracts) = props.build_external_contracts.clone() {
+            for path in external_contracts.iter() {
+                ensure!(path.0.matches(GLOB_PATH_SELECTOR).count() <= 1, format!("external contract path {} has multiple global path selectors, only one '*' selector is allowed", path.0));
+            }
+        }
+
         let contract_paths = contracts
             .iter()
             .map(|decl| decl.module_id().full_path(db.upcast_mut()))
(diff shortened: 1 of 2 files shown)


scarb/src/compiler/compilers/starknet_contract.rs
```diff
@@ -297,32 +306,40 @@ fn find_project_contracts(
...
+                .collect::<Vec<_>>();
+            let contracts = find_contracts(db, crate_ids.as_ref());
+            let filtered_contracts: Vec<ContractDeclaration> = contracts
+                .into_iter()
+                .filter(|decl| {
                     let contract_path = decl.module_id().full_path(db.upcast());
-                    contract_path == selector.full_path()
+                    external_contracts.iter().any(|selector| {
+                        if selector.is_wildcard() {
+                            contract_path.starts_with(&selector.package_path())
```
maciektr, line 330: I do not think this logic is correct. 
As of now, selector with `some::cairo::path::*` would match all contracts in `some` package.

scarb/src/compiler/compilers/starknet_contract.rs
```diff
@@ -78,6 +78,15 @@ impl ContractSelector {
         PackageName::new(parts.0)
     }
 
+    fn is_wildcard(&self) -> bool {
+        self.0.ends_with("::*")
```
maciektr, line 82: Please refer to `::` as `CAIRO_PATH_SEPARATOR` wherever possible.

scarb/src/compiler/compilers/starknet_contract.rs
```diff
@@ -78,6 +78,15 @@ impl ContractSelector {
...
+    }
+
+    fn package_path(&self) -> String {
+        let parts = self.0.split_once("::").unwrap_or((self.0.as_str(), ""));
```
maciektr, line 86: It's equivalent to `.package`.

scarb/src/compiler/compilers/starknet_contract.rs
maciektr: Please also add a test case for this scenario.

scarb/src/compiler/compilers/starknet_contract.rs
ametel01: implemented GLOB_PATH_SELECTOR const

scarb/src/compiler/compilers/starknet_contract.rs
ametel01: change package_path to partial_path selector with modified logic

scarb/src/compiler/compilers/starknet_contract.rs
ametel01: added GLOB_PATH_SEPARATOR const

scarb/src/compiler/compilers/starknet_contract.rs
ametel01: removed

mkaput:
> cargo fmt --check passing locally not sure why not on CI

Rust 1.72 got released and it changed formatting a bit

ametel01:
> > cargo fmt --check passing locally not sure why not on CI
> 
> Rust 1.72 got released and it changed formatting a bit
Updated rust and now all good

maciektr (commented):
Great work! 🔥 🔥 🔥 

One last thing to consider: 
Can we also verify that the path selector contains no more than a single `*`?
Consider emitting an Err result with https://docs.rs/anyhow/latest/anyhow/macro.ensure.html otherwise.

ametel01:
> Great work! 🔥 🔥 🔥
> 
> One last thing to consider: Can we also verify that the path selector contains no more than a single `*`? Consider emitting an Err result with https://docs.rs/anyhow/latest/anyhow/macro.ensure.html otherwise.

What kind of behaviour do you want in case ensure! doesn't pass? return anyhow::Error or log a message and continue with the function?

maciektr:
@ametel01 Since we implement prefixes only, I think we should ensure that the package selector has exactly one `*` or none and that it is the last character, as otherwise behaviour of `some::path::*::with::two::*` may be confusing to users. This should be an error.

ametel01:
I’m unsure where to put the check for ‘*’ selector is the last character of the path since since the is_wildcard() only returns true when ‘*’ is last.

scarb/src/compiler/compilers/starknet_contract.rs
```diff
@@ -291,38 +305,53 @@ fn find_project_contracts(
...
+                            }
+                        } else {
+                            contract_path == selector.full_path()
+                        }
+                    });
+                    if is_matched {
+                        Some(Ok::<ContractDeclaration, anyhow::Error>(decl))
+                    } else {
+                        None
+                    }
```
maciektr, line 347: I would rather show the parsing error to the user, so he can adjust his config appropriately. 

My suggestion would be, to validate the struct `Prop` which deserializes this value. 
https://github.com/software-mansion/scarb/blob/0db29e2b584b624a3f0fc4b8026d2aae3dfc0113/scarb/src/compiler/compilers/starknet_contract.rs#L186C9-L186C9
This way we could check all the selectors once,  and show the errors to the user. 

We can also have a simple test case with invalid selector to verify the validation.

scarb/src/compiler/compilers/starknet_contract.rs
```diff
@@ -86,6 +87,19 @@ impl ContractSelector {
         parts.1.to_string()
     }
 
+    fn is_wildcard(&self) -> bool {
+        self.0.ends_with(GLOB_PATH_SELECTOR)
+    }
+
+    fn partial_path(&self) -> Result<String, anyhow::Error> {
+        ensure!(self.0.matches(GLOB_PATH_SELECTOR).count() <= 1, format!("External dependency {} has multiple global path selectors; only one '*' selector is allowed.", self.0));
```
maciektr, line 95: Nit
```suggestion
        ensure!(self.0.matches(GLOB_PATH_SELECTOR).count() <= 1, format!("external contract path {} has multiple global path selectors, only one '*' selector is allowed", self.0));
```

maciektr (commented):
Hi!

Left one more small comment. 
 
Sorry for responding so late to your changes, I has been ooo for most of the week. 
If you have moved on to another challange already, please let me know.

scarb/src/compiler/compilers/starknet_contract.rs
ametel01: removed all this, now ensure is enforced on every path of the external-contracts in `compile`

scarb/src/compiler/compilers/starknet_contract.rs
ametel01: fixed

maciektr (approved):
This looks great! Thank you a lot for your help!

♥️♥️♥️♥️♥️

maciektr:
Would you mind squashing the commits so I can merge this?

ametel01:
> Would you mind squashing the commits so I can merge this?

Unable to squash commits, likely due to merging `main` into this branch.

mkaput:
Hmm, I think interactive rebasing on top of main should purge the merge commit and give you an ability to squash these comments. Nonetheless, I think I can squash merge this manually if you all like :)

ametel01:
> Hmm, I think interactive rebasing on top of main should purge the merge commit and give you an ability to squash these comments. Nonetheless, I think I can squash merge this manually if you all like :)

I attempted interactive rebase to remove the merge commit and squash the changes, but it complicated things further. Could you handle the squash merge on your end? Thanks!

ametel01:
I think i managed somehow.

mkaput:
OK, I will squash merge this anyway, so that we'll have a more descriptive commit message 😅 I'm sorry for mental shortcuts here; we generally use rebase merges & do as much as possible to maintain a linear and bisectable commit history 😄 But no worries, sorry for bothering.

I'm super thankful for this work and I'm happy to merge it!

❤️❤️❤️❤️❤️

PR #610, opened 2023-08-24
Use SerDe to save sierra program

### LINKED ISSUE #510: Replace textual Sierra representation with JSON serialization of the Sierra struct
Currently, when building a `[lib]` target we rely on sierra textual representation. This representation can then be used as an intermediate artifact, e.g. for `cairo-run` extension. This imposes a need to rely on the textual representation parser which is deemed to not be production grade, which may induce hard to trace errors. Instead we should serialize the sierra program to some serde format, that can serve as a "_runnable_" representation.

MrDenkoV:
Fix #510


extensions/scarb-cairo-run/src/main.rs
```diff
@@ -61,13 +62,13 @@ fn main() -> Result<()> {
...
-            &fs::read_to_string(path.clone())
-                .with_context(|| format!("failed to read Sierra file: {path}"))?,
-        )
-        .map_err(|e| anyhow!("{e}"))
-        .with_context(|| format!("failed to parse sierra program: {path}"))?;
+    let sierra_program = serde_json::from_str::<VersionedProgram>(
+        &fs::read_to_string(path.clone())
+            .with_context(|| format!("failed to read Sierra file: {path}"))?,
+    )
+    .map_err(|e| anyhow!("{e}"))
```
maciektr, line 69: Do we need this line? It was added because of the way that lalrpop has constructed error structs.

extensions/scarb-cairo-run/src/main.rs
```diff
@@ -61,13 +62,13 @@ fn main() -> Result<()> {
...
+    .with_context(|| format!("failed to deserialize sierra versioned program: {path}"))?
```
mkaput, line 70:
```suggestion
    .with_context(|| format!("failed to deserialize Sierra program: {path}"))?
```

that was leaking implementation details to the user, also casing

extensions/scarb-cairo-run/src/main.rs
mkaput:
```suggestion
```

scarb/tests/build.rs
```diff
@@ -528,10 +528,16 @@ fn sierra_replace_ids() {
         .assert()
         .success();
 
-    t.child("target/dev/hello.sierra")
-        .assert(predicates::str::contains(
-            "hello::example@0() -> (felt252);",
-        ));
+    t.child("target/dev/hello.sierra").assert(
```
mkaput, line 531: We need to discuss the extension with Cairo guys and ensure everybody likes it. Can you please start the conversation on our shared channel @MrDenkoV?

scarb/src/compiler/compilers/lib.rs
```diff
@@ -60,11 +61,15 @@ impl Compiler for LibCompiler {
 
         if props.sierra {
```
mkaput, line 62: We need to give users an ability to generate textual representation, in case some tooling relies on it. Let's do that in separate task, that should be quite high priority, becase we need to deliver it in the same release as this PR: #634

scarb/src/compiler/compilers/lib.rs
```diff
@@ -60,11 +61,15 @@ impl Compiler for LibCompiler {
...
+                format!("{}.sierra.json", unit.target().name),
                 "output file",
                 ws.config(),
             )?;
-            file.write_all(sierra_program.to_string().as_bytes())?;
+            file.write_all(
+                serde_json::to_string_pretty(&VersionedProgram::from((*sierra_program).clone()))
+                    .context("failed to serialize sierra versioned program")?
+                    .as_bytes(),
+            )?;
```
mkaput, line 72: 1. It's ineffective to generate text to string and then write it to file. You can do that in one operation: https://docs.rs/serde_json/latest/serde_json/fn.to_writer.html
2. `starknet-contract` generates minified JSON. Let's do that here too.

extensions/scarb-cairo-run/src/main.rs
```diff
@@ -61,13 +62,15 @@ fn main() -> Result<()> {
...
-        .with_context(|| format!("failed to parse sierra program: {path}"))?;
+    let sierra_program = serde_json::from_str::<VersionedProgram>(
+        &fs::read_to_string(path.clone())
+            .with_context(|| format!("failed to read Sierra file: {path}"))?,
+    )
+    .with_context(|| format!("failed to deserialize Sierra program: {path}"))?
+    .into_v1()
+    .with_context(|| {
+        format!("failed to parse file into program - incorrect Sierra version: {path}")
+    })?;
```
mkaput, line 73: 1. Using `-` to denote causability in errors seems inconsistent, given we have nice contexts mechanism in Anyhow
2. I think we can rely on `into_v1` function giving us the cause message, so maybe let's just do this:

```suggestion
    .into_v1()
    .with_context(|| format!("failed to load Sierra program: {path}"))?;
```

Cargo.toml
```diff
@@ -28,18 +28,18 @@ repository = "https://github.com/software-mansion/scarb"
...
+cairo-lang-filesystem = { git = "https://github.com/starkware-libs/cairo", rev = "98af2f8" }
+cairo-lang-formatter = { git = "https://github.com/starkware-libs/cairo", rev = "98af2f8" }
+cairo-lang-language-server = { git = "https://github.com/starkware-libs/cairo", rev = "98af2f8" }
+cairo-lang-runner = { git = "https://github.com/starkware-libs/cairo", rev = "98af2f8" }
+cairo-lang-semantic = { git = "https://github.com/starkware-libs/cairo", rev = "98af2f8" }
+cairo-lang-sierra = { git = "https://github.com/starkware-libs/cairo", rev = "98af2f8" }
+cairo-lang-sierra-to-casm = { git = "https://github.com/starkware-libs/cairo", rev = "98af2f8" }
+cairo-lang-starknet = { git = "https://github.com/starkware-libs/cairo", rev = "98af2f8" }
+cairo-lang-test-runner = { git = "https://github.com/starkware-libs/cairo", rev = "98af2f8" }
+cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo", rev = "98af2f8" , features = ["env_logger"] }
```
mkaput, line 42: please extract Cairo update to a separate commit

mkaput (changes requested):

mkaput (commented):
Looks good

mkaput (approved):

PR #611, opened 2023-08-24
Add target kind info to manifest dependency

maciektr:
commit-id:1aba1429

---

**Stack**:
- #651
- #650
- #649
- #629
- #628
- #626
- #627
- #615
- #597
- #612
- #611 ⬅

### DIFF
--- a/scarb/src/core/manifest/dependency.rs
+++ b/scarb/src/core/manifest/dependency.rs
@@ -1,15 +1,54 @@
 use std::fmt;
+use std::ops::Deref;
+use std::sync::Arc;
+use typed_builder::TypedBuilder;
 
-use crate::core::{DependencyVersionReq, PackageId, PackageName, SourceId, Summary};
+use crate::core::{DependencyVersionReq, PackageId, PackageName, SourceId, Summary, TargetKind};
 
+/// See [`ManifestDependencyInner`] for public fields reference.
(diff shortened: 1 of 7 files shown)


scarb/src/core/manifest/dependency.rs
mkaput, line 1: I'm a bit afraid these constructors will explode for us if we happen to add new fields to this data structure. Perhaps could we go with `DepKind` struct that Cargo has? That design has extra benefits: it has clear default value, and also leaves open option for adding `dev-dependencies` etc.

As for construction, having default kind that is easy to type won't make it necessary to add these constructors IMO, and we can add builder pattern here (including internal Arc, as this struct starts to grow large)

mkaput (changes requested):

scarb/src/core/manifest/dependency.rs
```diff
@@ -1,15 +1,59 @@
...
+pub enum DepKind {
+    Normal,
+    Target(TargetKind),
+}
+
+impl Default for DepKind {
+    fn default() -> Self {
+        Self::Normal
+    }
+}
```
mkaput, line 35: https://rust-lang.github.io/rfcs/3107-derive-default-enum.html

mkaput (approved):

mkaput (approved):

PR #612, opened 2023-08-24
Build dependency resolution with target_kind filter

maciektr:
commit-id:ade97ea3

---

**Stack**:
- #679
- #651
- #649
- #628
- #660
- #615
- #597
- #659
- #612 ⬅

### DIFF
--- a/scarb/src/core/resolver.rs
+++ b/scarb/src/core/resolver.rs
@@ -1,8 +1,9 @@
 use itertools::Itertools;
 use petgraph::graphmap::DiGraphMap;
-use petgraph::visit::{Dfs, Walker};
+use petgraph::visit::{Dfs, EdgeFiltered, Walker};
+use smallvec::SmallVec;
 
-use crate::core::PackageId;
+use crate::core::{PackageId, TargetKind};
 
 // TODO(#126): Produce lockfile out of this.
 /// Represents a fully-resolved package dependency graph.
@@ -13,7 +14,7 @@ pub struct Resolve {
     /// Directional graph representing package dependencies.
     ///
     /// If package `a` depends on package `b`, then this graph will contain an edge from `a` to `b`.
-    pub graph: DiGraphMap<PackageId, ()>,
+    pub graph: DiGraphMap<PackageId, DependencyEdge>,
 }
 
 impl Resolve {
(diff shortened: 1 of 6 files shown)


scarb/src/core/resolver.rs
```diff
@@ -38,4 +39,57 @@ impl Resolve {
             .iter(&self.graph)
             .unique()
     }
+
+    pub fn solution_with_target_kind(
+        &self,
+        root_package: PackageId,
+        target_kind: TargetKind,
+    ) -> Vec<PackageId> {
```
maciektr, line 47: Maybe we could optimise it to not collect vec? 🤔

mkaput: why not?

scarb/src/core/resolver.rs
```diff
@@ -38,4 +39,57 @@ impl Resolve {
             .iter(&self.graph)
             .unique()
     }
+
+    pub fn solution_with_target_kind(
```
mkaput, line 43: Are we calling `solution_of` anywhere now?

scarb/src/core/resolver.rs
```diff
@@ -38,4 +39,57 @@ impl Resolve {
...
+
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct DependencyEdge(SmallVec<[TargetKind; 4]>);
+
+impl DependencyEdge {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn accepts_target_kind(&self, target_kind: TargetKind) -> bool {
```
mkaput, line 67:
```suggestion
    pub fn accepts_target(&self, target_kind: TargetKind) -> bool {
```

IMHO pushing `_kind` everywhere is superfluous

scarb/src/core/resolver.rs
```diff
@@ -38,4 +39,57 @@ impl Resolve {
...
+        });
+        Dfs::new(&filtered_graph, root_package)
+            .iter(&filtered_graph)
+            .unique()
+            .collect_vec()
+    }
+}
+
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct DependencyEdge(SmallVec<[TargetKind; 4]>);
```
mkaput, line 60: Why 4 instead of 1? Can't we represent wildcards here?

mkaput (changes requested):

scarb/src/core/resolver.rs
maciektr: Right now empty vec = wildcard, hence https://github.com/software-mansion/scarb/blob/ae88e56894898b59005a6f7de1361c1829df07a3/scarb/src/core/resolver.rs#L79

scarb/src/core/resolver.rs
maciektr: The iterator references values owned by `filtered_graph` variable, I haven't wrapped my head how to reorganize the code yet

mkaput: In general, the solutions here won't be massive, and it's not really that important to try to optimize out this vec collection. If that's hard, then let's skip trying to do that now.

scarb/src/core/package/name.rs
mkaput, line 1: not in scope of this PR?

scarb/src/ops/compile.rs
```diff
@@ -32,7 +32,7 @@ pub fn compile(packages: Vec<PackageId>, ws: &Workspace<'_>) -> Result<()> {
 }
 
 fn compile_unit(unit: CompilationUnit, ws: &Workspace<'_>) -> Result<()> {
-    let package_name = unit.main_package_id.name.clone();
+    let package_name = unit.main_package_id.name.to_string();
```
mkaput, line 35: why?

mkaput (approved):

mkaput (approved):

Issue #608, opened 2023-08-24
CI: Stop calling `scarb build` from as many tests as possible


mkaput:
Cairo code compilation is slow, especially when the compiler is built in debug mode. But we tend(ed) to overuse `scarb build` command in many E2E test, because that was the only available way of triggering workspace resolution etc. in the past. [This resulted in many tests taking almost a **minute!** to execute in our CI](https://github.com/software-mansion/scarb/actions/runs/5954119709/job/16150069256#step:6:39).

Now we have `scarb metadata` and `scarb fetch`, and possibly `scarb check` in the future (#137), and these are much, much faster. Have a look at all our tests and try to avoid `scarb build` there as much as possible.

Issue #137, opened 2023-03-01
Investigate add `scarb check`


mkaput:
What is unsure is if Sierra codegen is a noticeable factor

mkaput:
For benchmarking, I recommend using https://github.com/sharkdp/hyperfine

PR #616, opened 2023-08-25
Reduce number of scarb build calls in tests

MrDenkoV:
Fix (#608)

### DIFF
--- a/scarb/tests/build.rs
+++ b/scarb/tests/build.rs
@@ -69,7 +69,7 @@ fn compile_with_syntax_error() {
         .build(&t);
 
     Scarb::quick_snapbox()
-        .arg("build")
+        .arg("build") // TODO(#137): Change build to check for faster and lighter test.
         .current_dir(&t)
         .assert()
         .code(1)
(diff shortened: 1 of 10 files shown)


mkaput:
What is missing still here, that this is a draft?

MrDenkoV:
I was thinking #137 will be added soon and wanted to wait for it to be done to change a couple more tests. I will change the number in Todo to #137 and make the PR ready.
@mkaput

mkaput:
> I was thinking #137 will be added soon

Not gonna happen

scarb/tests/build.rs
```diff
@@ -69,7 +69,7 @@ fn compile_with_syntax_error() {
         .build(&t);
 
     Scarb::quick_snapbox()
-        .arg("build")
+        .arg("build") // Todo(#137) check
```
mkaput, line 72: Could you please use the standard pattern for todo comments that we use elsewhere? :)

mkaput (changes requested):

mkaput (approved):

PR #618, opened 2023-08-25
Set Cargo resolver version to 2

maciektr:


mkaput (approved):
Doesn’t this change anything in the lock file ?

maciektr:
I am not getting any changes in Cargo.lock 👁️ 
My guess would be we are already using the `2` resolver, as we set edition to `2021` everywhere, so this just gets rid of the warning.  
https://github.com/software-mansion/scarb/blob/05b5bc1cfa2d9ae3fbcb8f1f9d0ced7281877f2f/Cargo.toml#L18

PR #625, opened 2023-08-29
Fix `errors_when_missing_script_and_cairo` test

### LINKED ISSUE #468: Test `errors_when_missing_script_and_cairo_test` is failing if asdf-scarb is installed on host matchine
```
--- Expected
++++ actual:   stdout
   1      - error: no such command: `cairo-test`
        1 + error: process did not exit successfully: exit status: 2
```

Proposal fix: make Scarb isolation stuff in tests more isolating

szymmis:
Fixes #468

Test was failing because of asdf's Scarb installation interfering with it. 
Solution to that problem is to run Scarb process without `ASDF_DATA_DIR` paths in `PATH` env var.

### DIFF
--- a/utils/scarb-test-support/src/filesystem.rs
+++ b/utils/scarb-test-support/src/filesystem.rs
@@ -1,5 +1,6 @@
 use std::ffi::OsString;
-use std::{env, iter};
+use std::path::PathBuf;
+use std::{env, iter, vec};
 
 use assert_fs::prelude::*;
 use assert_fs::TempDir;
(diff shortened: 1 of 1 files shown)


utils/scarb-test-support/src/filesystem.rs
```diff
@@ -23,9 +24,23 @@ pub fn write_simple_hello_script(name: &str, t: &TempDir) {
     );
 }
 
+fn os_path_without_asdf_dir() -> OsString {
+    let os_path = env::var_os("PATH").unwrap();
+
+    if let Ok(asdf_dir) = env::var("ASDF_DATA_DIR") {
```
mkaput, line 30: I do not have this env on my machine, but I do have the following:

```bash
ASDF_DIR=/Users/mk/.asdf
```

maybe we can also consider this one?

mkaput (changes requested):

mkaput (approved):
I can confirm this test works now on my machine! 🥰

PR #627, opened 2023-08-29
Do not print test prefix

maciektr:
**Stack**:
- #651
- #650
- #649
- #629
- #628
- #626
- #627 ⬅
- #615
- #597
- #612

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #626, opened 2023-08-29
Update cairo compiler revision

maciektr:
**Stack**:
- #651
- #649
- #628
- #626 ⬅
- #660
- #615
- #597
- #659
- #612

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #615, opened 2023-08-25
Use ws solution per target kind

maciektr:
commit-id:75463384

---

**Stack**:
- #688
- #679
- #651
- #649
- #628
- #660
- #615 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #614, opened 2023-08-25
Add TestPlugin dependency

maciektr:
**Stack**:
- #630
- #629
- #628
- #626
- #627
- #615
- #614 ⬅
- #597
- #612
- #611

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #628, opened 2023-08-29
Compile test targets

maciektr:
**Stack**:
- #627
- #626
- #615
- #614
- #597
- #612
- #611
- #596


scarb/src/ops/resolve.rs
```diff
@@ -337,7 +337,11 @@ fn is_integration_test(target: &Target, lib_target: Option<Target>) -> bool {
 
 /// Build a set of `cfg` items to enable while building the compilation unit.
 fn build_cfg_set(target: &Target) -> CfgSet {
-    CfgSet::from_iter([Cfg::kv("target", target.kind.clone())])
+    let mut cfg = vec![Cfg::kv("target", target.kind.clone())];
+    if target.is_test() {
+        cfg.push(Cfg::name("test"));
+    }
+    CfgSet::from_iter(cfg)
```
mkaput, line 344:
```suggestion
    let mut cfg = CfgSet::from_iter([
        Cfg::kv("target", target.kind.clone())
    ]);
    if target.is_test() {
        cfg.insert(Cfg::name("test"));
    }
    cfg
```

nitpick, isn't this simpler? :)

scarb/src/compiler/helpers.rs
```diff
@@ -31,3 +34,22 @@ pub fn collect_main_crate_ids(unit: &CompilationUnit, db: &RootDatabase) -> Vec<
...
+    file_name: &str,
+    description: &str,
+    target_dir: &Filesystem<'_>,
+    ws: &Workspace<'_>,
+    value: impl Serialize,
+) -> Result<()> {
+    let mut file = target_dir.open_rw(file_name, description, ws.config())?;
+    serde_json::to_writer(&mut *file, &value)
+        .with_context(|| format!("failed to serialize {file_name}"))
+}
```
mkaput, line 55: this helper is so simple that I wonder if we shouldn't just remove it

scarb/src/compiler/db.rs
```diff
@@ -39,10 +42,19 @@ fn build_project_config(unit: &CompilationUnit) -> Result<ProjectConfig> {
...
+                .clone()
+                .ends_with(DEFAULT_SOURCE_PATH)
+            {
+                CrateRootPath::Simple(component.target.source_root().into())
+            } else {
+                CrateRootPath::Entrypoint(DetailedEntrypointPath {
+                    entrypoint: component.target.source_path.clone().into(),
+                })
+            };
+            (component.cairo_package_name(), crate_root)
```
mkaput, line 57: ugh, this doesn't sound good as I see it now. This seems to be easily unifyable on API surface, perhaps we should fix compiler internal issues blocking this instead of convoluting its API?

scarb/src/compiler/compilers/test.rs
```diff
@@ -13,10 +17,44 @@ impl Compiler for TestCompiler {
...
+                    .print(TypedMessage::naked_text("diagnostic", &diagnostic));
+            }
+        });
+
+        diagnostics_reporter
+            .with_extra_crates(&main_crate_ids)
+            .ensure(db)?;
+
+        let test_compilation =
+            compile_test_prepared_db(db, starknet, main_crate_ids, test_crate_ids)?;
```
mkaput, line 47: This is **not** going to scale. All Scarb changes to root database flow will require upstream changes in the compiler. We need to be able to end-to-end control what happens with the root database. Perhaps compiler APIs should focus on collecting test-specific information only?

mkaput (changes requested):

scarb/src/compiler/compilers/test.rs
maciektr: Discussed offline.

scarb/src/compiler/db.rs
```diff
@@ -30,7 +34,42 @@ pub(crate) fn build_scarb_root_database(
         }
     }
 
-    b.build()
+    let mut db = b.build()?;
+    inject_virtual_wrapper_lib(&mut db, unit)?;
+    Ok(db)
+}
+
+fn inject_virtual_wrapper_lib(db: &mut RootDatabase, unit: &CompilationUnit) -> Result<()> {
```
mkaput, line 42: document please

scarb/src/compiler/db.rs
```diff
@@ -30,7 +34,42 @@ pub(crate) fn build_scarb_root_database(
...
+            })?;
+            let module_id = ModuleId::CrateRoot(crate_id);
+            let file_id = db.module_main_file(module_id).unwrap();
+            // Inject virtual lib file wrapper.
+            db.as_files_group_mut()
+                .override_file_content(file_id, Some(Arc::new(format!("mod {file_stem};"))));
+            Ok(())
+        })
+        .collect::<Result<Vec<()>>>()
+        .map(|_| ())
```
mkaput, line 72: Wouldn't it be easier to use a normal for loop here? 😅

mkaput (changes requested):

scarb/src/compiler/db.rs
```diff
@@ -30,7 +34,49 @@ pub(crate) fn build_scarb_root_database(
...
+                .file_name()
+                .map(|file_name| file_name != DEFAULT_MODULE_MAIN_FILE)
+                .unwrap_or(false)
+        })
+        .collect();
+
+    for component in components {
+        let crate_name = component.cairo_package_name();
+        let crate_id = db.intern_crate(CrateLongId::Real(crate_name));
+        let file_stem = component.target.source_path.file_stem().ok_or_else(|| {
```
mkaput, line 66: hmm, how this will react when we create a file `tests/ęśąćż.cairo` or `tests/file space.cairo`? both are not valid identifiers, are we validating these?

not a topic to fix in this PR, I'd rather create a bug report labelled as good first issue 😄

mkaput (approved):

PR #632, opened 2023-08-30
Disallow uppercase letters in package name

### LINKED ISSUE #591: Disallow upper case letters in package names
### Problem

We think we should stop allowing uppercase letters in package names:
1. Allowing both `foo` and `Foo` packages to exist may be dangerous.
2. Package names must be valid & idiomatic Cairo module names, and these are `snake_case`.
3. We copied current behaviour from Cargo, and it allows upper case only because it didn't care earlier. Rust [RFC 430](https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md) established `snake_case` to be the way to go for crates, though Cargo packages are still [unclear](https://github.com/rust-lang/api-guidelines/discussions/29#discussioncomment-6659367).

### Proposed Solution

Make Scarb reject uppercase letters in package names. To do that, patch the [`PackageName` struct](https://github.com/software-mansion/scarb/blob/e9078ddfc5cf766dee8ab1054b911d204af73170/scarb/src/core/package/name.rs#L43-L83). Don't forget about tests. Also, please make a special condition for this that will print a tailored error message, guiding users what to do in order to fix it.

### Notes

_No response_

MrDenkoV:
Fix #591

### DIFF
--- a/scarb/src/core/package/name.rs
+++ b/scarb/src/core/package/name.rs
@@ -49,6 +49,15 @@ impl PackageName {
             bail!("underscore cannot be used as package name");
         }
 
+        if !name.eq(&name.to_ascii_lowercase()) {
+            bail!(
+                "invalid package name: `{name}`\n\
+                note: usage of ASCII uppercase letters in the package name has been disallowed\n\
+                help: change package name to: {}",
+                name.to_ascii_lowercase()
+            )
+        }
+
         let mut chars = name.chars();
 
         // Validate first letter.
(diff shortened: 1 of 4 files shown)


scarb/src/core/package/name.rs
```diff
@@ -49,6 +49,15 @@ impl PackageName {
             bail!("underscore cannot be used as package name");
         }
 
+        if !name.eq(&name.to_ascii_lowercase()) {
+            bail!(
+                "invalid package name: `{name}`\n\
+                note: usage of ASCII uppercase letters in the package name has been deprecated\n\
+                help: change the name of the package to `{}`",
```
mkaput, line 56:
```suggestion
                help: change package name to: {}",
```

mkaput (approved):

mkaput (approved):
lgtm!

PR #633, opened 2023-08-31
Support import sorting via manifest configuration

### LINKED ISSUE #609: Add imports sorting to formatter config
### Problem

Cairo-fmt now supports imports sorting. 
https://github.com/starkware-libs/cairo/pull/3778

### Proposed Solution

_No response_

### Notes

_No response_

szymmis:
Resolves #609

Introduces `[tool.fmt]` section to `Scarb.toml` with `sort-imports` boolean flag

### DIFF
--- a/scarb/src/ops/fmt.rs
+++ b/scarb/src/ops/fmt.rs
@@ -10,6 +10,7 @@ use tracing::{info, warn};
 
 use crate::core::workspace::Workspace;
 use crate::core::PackageId;
+use crate::internal::serdex::toml_merge;
 
 #[derive(Debug)]
 pub struct FmtOptions {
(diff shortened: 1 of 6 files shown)


scarb/src/ops/fmt.rs
```diff
@@ -21,15 +21,18 @@ pub struct FmtOptions {
...
     let packages = ws
         .members()
         .filter(|pkg| opts.packages.contains(&pkg.id))
         .collect::<Vec<_>>();
     let Some((first_package, packages)) = packages.split_first() else {
         return Ok(true);
     };
+
+    let config = first_package.manifest.fmt_config.clone().into();
+    let fmt = CairoFormatter::new(config);
```
mkaput, line 34: There's a bug here: what if particular packages declare their own formatter configs? I think we should sequentially walk over all packages here. It's a tradeoff, but it shouldn't be so severe I think.

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -701,6 +704,20 @@ impl TomlManifest {
             })
             .transpose()
     }
+
+    fn collect_fmt_config(&self, tool: &Option<TomlToolsDefinition>) -> Result<ManifestFmtConfig> {
```
mkaput, line 708: I gave this PR a thought and I think this code does it wrong 😞 I didn't investigate the context when outlining implementation steps; my bad.
1. We should stick with original naming, so this field should be named `sort-module-level-items`, for consistency.
2. I don't think we should trigger errors when reading the manifest, if user has errors in `[tool.*]` sections. These should pop only in affected tools, IMHO. This code actually does not validate the values at all, it just ignores them.
3. There is a way how to deduplicate stuff nicely!

So first of all, do note that Cairo's `FormatterConfig` has `Default` implementation. I think we should submit a PR which adds the following for this struct:

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(rename_all = "kebab-case")]
pub struct FormatterConfig {
    tab_size: usize,
    max_line_length: usize,
    sort_module_level_items: bool,
}
```

This way we will keep upstream 1) default values 2) canonical serde of this struct.

Now, about parsing. We have very nifty methods in our `Package` struct: https://github.com/software-mansion/scarb/blob/d98fca8190ff330109b414ed70634b569732a6b5/scarb/src/core/package/mod.rs#L92-L113

We also have this very useful helper function: https://github.com/software-mansion/scarb/blob/c412b06b0e1069ea1ca503c945d5e8dfd12bb398/scarb/src/internal/serdex.rs#L8-L25

What we can do with this the following. All of this will happen within `fmt` op **only**!

```rust
let mut config = FormatterConfig::default();
if let Some(overrides) = package.tool_metadata("fmt") {
    config = toml_merge(&config, overrides)?;
}
```

_voila!_

mkaput (changes requested):

mkaput (commented):
This looks great, we just need to update Cairo and fix the per-package config bug 🎉

mkaput (approved):

scarb/src/ops/fmt.rs
```diff
@@ -21,30 +22,32 @@ pub struct FmtOptions {
...
+
     let packages = ws
         .members()
-        .filter(|pkg| opts.packages.contains(&pkg.id))
-        .collect::<Vec<_>>();
-    let Some((first_package, packages)) = packages.split_first() else {
-        return Ok(true);
-    };
-    let mut walk = fmt.walk(first_package.root().as_std_path());
+        .filter(|package| opts.packages.contains(&package.id));
```
mkaput, line 30: nitpick: sounds like you want to use within the `for` loop https://github.com/software-mansion/scarb/blob/d98fca8190ff330109b414ed70634b569732a6b5/scarb/src/core/workspace.rs#L112-L115

note: the old code looked like this because it collected everything to vector; now you can ignore it

mkaput (approved):

PR #637, opened 2023-09-01
Add textual Sierra artifact support

### LINKED ISSUE #634: Add `lib.sierra-text` property to generate textual Sierra representation of package's code
Add the following property to Scarb.toml

```toml
[lib]
sierra-text = true
```

Defaults to `false`.

Enabling it will generate the "original" `*.sierra` artefact.

MrDenkoV:
Fix #634

### DIFF
--- a/scarb/src/compiler/compilers/lib.rs
+++ b/scarb/src/compiler/compilers/lib.rs
@@ -18,6 +18,7 @@ pub struct LibCompiler;
 struct Props {
     pub sierra: bool,
     pub casm: bool,
+    pub sierra_text: bool,
 }
 
 impl Default for Props {
@@ -25,6 +26,7 @@ impl Default for Props {
         Self {
             sierra: true,
             casm: false,
+            sierra_text: false,
         }
     }
 }
@@ -41,9 +43,9 @@ impl Compiler for LibCompiler {
         ws: &Workspace<'_>,
     ) -> Result<()> {
         let props: Props = unit.target().props()?;
-        if !props.sierra && !props.casm {
+        if !props.sierra && !props.casm && !props.sierra_text {
             ws.config().ui().warn(
-                "both Sierra and CASM lib targets have been disabled, \
+                "Sierra, textual Sierra and CASM lib targets have been disabled, \
                 Scarb will not produce anything",
             );
         }
@@ -72,6 +74,15 @@ impl Compiler for LibCompiler {
             .with_context(|| {
                 format!("failed to serialize Sierra program {}", unit.target().name)
             })?;
+        }
+
+        if props.sierra_text {
+            let mut file = target_dir.open_rw(
+                format!("{}.sierra", unit.target().name),
+                "output file",
+                ws.config(),
+            )?;
+            file.write_all(sierra_program.to_string().as_bytes())?;
         }
 
         if props.casm {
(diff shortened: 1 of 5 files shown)


website/docs/reference/targets.md
```diff
@@ -21,13 +21,18 @@ The library target accepts following configuration parameters, with default valu
...
 ```
 
-By default, the library target builds unprocessed Sierra code in textual form for the package.
+By default, the library target builds unprocessed Sierra code in JSON form for the package.
 When the `casm` option is enabled, Scarb can also automatically compile this Sierra code down to CASM.
 
+:::warning
+Textual Sierra is being discouraged, try using JSON Sierra instead.
+:::
+
```
mkaput, line 35: I don't think we need to shout at the users with a warning box (the less boxes, the generally better, because people will actually pay attention to these if there are few of them). Adding just a single sentence saying that the textual Sierra is fine for debugging would be cool IMHO

website/docs/cheatsheet.md
```diff
@@ -28,21 +28,26 @@ Built Sierra code of this package will be written to `target/dev` directory.
 `scarb clean` cleans `target` directory.
 :::
 
-### Building CASM
+### Building CASM and textual Sierra
```
mkaput, line 31: This is a cheatsheet, we don't have to showcase all features here, and this particular one is very minor. Please revert changes in this file.

scarb/src/compiler/compilers/lib.rs
```diff
@@ -74,6 +76,18 @@ impl Compiler for LibCompiler {
...
+        if props.sierra_text {
+            let mut file = target_dir.open_rw(
+                format!("{}.sierra", unit.target().name),
+                "output file",
+                ws.config(),
+            )?;
+            file.write_all(sierra_program.to_string().as_bytes())?;
+            ws.config()
+                .ui()
+                .warn("textual Sierra lib target is being discouraged, use JSON Sierra instead")
```
mkaput, line 88: I don't think this warning is desired. It's not bad to use this output kind if you know what you do. And if users did pay attention to the docs in order to find this feature, then I conclude they do know what they do? Wdyt? I suggest removing this.

website/docs/reference/targets.md
```diff
@@ -21,13 +21,18 @@ The library target accepts following configuration parameters, with default valu
 
 ```toml
 [lib]
-sierra = true  # Enable Sierra codegen.
+sierra = true  # Enable JSON Sierra codegen.
```
mkaput, line 24: Let's call it just `Sierra`

mkaput (commented):
Looks good, I have minor comments

scarb/src/compiler/compilers/lib.rs
mkaput: Your worries are definitely correct! But I think we can tackle this from the opposite side: by pushing tools to migrate to JSON :)

website/docs/reference/targets.md
```diff
@@ -23,10 +23,12 @@ The library target accepts following configuration parameters, with default valu
 [lib]
 sierra = true  # Enable Sierra codegen.
 casm = false   # Enable CASM codegen.
+sierra-text = false # Enable textual Sierra codegen.
```
mkaput, line 26:
```suggestion
sierra = true        # Enable Sierra codegen.
casm = false         # Enable CASM codegen.
sierra-text = false  # Enable textual Sierra codegen.
```

nitpick

mkaput (approved):
Looks good! Please squash commits and rebase on top of `main`

PR #649, opened 2023-09-07
Skip test target compilation without --test flag

maciektr:
commit-id:a71d73f9

---

**Stack**:
- #688
- #679
- #651
- #649 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #629, opened 2023-08-29
Use compiled test targets in test runner

maciektr:
**Stack**:
- #651
- #650
- #649
- #629 ⬅
- #628
- #626
- #627
- #615
- #597
- #612

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #650, opened 2023-09-07
Fix sole target check for ui printing

maciektr:
**Stack**:
- #651
- #650 ⬅
- #649
- #629
- #628
- #626
- #627
- #615
- #597
- #612

### DIFF
--- a/scarb/src/compiler/compilation_unit.rs
+++ b/scarb/src/compiler/compilation_unit.rs
@@ -81,10 +81,6 @@ impl CompilationUnit {
         config.target_dir().child(self.profile.as_str())
     }
 
-    pub fn is_sole_for_package(&self) -> bool {
-        self.main_component().package.manifest.targets.len() >= 2
-    }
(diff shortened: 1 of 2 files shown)


scarb/src/compiler/compilation_unit.rs
```diff
@@ -93,11 +89,11 @@ impl CompilationUnit {
         format!("{}-{}", self.main_package_id.name, self.digest())
     }
 
-    pub fn name(&self) -> String {
+    pub fn name(&self, is_sole_for_package: bool) -> String {
```
mkaput, line 92: I don't like the fact that this getter is getting an extra argument. Especially as we can define a clear logic, that should be computable within the old `is_sole_for_package` check: we can just not count test targets in that condition.

mkaput (changes requested):

maciektr:
Closing pull request: commit has gone away

PR #651, opened 2023-09-07
Auto-detect test targets

maciektr:
**Stack**:
- #650
- #649
- #629
- #628
- #626
- #627
- #615
- #597
- #612
- #611


scarb/tests/build_targets.rs
```diff
@@ -184,6 +184,7 @@ fn compile_dep_not_a_lib() {
         .assert()
         .failure()
         .stdout_matches(indoc! {r#"
+            warn: hello v1.0.0 ([..]) ignoring invalid dependency `dep` which is missing a lib or cairo-plugin target
```
mkaput, line 187: hmm why did this line appear? are we able to avoid it?

mkaput (changes requested):

scarb/tests/build_targets.rs
maciektr: We are now pulling deps for each generated compilation unit https://github.com/software-mansion/scarb/blob/cc3ff3894d6aef6d8a4d5d0323a103a6bff8cede/scarb/src/ops/resolve.rs#L317

maciektr: discussed offline

mkaput (approved):

PR #653, opened 2023-09-08
Fix macos failing on cairo run extensions test

szymmis:


.github/workflows/_check-release.yml
```diff
@@ -1,6 +1,7 @@
 name: Release Check
 
 on:
+  workflow_dispatch:
```
mkaput, line 4: missing inputs, this will crash I believe without them

extensions/scarb-cairo-run/Cargo.toml
```diff
@@ -17,6 +17,7 @@ scarb-metadata = { path = "../../scarb-metadata" }
 scarb-ui = { path = "../../utils/scarb-ui" }
 serde.workspace = true
 serde_json.workspace = true
+assert_fs.workspace = true
```
mkaput, line 20: sort these please

extensions/scarb-cairo-run/tests/examples.rs
```diff
@@ -42,12 +48,17 @@ fn package_not_built() {
...
         .join("hello_world");
+
+    let t = TempDir::new().unwrap();
+
     Command::new(cargo_bin("scarb"))
+        .env("SCARB_TARGET_DIR", t.path())
         .arg("clean")
         .current_dir(example.clone())
         .assert()
         .success();
```
mkaput, line 59: this is superfluous now, as we have fresh target dir

mkaput (changes requested):

.github/workflows/_check-release.yml
szymmis: Workflow dispatch doesn't take any inputs. It's exactly the same situations as in `website-deploy` and `nightly` workflows.

mkaput: But this workflow actually has `${{ inputs.* }}` expressions, and I believe actions runner will yell that it doesn't know these inputs, when executed via dispatch.

szymmis: Yeah, I don't think it would crash, but it is possible that it would complain about it so it makes sense to specify those.

PR #655, opened 2023-09-08
Infer readme

### LINKED ISSUE #159: Infer value of the `readme` field by looking for readme file when converting TomlManifest to Manifest
Do like Cargo does: https://doc.rust-lang.org/cargo/reference/manifest.html#the-readme-field

MrDenkoV:
Fix #159

### DIFF
--- a/scarb/src/core/manifest/toml_manifest.rs
+++ b/scarb/src/core/manifest/toml_manifest.rs
@@ -4,10 +4,12 @@ use std::fs;
 use std::path::PathBuf;
 
 use anyhow::{anyhow, bail, ensure, Context, Result};
-use camino::Utf8Path;
+use camino::{Utf8Path, Utf8PathBuf};
 use itertools::Itertools;
+use pathdiff::diff_utf8_paths;
 use semver::{Version, VersionReq};
-use serde::{Deserialize, Serialize};
+use serde::{de, Deserialize, Serialize};
+use serde_untagged::UntaggedEnumVisitor;
 use smol_str::SmolStr;
 use tracing::trace;
 use url::Url;
@@ -103,7 +105,7 @@ pub struct PackageInheritableFields {
     pub keywords: Option<Vec<String>>,
     pub license: Option<String>,
     pub license_file: Option<String>,
-    pub readme: Option<String>,
+    pub readme: Option<PathOrBool>,
     pub repository: Option<String>,
     pub cairo_version: Option<VersionReq>,
 }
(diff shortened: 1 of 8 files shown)


scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -704,6 +744,56 @@ impl TomlManifest {
...
+            .with_context(|| {
+                format!(
+                    "failed to find the readme at {}",
+                    prefix.parent().unwrap().join(readme)
+                )
+            }),
+    }
+}
+
+const DEFAULT_README_FILES: [&str; 3] = ["README.md", "README.txt", "README"];
```
mkaput, line 784:
```suggestion
const DEFAULT_README_FILES: &[&str] = ["README.md", "README.txt", "README"];
```

you can do this, it's easier to maintain later on :)

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -704,6 +744,56 @@ impl TomlManifest {
...
+
+/// Checks if a file with any of the default README file names exists in the package root.
+/// If so, returns a `Utf8Path` with that name.
+fn default_readme_from_package_root(package_root: &Utf8Path) -> Option<&Utf8Path> {
+    for &readme_filename in DEFAULT_README_FILES.iter() {
```
mkaput, line 789:
```suggestion
    for &readme_filename in DEFAULT_README_FILES {
```

there's builtin `IntoIter` implementation for `&[T]`

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -704,6 +744,56 @@ impl TomlManifest {
...
+    readme: Option<&PathOrBool>,
+) -> Result<Option<Utf8PathBuf>> {
+    match readme {
+        None => abs_canonical_path(
+            package_root,
+            default_readme_from_package_root(package_root.parent().unwrap()),
+        ),
+        Some(value) => match value {
+            PathOrBool::Bool(false) => Ok(None),
+            PathOrBool::Bool(true) => abs_canonical_path(package_root, Some("README.md".into())),
```
mkaput, line 759: Shouldn't we look for `default_readme_from_package_root` here also? This looks inconsistent to me

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -704,6 +744,56 @@ impl TomlManifest {
     }
 }
 
+/// Returns the absolute canonical path of the README file for a [`TomlPackage`].
+pub fn readme_for_package(
+    package_root: &Utf8Path,
+    readme: Option<&PathOrBool>,
+) -> Result<Option<Utf8PathBuf>> {
+    match readme {
+        None => abs_canonical_path(
```
mkaput, line 753:
```suggestion
        Some(PathOrBool(true)) | None => abs_canonical_path(
```

you can do this. note, that you'll have to flatten nested `match` expression in `Some(value)` case, i.e.

```rust
match readme {
    Some(PathOrBool::Bool(true)) | None => smth,
    Some(PathOrBool::Bool(false)) => smth,
    Some(PathOrBool::Path(p)) => smth
}
```

scarb/src/core/manifest/toml_manifest.rs
MrDenkoV: Now it is accurate with the documentation (manifest.md change) and how cargo does it.
I don't have a strong opinion on this.
Perhaps setting `readme = true` and not setting it at all, should be treated differently, but I can change it and add the check for `readme = true`, that ensures the file exists.

scarb/src/core/manifest/toml_manifest.rs
MrDenkoV: Thank you for this reminder!
However, in this case I think it's better to have those match arms separate, instead of combining them to add an if later to separate them (I want to make sure a readme exists if it's set to true, but doesn't have to exist if readme isn't set at all).

mkaput: Yeah, what you have is fine, we can leave this as-is :)

mkaput (approved):
👏🏻

PR #657, opened 2023-09-08
Scaffold `scarb package` command, implementing `--list` arg

mkaput:
**Stack**:
- #697
- #710
- #726
- #675
- #657 ⬅


szymmis (approved):

scarb/src/ops/package.rs
```diff
@@ -0,0 +1,194 @@
...
+            } else {
+                ws.config().ui().print(message);
+            }
+
+            package_one_impl(*pkg, opts, ws)
+        })
+        .collect()
+}
+
+pub fn package_one(
```
maciektr, line 69: Do we need this? I do not think this is used.

maciektr (approved):

scarb/src/ops/package.rs
mkaput: Right, I was too much forward-thinking (this may be useful for `scarb publish` 😅). I'll remove it.

PR #658, opened 2023-09-08
Fix exec_replace

### LINKED ISSUE #125: Properly implement `scarb::process::exec_replace`
Fix the TODO comment left there. Doc comment should fully describe what should happen internally.

https://github.com/software-mansion/scarb/blob/7bf281cc01ddc9791275a57aa91f65077b7795f4/scarb/src/process.rs#L14-L48

### LINKED ISSUE #129: Fix subcommand CTRL+C test
https://github.com/software-mansion/scarb/blob/7bf281cc01ddc9791275a57aa91f65077b7795f4/scarb/tests/e2e/subcommand.rs#L140-L182

MrDenkoV:
Fix #125, Fix #129

### DIFF
--- a/scarb/src/process.rs
+++ b/scarb/src/process.rs
@@ -12,7 +12,6 @@ use scarb_ui::components::{Spinner, Status};
 
 use crate::core::Config;
 
-// TODO(#125): Do what is documented here, take a look at what cargo-util does.
 /// Replaces the current process with the target process.
 ///
 /// On Unix, this executes the process using the Unix syscall `execvp`, which will block this
@@ -30,21 +29,58 @@ use crate::core::Config;
 /// itself later exits.
 #[tracing::instrument(level = "debug")]
 pub fn exec_replace(cmd: &mut Command) -> Result<()> {
-    let exit_status = cmd
-        .spawn()
-        .with_context(|| format!("failed to spawn: {}", cmd.get_program().to_string_lossy()))?
-        .wait()
-        .with_context(|| {
-            format!(
-                "failed to wait for process to finish: {}",
-                cmd.get_program().to_string_lossy()
-            )
-        })?;
+    imp::exec_replace(cmd)
+}
 
-    if exit_status.success() {
-        Ok(())
-    } else {
-        bail!("process did not exit successfully: {exit_status}");
+#[cfg(unix)]
+mod imp {
+    use anyhow::{bail, Result};
+    use std::os::unix::process::CommandExt;
+    use std::process::Command;
+
+    pub fn exec_replace(cmd: &mut Command) -> Result<()> {
+        let err = cmd.exec();
+        bail!("process did not exit successfully: {err}")
+    }
+}
+
+#[cfg(windows)]
+mod imp {
+    use anyhow::{bail, Context, Result};
+    use std::process::Command;
+    use windows_sys::Win32::Foundation::{BOOL, FALSE, TRUE};
+    use windows_sys::Win32::System::Console::SetConsoleCtrlHandler;
+
+    unsafe extern "system" fn ctrlc_handler(_: u32) -> BOOL {
+        // Do nothing; let the child process handle it.
+        TRUE
+    }
+
+    pub fn exec_replace(cmd: &mut Command) -> Result<()> {
+        unsafe {
+            if SetConsoleCtrlHandler(Some(ctrlc_handler), TRUE) == FALSE {
+                panic!("Could not set Ctrl-C handler.");
+            }
+        }
+
+        // Just execute the process as normal.
+
+        let exit_status = cmd
+            .spawn()
+            .with_context(|| format!("failed to spawn: {}", cmd.get_program().to_string_lossy()))?
+            .wait()
+            .with_context(|| {
+                format!(
+                    "failed to wait for process to finish: {}",
+                    cmd.get_program().to_string_lossy()
+                )
+            })?;
+
+        if exit_status.success() {
+            Ok(())
+        } else {
+            bail!("process did not exit successfully: {exit_status}");
+        }
     }
 }
 
--- a/scarb/tests/subcommand.rs
+++ b/scarb/tests/subcommand.rs
@@ -1,10 +1,12 @@
 use std::io::Read;
 use std::net::TcpListener;
-use std::process::{Child, Stdio};
+use std::process::{Child, Command};
 use std::{env, io};
 
 use assert_fs::TempDir;
-use indoc::{formatdoc, indoc};
+#[cfg(unix)]
+use indoc::indoc;
+use scarb_test_support::cargo::cargo_bin;
 
 use scarb_test_support::command::Scarb;
 use scarb_test_support::filesystem::{path_with_temp_dir, write_script, write_simple_hello_script};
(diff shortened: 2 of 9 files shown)


scarb/tests/subcommand.rs
```diff
@@ -115,9 +115,7 @@ fn env_scarb_log_is_passed_verbatim() {
         .success();
 }
 
-// TODO(#129): Fix this test.
 #[test]
-#[ignore] // something doesn't work here
```
mkaput, line 120: WOW

scarb/src/process.rs
mkaput, line 1: can you please remove the TODO line here?: https://github.com/software-mansion/scarb/pull/658/files#diff-c6dd60aae938af440d1b0c7787bdcdd9330cd65cb1c6d5ebec2e3209f7ac8791R15

scarb/src/process.rs
```diff
@@ -30,21 +30,58 @@ use crate::core::Config;
...
+
+    unsafe extern "system" fn ctrlc_handler(_: u32) -> BOOL {
+        // Do nothing; let the child process handle it.
+        TRUE
+    }
+
+    pub fn exec_replace(cmd: &Command) -> Result<()> {
+        unsafe {
+            if SetConsoleCtrlHandler(Some(ctrlc_handler), TRUE) == FALSE {
+                return Err(ProcessError::new("Could not set Ctrl-C handler.", None, None).into());
```
mkaput, line 63: 1. Where does `ProcessError` come from?
2. Woudln't it make sense to just panic here?

scarb/src/process.rs
```diff
@@ -30,21 +30,58 @@ use crate::core::Config;
...
+    use std::process::Command;
+
+    pub fn exec_replace(cmd: &mut Command) -> Result<()> {
+        let err = cmd.exec();
+        bail!("process did not exit successfully: {err}")
+    }
+}
+
+#[cfg(windows)]
+mod imp {
```
mkaput, line 49: Have you tested this on Windows?

scarb/src/process.rs
```diff
@@ -30,21 +30,58 @@ use crate::core::Config;
...
-
-    if exit_status.success() {
-        Ok(())
-    } else {
-        bail!("process did not exit successfully: {exit_status}");
+    imp::exec_replace(cmd)
+}
+
+#[cfg(unix)]
+mod imp {
```
mkaput, line 37: Does this compile and run on Linux and macOS?

mkaput (changes requested):
I am amazed that this fixes #129 

Let's test this more thoroughly

scarb/src/process.rs
MrDenkoV: 😅

scarb/src/process.rs
MrDenkoV: Doesn't work on Windows, our test hangs... 😰

maciektr:
👋 
On the topic of our offline conversation about ctrl_c test on windows 🪟 :

The error you have been seeing seems to come from the subcommand spawn. 
```
This version of %1 is not compatible with the version of Windows you're running. Check your computer's system information and then contact the software publisher. (os error 216)
```
It seems that the script definition used there is just not runnable on windows. (https://github.com/software-mansion/scarb/blob/f8494ed3290c5ae45aa7015a2cf36f38c12f58ff/scarb/tests/subcommand.rs#L144) 

After changing the script definition, I was able to run it successfully and ensure it has been killed by ctrl_c. 
To test this, I have created a `main.rs` file in `scarb-test-support` crate with following content: https://gist.github.com/maciektr/6b6c18a358b5792912bc0d0261ef3ac2
Then I have used this in tests instead.
```
    let addr = {
        let addr = listener.local_addr().unwrap();
        addr.to_string()
    };

    let mut child = Scarb::new()
        .std()
        .arg("test-support")
        .arg("hang-on-tcp")
        .arg("--address")
        .arg(addr)
```


I have not been able to make this check work though. I do not understand what is the purpose of this. 
```
    match sock.read(&mut [0; 10]) {
        Ok(n) => assert_eq!(n, 0),
        Err(e) => assert_eq!(e.kind(), io::ErrorKind::ConnectionReset),
    }
```
I know it has been copied from Cargo verbatim. Maybe it is enough for now to only check if the process has been killed? cc @mkaput

mkaput:
> I have not been able to make this check work though. I do not understand what is the purpose of this.
> 
> ```
>     match sock.read(&mut [0; 10]) {
>         Ok(n) => assert_eq!(n, 0),
>         Err(e) => assert_eq!(e.kind(), io::ErrorKind::ConnectionReset),
>     }
> ```
> 
> I know it has been copied from Cargo verbatim. Maybe it is enough for now to only check if the process has been killed? cc @mkaput

This check tests if the socket either closed successfully without any unread rubbish left in the buffer, or died by the second party being terminated.

> It seems that the script definition used there is just not runnable on windows. (
> 
> https://github.com/software-mansion/scarb/blob/f8494ed3290c5ae45aa7015a2cf36f38c12f58ff/scarb/tests/subcommand.rs#L144
> 
> )
> After changing the script definition, I was able to run it successfully and ensure it has been killed by ctrl_c. To test this, I have created a `main.rs` file in `scarb-test-support` crate with following content: https://gist.github.com/maciektr/6b6c18a358b5792912bc0d0261ef3ac2 Then I have used this in tests instead.
> 
> ```
>     let addr = {
>         let addr = listener.local_addr().unwrap();
>         addr.to_string()
>     };
> 
>     let mut child = Scarb::new()
>         .std()
>         .arg("test-support")
>         .arg("hang-on-tcp")
>         .arg("--address")
>         .arg(addr)
> ```
> 
> May not be the cleanest flow, but imho worth considering.

I like this! I'd only rather use `Command::new(cargo_bin!("scarb-test-support"))` instead of `Scarb::new().std()`, seems more clear.

scarb/src/process.rs
mkaput: How is it now?

utils/scarb-test-support/src/main.rs
```diff
@@ -0,0 +1,39 @@
...
+
+fn hang_on_tcp(args: HangOnTcpArgs) -> Result<()> {
+    use std::io::Read;
+    use std::net::TcpStream;
+
+    let address: &str = args.address.as_ref();
+
+    let mut socket = TcpStream::connect(address).unwrap();
+    let _ = socket.read(&mut [0; 10]);
+    panic!("that read should never return");
```
mkaput, line 38: https://doc.rust-lang.org/stable/std/macro.unreachable.html

utils/scarb-test-support/src/main.rs
```diff
@@ -0,0 +1,39 @@
...
+    address: String,
+}
+
+fn main() {
+    let args: Args = Args::parse();
+    let result = match args.command {
+        Command::HangOnTcp(args) => hang_on_tcp(args),
+    };
+    result.unwrap();
+}
```
mkaput, line 28:
```suggestion
fn main() -> Result<()> {
    let args = Args::parse();
    match args.command {
        Command::HangOnTcp(args) => hang_on_tcp(args),
    }
}
```

scarb/tests/subcommand.rs
```diff
@@ -115,38 +117,18 @@ fn env_scarb_log_is_passed_verbatim() {
...
 fn ctrl_c_kills_everyone() {
-    let t = assert_fs::TempDir::new().unwrap();
     let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+    let addr = {
+        let addr = listener.local_addr().unwrap();
+        addr.to_string()
+    };
```
mkaput, line 126:
```suggestion
    let addr = listener.local_addr().unwrap().to_string();
```

scarb/src/process.rs
MrDenkoV: It works! C:

utils/scarb-test-support/tests/examples.rs
```diff
@@ -0,0 +1,38 @@
+use scarb_test_support::command::Scarb;
+
+#[test]
+fn test_help() {
```
mkaput, line 4: what's the reason behind adding this?

mkaput (changes requested):

utils/scarb-test-support/tests/examples.rs
MrDenkoV: It's for cargo test to create the binary "scarb-test-support".
It helps cargo_bin("scarb-test-support").

mkaput: Can you please:
1. Add an explanatory comment to this file.
2. Rename the test `binary_dependencies_hack` and leave its body empty. I do not want to test help output, especially in such exotic place.
3. Rename test file to something more meaningful. Even simple `hack.rs` is better than `examples.rs` IMO

mkaput (changes requested):

utils/scarb-test-support/tests/examples.rs
mkaput: And also, please create an issue in our tracker to fix this hack with https://github.com/rust-lang/cargo/issues/9096 once it's released in stable Rust.

MrDenkoV: Done!

mkaput (approved):

PR #666, opened 2023-09-12
Fix colours in subcommands

### LINKED ISSUE #598: Make Scarb preserve terminal colors of arbitrary scripts
### Problem

Currently, when defining arbitrary scripts in `Scarb.toml` (for example `test = 'snforge'`), Scarb does not retain the original output colors. 

Scarb.toml: 
```
[package]
name = "bug_project"
version = "0.1.0"

[scripts]
test = 'snforge'
```
Running `snforge` : 
image

Running `scarb test` (test is defined as snforge): 
image

### Proposed Solution

Make Scarb preserve the colors.

### Notes

_No response_

MrDenkoV:
Fix #598

### DIFF
--- a/scarb/src/core/manifest/scripts.rs
+++ b/scarb/src/core/manifest/scripts.rs
@@ -18,7 +18,7 @@ impl ScriptDefinition {
         let additional_args = args
             .iter()
             // surround all the additional arguments in double quotes
-            // and santize any command substition
+            // and sanitize any command substitution
             .map(|a| {
                 format!(
                     "\"{}\"",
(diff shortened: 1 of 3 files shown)


mkaput (changes requested):
As discussed f2f, this isn't the solution for the problem. Instead we should detect if `deno_task_shell` script is pipe-less, and simply call `exec_replace` in that case. We do not want to break pipes in scripts etc.

mkaput (changes requested):
Looks good! I have just some nitpicks

mkaput (approved):

mkaput (approved):

PR #710, opened 2023-09-25
Implement normalization of manifests for publishing

mkaput:
**Stack**:
- #758
- #757
- #748
- #746
- #733
- #697
- #732
- #710 ⬅


scarb/src/core/publishing/manifest_normalization.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    Ok(TomlManifest {
+        package,
+        workspace: None,
+        dependencies,
+        lib: None,
+        // TODO(mkaput): Allow packaging Cairo plugins.
+        cairo_plugin: None,
+        test: None,
+        target: None,
```
maciektr, line 37: Shouldn't we include package targets? 🤔

scarb/src/core/publishing/manifest_normalization.rs
```diff
@@ -0,0 +1,146 @@
...
+            let toml_dep = generate_dependency(dep)?;
+            Ok((name, MaybeWorkspace::Defined(toml_dep)))
+        })
+        .collect()
+}
+
+fn generate_dependency(dep: &ManifestDependency) -> Result<TomlDependency> {
+    assert!(
+        !dep.source_id.is_std(),
+        "Built-in dependencies should not be included in published manifests."
```
maciektr, line 87: Do I understand correctly, that `"starknet"` will not trigger this assert because it is registry dep until patched? Would it make sense to add it to one of the tests to make sure?

scarb/src/core/publishing/manifest_normalization.rs
```diff
@@ -0,0 +1,146 @@
...
+    assert!(
+        path.is_absolute(),
+        "Manifest parser is expected to canonicalize paths for README/LICENSE files."
+    );
+
+    let path = if let Ok(relative_path) = path.strip_prefix(pkg.root()) {
+        relative_path.to_owned()
+    } else {
+        // This path points outside the package root. `scarb package` will copy it
+        // into the root, so we have to adjust the path to this location.
```
maciektr, line 139: Is this implemented already? 🤔

maciektr (approved):

scarb/src/core/publishing/manifest_normalization.rs
mkaput: Nope, this will be implemented by next PR in the stack

scarb/src/core/publishing/manifest_normalization.rs
mkaput: It is tested: https://github.com/software-mansion/scarb/blob/e209612a1f8ed85a186720c8435c5fba49ff6c71/scarb/tests/package.rs#L325-L326

scarb/src/core/publishing/manifest_normalization.rs
mkaput: They are not needed currently. We are only interested in dependencies, which are inconstructible currently anyway, so I didn't bother with them here.

maciektr (approved):

szymmis (approved):

PR #722, opened 2023-09-27
Build packages filter

maciektr:
commit-id:d571a145

---

**Stack**:
- #727
- #723
- #722 ⬅

### DIFF
--- a/utils/scarb-ui/src/args/packages_filter.rs
+++ b/utils/scarb-ui/src/args/packages_filter.rs
@@ -1,4 +1,5 @@
 use std::collections::HashSet;
+use std::ffi::OsString;
 use std::fmt;
 
 use anyhow::{bail, ensure, Result};
@@ -6,6 +7,8 @@ use camino::{Utf8Path, Utf8PathBuf};
 use indoc::{formatdoc, indoc};
 
 use scarb_metadata::{Metadata, PackageMetadata};
+
+const PACKAGES_FILTER_DELIMITER: char = ',';
 
 /// [`clap`] structured arguments that provide package selection.
 ///
@@ -27,7 +30,7 @@ pub struct PackagesFilter {
         short,
         long,
         default_value = "*",
-        value_delimiter = ',',
+        value_delimiter = PACKAGES_FILTER_DELIMITER,
         value_name = "SPEC",
         env = "SCARB_PACKAGES_FILTER"
     )]
@@ -67,7 +70,7 @@ impl PackagesFilter {
             .iter()
             .map(|s| s.to_string())
             .collect::<Vec<_>>()
-            .join(",");
+            .join(PACKAGES_FILTER_DELIMITER.to_string().as_str());
         let found = Self::do_match_all::<S>(specs, self.workspace, members)?;
 
         ensure!(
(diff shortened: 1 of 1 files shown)


utils/scarb-ui/src/args/packages_filter.rs
```diff
@@ -99,6 +102,23 @@ impl PackagesFilter {
         Self::do_match_all::<S>(specs, self.workspace, members)
     }
 
+    /// Get the packages filter as a OsString.
+    ///
+    /// This value can be passed as `SCARB_PACKAGES_FILTER` variable to child processes.
+    pub fn build_packages_filter<S: PackagesSource>(&self, source: &S) -> Result<OsString> {
```
mkaput, line 108: I don't feel convinced this will be a good API. Mind that this is a library code, we need to think about such things.


I wonder if it wouldn't make sense to split this into two phases:

```rust
pub fn generate_for<S: PackagesSource>(source: &S) -> Self { todo!() }
pub fn to_env(&self) -> OsString { todo!() }

impl PackagesSource for &[Package] {}
```

This way, we can technically generate packages filter for an already filtered workspace without bothering who actually did the filtering. Wdyt?

mkaput (changes requested):

utils/scarb-ui/src/args/packages_filter.rs
```diff
@@ -99,6 +102,30 @@ impl PackagesFilter {
         Self::do_match_all::<S>(specs, self.workspace, members)
     }
 
+    /// Generate a new [`PackagesFilter`] for the given slice  of packages.
+    ///
+    /// This is useful when you want to build an env filter from matched packages.
+    /// See [`PackagesFilter::to_env`] for more details.
+    pub fn generate_for<S: PackagesSource>(packages: &[S::Package]) -> Self {
```
mkaput, line 109:
```suggestion
    pub fn generate_for<S: PackagesSource>(packages: impl Iterator<Item=S::Package>) -> Self {
```

utils/scarb-ui/src/args/packages_filter.rs
```diff
@@ -99,6 +102,30 @@ impl PackagesFilter {
...
+            .iter()
+            .map(|p| S::package_name_of(p).to_string())
+            .collect();
+        Self {
+            package: names,
+            workspace: false,
+        }
+    }
+
+    /// Get the packages filter as an OsString.
```
mkaput, line 120:
```suggestion
    /// Get the packages filter as an [`OsString`].
```

utils/scarb-ui/src/args/packages_filter.rs
```diff
@@ -99,6 +102,30 @@ impl PackagesFilter {
...
+    ///
+    /// This value can be passed as `SCARB_PACKAGES_FILTER` variable to child processes.
+    pub fn to_env(self) -> OsString {
```
mkaput, line 123: Hmm, https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.env accepts anything `AsRef<OsStr>`, so this means, we could return plain `String` here, and that API could still be ergonomic in main case.

mkaput (changes requested):

utils/scarb-ui/src/args/packages_filter.rs
maciektr: Yes, we could, OsString emphasizes the point of this function though - I am not sure if we should encourage parsing this string for other use cases 🤔

mkaput: True, let's leave this as-is then

mkaput (approved):

PR #723, opened 2023-09-27
Use packages filter env var in cairo-test

maciektr:
commit-id:83b56a17

---

**Stack**:
- #727
- #723 ⬅

### DIFF
--- a/extensions/scarb-cairo-test/src/main.rs
+++ b/extensions/scarb-cairo-test/src/main.rs
@@ -35,6 +35,14 @@ fn main() -> Result<()> {
 
     check_scarb_version(&metadata);
 
+    let matched = args.packages_filter.match_many(&metadata)?;
+    let filter = PackagesFilter::generate_for::<Metadata>(matched.iter());
+    ScarbCommand::new()
+        .arg("build")
+        .arg("--test")
+        .env("SCARB_PACKAGES_FILTER", filter.to_env())
+        .run()?;
+
     let profile = env::var("SCARB_PROFILE").unwrap_or("dev".into());
     let default_target_dir = metadata.runtime_manifest.join("target");
     let target_dir = metadata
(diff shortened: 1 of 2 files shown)


extensions/scarb-cairo-test/src/main.rs
```diff
@@ -35,6 +35,15 @@ fn main() -> Result<()> {
 
     check_scarb_version(&metadata);
 
+    ScarbCommand::new()
+        .arg("build")
+        .arg("--test")
+        .env(
+            "SCARB_PACKAGES_FILTER",
+            args.packages_filter.build_packages_filter(&metadata)?,
+        )
```
mkaput, line 44: Don't we want to inherit this env from `$SCARB` process?

extensions/scarb-cairo-test/tests/examples.rs
```diff
@@ -8,6 +8,7 @@ use test_for_each_example::test_for_each_example;
 fn cairo_test(example: &Path) {
     Command::new(cargo_bin("scarb"))
         .arg("cairo-test")
+        .arg("--workspace")
```
mkaput, line 11: I don't understand, what does this change?

mkaput (changes requested):

extensions/scarb-cairo-test/tests/examples.rs
maciektr: The workspaces example from https://github.com/software-mansion/scarb/tree/main/examples/workspaces does define the root package, so cairo-test called in workspace root would test only a single package (the root) without passing `--workspace` flag.

extensions/scarb-cairo-test/src/main.rs
maciektr: Actually, I think it shouldn't? 🤔 
- `scarb test` will consume the packages filter and run tests (i.e. `scarb ciaro-test`) on each package separately. Passing  `scarb test` packages filter to `scarb cairo-test` could break this behaviour. 
- `scarb cairo-test` should be able to run workspace tests on it's own, which requires building packages chosen with packages filter through scarb build.

mkaput (approved):

mkaput (approved):

mkaput (approved):

PR #726, opened 2023-09-28
DepBuilder DSL

mkaput:
This PR adds a new DSL for specifying dependencies in `ProjectBuilder` and `WorkspaceBuilder`. It is capable of expressing all cases for which we have been falling back to strings currently, and also it can now do things like amending auto-generated deps for e.g. paths.

This PR updates all string- and TOML table-based usages of old API. It does not migrate explicit paths usage to auto-generation for `ChildPaths` whereverr we want to ensure paths are properly treated as relative.

---

**Stack**:
- #697
- #710
- #726 ⬅


maciektr (approved):
🔥 🚒

PR #743, opened 2023-10-03
Write actual git revision for git source ids

maciektr:
commit-id:557f1ffa

---

**Stack**:
- #745
- #744
- #743 ⬅

### DIFF
--- a/scarb/src/core/source/id.rs
+++ b/scarb/src/core/source/id.rs
@@ -21,7 +21,7 @@ use crate::sources::canonical_url::CanonicalUrl;
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct SourceId(&'static SourceIdInner);
 
-#[derive(Eq, PartialEq, Ord, PartialOrd, Hash)]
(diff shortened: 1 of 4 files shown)


scarb/src/core/source/id.rs
```diff
@@ -30,6 +30,8 @@ pub struct SourceIdInner {
     pub kind: SourceKind,
     /// The canonical URL of this source, used for internal comparison purposes.
     pub canonical_url: CanonicalUrl,
+    /// For example, the exact Git revision of the specified branch for a Git Source.
+    pub precise: Option<String>,
```
mkaput, line 34: Back when initially implementing this struct I was thinking of putting `precise` within `Git` kind because we won't use it elsewhere? Wdyt?

mkaput (changes requested):

scarb/src/core/source/id.rs
maciektr: Tbh I am ok with both of them :)

mkaput (changes requested):

mkaput (approved):

PR #767, opened 2023-10-06
Implement base data models for registry index

mkaput:
**Stack**:
- #794
- #793
- #790
- #792
- #783
- #767 ⬅

### DIFF
--- a/scarb/src/core/registry/index/record.rs
+++ b/scarb/src/core/registry/index/record.rs
@@ -1,0 +1,34 @@
+use semver::{Version, VersionReq};
+use serde::{Deserialize, Serialize};
+
+use crate::core::PackageName;
+
+pub type IndexRecords = Vec<IndexRecord>;
+
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct IndexRecord {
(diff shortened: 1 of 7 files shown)


szymmis (approved):

scarb/src/core/registry/index/record.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub version: Version,
+    #[serde(rename = "deps")]
+    pub dependencies: IndexDependencies,
+    #[serde(rename = "cksum")]
+    pub checksum: String,
+    #[serde(default = "default_false", skip_serializing_if = "is_false")]
+    pub no_core: bool,
+}
+
+pub type IndexDependencies = Vec<IndexDependency>;
```
szymmis, line 20: Is this a common pattern in Rust to alias vectors?

szymmis (approved):

scarb/src/core/registry/index/record.rs
mkaput: It's highly contextual. Here, I do really want to say that the root serde type is a list of index dependencies; hence I defined the type alias, so it's a known name. Also did this so that this `Vec` won't be propagated throughout the codebase.

PR #778, opened 2023-10-10
Generate lock files

maciektr:
commit-id:5ad45dc3

---

**Stack**:
- #781
- #800
- #780
- #779
- #778 ⬅


scarb/src/core/lockfile.rs
```diff
@@ -90,6 +90,22 @@ impl Lockfile {
         Self::new(packages)
     }
 
+    pub fn from_path(path: impl AsRef<Utf8Path>) -> Result<Self> {
+        if path.as_ref().is_file() {
+            let content = fsx::read_to_string(path.as_ref())
+                .with_context(|| format!("Failed to read lockfile at {}", path.as_ref()))?;
```
mkaput, line 96: Unconventional error message

scarb/src/core/lockfile.rs
```diff
@@ -15,7 +15,7 @@ pub enum LockVersion {
     V1 = 1,
 }
 
-#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Eq, PartialEq, Default, Serialize, Deserialize)]
```
mkaput, line 18: Could be part of earlier PR?

scarb/src/core/workspace.rs
mkaput, line 1: Reading and writing lockfile should be OPs IMO

mkaput (changes requested):

scarb/src/ops/lockfile.rs
```diff
@@ -0,0 +1,14 @@
...
+
+pub fn read_lockfile(ws: &Workspace<'_>) -> Result<Lockfile> {
+    Lockfile::from_path(ws.lockfile_path())
+}
+
+pub fn write_lockfile(ws: &Workspace<'_>, lockfile: Lockfile) -> Result<()> {
+    let path = ws.lockfile_path();
+    fsx::write(path, lockfile.render()?.as_bytes())?;
+    Ok(())
+}
```
mkaput, line 14: let's not forget about `#[tracing::instrument]`?

scarb/src/ops/resolve.rs
```diff
@@ -89,7 +91,12 @@ pub fn resolve_workspace(ws: &Workspace<'_>) -> Result<WorkspaceResolve> {
                 .map(|pkg| pkg.manifest.summary.clone())
                 .collect::<Vec<_>>();
 
-            let resolve = resolver::resolve(&members_summaries, &patched, ws.config().ui()).await?;
+            let lockfile: Lockfile = read_lockfile(ws)?;
```
mkaput, line 94: Shouldn't we ignore lockfile parsing errors? At least print them as warnings and overwrite the lockfile afterwards 🤔

mkaput (changes requested):

scarb/src/ops/resolve.rs
maciektr: Hmm, I am not really sure. Ignoring parsing errors would be inconsistent from what Cargo does:
```
error: failed to parse lock file at: /Users/maciektr/Projects/cairo/Cargo.lock

Caused by:
  missing field `name`
  in `package`
```
So requiring an action from user (i.e. to remove / reset lock) may be intentional. 

On the other hand, npm seems to ignore invalid locks and overwrite without an error.

mkaput: Interesting. I suggested this exactly because I followed NPM logic. Didn't know that Cargo is erroring here. So let's stay with Cargo's behaviour, as we used to do. Thx.

scarb/src/ops/lockfile.rs
```diff
@@ -0,0 +1,16 @@
...
+use crate::internal::fsx;
+use anyhow::Result;
+
+#[tracing::instrument(skip_all, level = "debug")]
+pub fn read_lockfile(ws: &Workspace<'_>) -> Result<Lockfile> {
+    Lockfile::from_path(ws.lockfile_path())
+}
+
+#[tracing::instrument(skip_all, level = "debug")]
+pub fn write_lockfile(ws: &Workspace<'_>, lockfile: Lockfile) -> Result<()> {
```
mkaput, line 12:
```suggestion
pub fn write_lockfile(lockfile: Lockfile, ws: &Workspace<'_>) -> Result<()> {
```

mkaput (approved):

scarb/src/ops/lockfile.rs
```diff
@@ -0,0 +1,16 @@
...
+    let path = ws.lockfile_path();
+    fsx::write(path, lockfile.render()?.as_bytes())?;
```
mkaput, line 14: Wait... I think we should lock this file exclusively here to prevent corruption in case Scarb is executed multiple times in parallel... 🤔

mkaput (changes requested):

scarb/src/ops/lockfile.rs
maciektr: Good catch, thank you!

mkaput: Don't forget to also lock in shared mode in read op! 😉

PR #783, opened 2023-10-10
Implement `Checksum` and `Digest` mechanisms

mkaput:
**Stack**:
- #793
- #790
- #792
- #783 ⬅

### DIFF
--- a/scarb/src/core/checksum.rs
+++ b/scarb/src/core/checksum.rs
@@ -1,0 +1,171 @@
+use std::fmt;
+use std::fmt::Write;
+use std::str;
+use std::str::FromStr;
+
+use anyhow::{bail, ensure, Context, Result};
+use data_encoding::{Encoding, HEXLOWER_PERMISSIVE};
+use serde::{Deserialize, Serialize};
+use sha2::Digest as _;
+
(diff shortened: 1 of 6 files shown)


szymmis (approved):

scarb/src/core/checksum.rs
```diff
@@ -0,0 +1,171 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Checksum({self})")
+    }
+}
+
+pub struct Digest(sha2::Sha256);
+
+impl Digest {
+    /// Get recommended digest algorithm.
+    pub fn recommended() -> Self {
```
maciektr, line 114: Why not just default? 🤔

maciektr (approved):

scarb/src/core/checksum.rs
mkaput: I have thought about this and I concluded `default` is not precise enough? It creates an incentive to use default everywhere, while programmers should carefully decide whether they want to create a digest using the currently preferred algorithm, or create one with an algorithm matching the checksum we got to verify.

PR #791, opened 2023-10-12
Allow specifying custom source patches

szymmis:


scarb/src/core/config.rs
```diff
@@ -36,6 +38,7 @@ pub struct Config {
     offline: bool,
     compilers: CompilerRepository,
     cairo_plugins: CairoPluginRepository,
+    custom_source_patches: Option<Vec<ManifestDependency>>,
```
mkaput, line 41: Let's add a comment here and in actual usage place that this is Dojo-specific feature, to be removed when we'll decouple Dojo from Scarb-as-a-library 😃

mkaput (commented):
@tarrencev can you please check this out and verify it works for you? 😃 

I'll approve this PR when we'll get confirmation this works in Sozo.

scarb/src/core/config.rs
```diff
@@ -328,6 +338,11 @@ impl ConfigBuilder {
         self
     }
 
+    pub fn custom_source_patches(mut self, custom_self_patches: Vec<ManifestDependency>) -> Self {
```
tarrencev, line 341:
```suggestion
    pub fn set_custom_source_patches(mut self, custom_self_patches: Vec<ManifestDependency>) -> Self {
```

tarrencev (commented):
This looks good!

Just to be sure, we'll plan to call this when building the config here: https://github.com/dojoengine/dojo/blob/76b6e4ad61fa8b3a54e55c4455aec5bf02521c98/crates/sozo/src/main.rs#L35

scarb/src/core/config.rs
mkaput: actually all methods in this builder skip the `set_` part 😅

mkaput (approved):

PR #790, opened 2023-10-11
Initial implementation of `RegistrySource`

mkaput:
This implementation focuses solely on local registries, and it has almost no security.

---

**Stack**:
- #819
- #818
- #809
- #808
- #807
- #806
- #805
- #803
- #802
- #799
- #798
- #793
- #790 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #794, opened 2023-10-12
Bring back `block_on` helper function

mkaput:
The frequency of `ws.config().tokio_handle().block_on()` spell has increased significantly a lot, thus it makes sense to bring back this helper after months of being buried.

---

**Stack**:
- #794 ⬅
- #793
- #790
- #792
- #783
- #767

### DIFF
--- a/scarb/src/core/config.rs
+++ b/scarb/src/core/config.rs
@@ -338,3 +338,21 @@ impl ConfigBuilder {
         self
     }
 }
+
+pub trait GetConfig {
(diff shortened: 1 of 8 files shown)


maciektr:
TBH I am not really sure if we need this? 

This helper still gets <= 5 uses throughout whole codebase. Also, `block_on` is only a bit shorter, while being less explicit (i.e. the interaction with tokio runtime is hidden from reader's eyes).

mkaput:
As discussed this privately, closing this PR as it's indeed no so much improvement.

PR #798, opened 2023-10-16
Add basic tests for local registry

mkaput:
These tests depend on both #790 and #793, hence they come in separate PR.

---

**Stack**:
- #819
- #818
- #809
- #808
- #807
- #806
- #805
- #803
- #802
- #799
- #798 ⬅

### DIFF
--- a/scarb/tests/local_registry.rs
+++ b/scarb/tests/local_registry.rs
@@ -6,7 +6,122 @@ use url::Url;
 
 use scarb_test_support::command::Scarb;
 use scarb_test_support::fsx::ChildPathEx;
-use scarb_test_support::project_builder::ProjectBuilder;
+use scarb_test_support::project_builder::{Dep, DepBuilder, ProjectBuilder};
+use scarb_test_support::registry::local::LocalRegistry;
+
+#[test]
+fn usage() {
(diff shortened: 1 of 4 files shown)


scarb/tests/local_registry.rs
```diff
@@ -6,7 +6,124 @@ use url::Url;
...
+    ProjectBuilder::start()
+        .name("foo")
+        .version("0.1.0")
+        .dep("bar", Dep.version("1").registry(&registry))
+        .lib_cairo(r#"fn f() -> felt252 { bar::f() }"#)
+        .build(&t);
+
+    // FIXME(mkaput): Why are verbose statuses not appearing here?
+    Scarb::quick_snapbox()
+        .arg("build") // TODO(#137): Change build to check for faster and lighter test.
```
maciektr, line 33: Is `fetch` not enough? I.e. with `[..] Unpacking bar v1.0.0 ([..])` only?

maciektr (approved):

scarb/tests/local_registry.rs
mkaput: Good catch! Will push a fix soon.

PR #800, opened 2023-10-16
Omit std source packages from lockfile

maciektr:
commit-id:ff227207

---

**Stack**:
- #781
- #800 ⬅


scarb/src/core/lockfile.rs
```diff
@@ -245,10 +251,7 @@ mod tests {
 
     #[test]
     fn empty() {
-        let lock = Lockfile {
-            version: Default::default(),
-            packages: Default::default(),
-        };
+        let lock = Lockfile::new(vec![].into_iter());
```
mkaput, line 254:
```suggestion
        let lock = Lockfile::new([]);
```

This should be possible I believe if new takes `impl IntoIter`

mkaput (approved):

PR #803, opened 2023-10-18
Initial implementation of HTTP registry client

mkaput:
**Stack**:
- #819
- #818
- #809
- #808
- #807
- #806
- #805
- #803 ⬅

### DIFF
--- a/scarb/src/core/registry/client/http.rs
+++ b/scarb/src/core/registry/client/http.rs
@@ -1,0 +1,163 @@
+use std::path::PathBuf;
+use std::sync::Arc;
+
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+use fs4::tokio::AsyncFileExt;
+use futures::StreamExt;
+use reqwest::StatusCode;
+use tokio::fs::OpenOptions;
+use tokio::io;
+use tokio::io::BufWriter;
+use tokio::sync::OnceCell;
+use tracing::{debug, trace};
+
+use crate::core::registry::client::RegistryClient;
+use crate::core::registry::index::{IndexConfig, IndexRecords};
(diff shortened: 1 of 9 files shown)


szymmis (approved):

scarb/src/core/registry/client/http.rs
```diff
@@ -0,0 +1,163 @@
...
+                .join(IndexConfig::WELL_KNOWN_PATH)
+                .expect("Registry config URL should always be valid.");
+            debug!("fetching registry config: {index_config_url}");
+
+            let index_config = config
+                .http()?
+                .get(index_config_url)
+                .send()
+                .await?
+                .error_for_status()?
```
maciektr, line 63: Should we have some retry policy for all this calls here?

maciektr (approved):

scarb/src/core/registry/client/http.rs
mkaput: I don't think so. This is not a half-autonomous service, but just a one-off command line application. If something is off, we print the issue to the user, and they interactively decide whether to restart or no.

maciektr: Yes, but on the other hand, we make a bunch of them. That's why I wonder if we might have a problem if one download out of say 20 fails on unreliable network? 🤔

mkaput: With caching, the client will be idempotently dealing with this case and only pulling information it is missing so I think this would mitigate this issue?

mkaput: Also FYI: Cargo doesn't do retries and I don't think any other pkgmgr does

PR #808, opened 2023-10-20
Introduce `RegistryClientCache` stub

mkaput:
This PR lays groundwork for implementing caching of `RegistryClient` instances.
It just sets up all the types and refactors, no caching is done yet.

There is also a small functional change: local registry client does not print
`Unpacking` status in normal mode. This wasn't a big deal, and it allowed removing
the `is_offline` method from `RegistryClient`.

---

**Stack**:
- #892
- #846
- #845
- #844
- #819
- #818
- #809
- #808 ⬅

### DIFF
--- a/scarb/src/core/registry/client/cache.rs
+++ b/scarb/src/core/registry/client/cache.rs
@@ -1,0 +1,66 @@
+use std::path::PathBuf;
+
+use anyhow::{bail, Result};
+use tracing::trace;
+
+use crate::core::registry::client::{RegistryClient, RegistryResource};
+use crate::core::registry::index::IndexRecords;
+use crate::core::{Config, ManifestDependency, PackageId};
+
+pub struct RegistryClientCache<'c> {
+    client: Box<dyn RegistryClient + 'c>,
+    _config: &'c Config,
+}
+
+impl<'c> RegistryClientCache<'c> {
+    pub fn new(client: Box<dyn RegistryClient + 'c>, config: &'c Config) -> Result<Self> {
+        Ok(Self {
+            client,
+            _config: config,
+        })
+    }
+
+    /// Layer over [`RegistryClient::get_records`] that caches the result.
+    ///
+    /// It takes [`ManifestDependency`] instead of [`PackageName`] to allow performing some
+    /// optimizations by pre-filtering index records on cache-level.
+    #[tracing::instrument(level = "trace", skip_all)]
+    pub async fn get_records_with_cache(
+        &self,
+        dependency: &ManifestDependency,
+    ) -> Result<IndexRecords> {
+        match self.client.get_records(dependency.name.clone()).await? {
+            RegistryResource::NotFound => {
+                trace!("package not found in registry, pruning cache");
+                bail!("package not found in registry: {dependency}")
+            }
+            RegistryResource::InCache => {
+                trace!("getting records from cache");
+                todo!()
+            }
+            RegistryResource::Download { resource, .. } => {
+                trace!("got new records, invalidating cache");
+                Ok(resource)
+            }
+        }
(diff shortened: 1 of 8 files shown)


scarb/src/core/registry/client/cache.rs
```diff
@@ -0,0 +1,75 @@
...
+            _config: config,
+        })
+    }
+
+    /// Layer over [`RegistryClient::get_records`] that caches the result.
+    ///
+    /// It takes [`ManifestDependency`] instead of [`PackageName`] to allow performing some
+    /// optimizations by pre-filtering index records on cache-level.
+    #[tracing::instrument(level = "trace", skip_all)]
+    pub async fn get_records_with_cache(
```
maciektr, line 28: Do we need `_with_cache` suffixes here?

scarb/src/sources/registry.rs
```diff
@@ -120,28 +128,36 @@ impl<'c> Source for RegistrySource<'c> {
             .iter()
             // NOTE: We filter based on IndexRecords here, to avoid unnecessarily allocating
             //   PackageIds just to abandon them soon after.
+            // NOTE: Technically, RegistryClientCache may already have filtered the records,
+            //   but it is not required to do so, so we do it here again as a safety measure.
```
maciektr, line 132: I do not see a reason why it would not be required to do so since it gets full ManifestDependency now 🤔

scarb/src/sources/registry.rs
```diff
@@ -120,28 +128,36 @@ impl<'c> Source for RegistrySource<'c> {
...
-            self.config
-                .ui()
-                .print(Status::new("Downloading", &id.to_string()));
-        }
+        let before_network = {
+            let source_id = self.source_id;
+            let network_allowed = self.config.network_allowed();
+            let ui = self.config.ui();
+            Box::new(move || {
+                ensure!(
```
maciektr, line 145: Why isn't this a client's responsibility, to decide if query is possible?

scarb/src/core/registry/client/cache.rs
mkaput: I wanted to be more explicit at usage place, especially as this method has a different signature than the one in `RegistryClient` trait.

scarb/src/sources/registry.rs
mkaput: I would rather not mix this responsibility into client objects and cache. I want to treat cache as a layer over clients that has +/- the same usage contract. Only the cache layer does the filtering, clients themselves do not.

scarb/src/sources/registry.rs
mkaput: Hmm, that's true. My reasoning behind this code was that I wanted to control UI interactions from RegistrySource only rather than smearing it throughout the codebase.

Cargo does smearing and it has a weird `silent` field in clients because it uses clients for waiting for publishing and yet, it doesn't want them to print these “Downloading” statuses along the way.

I'll think about this.

scarb/src/core/registry/client/cache.rs
maciektr: > has a different signature 
This is a strong point indeed

scarb/src/sources/registry.rs
mkaput: Done. I have removed this callback entirely, and replaced it with `Config::online_http` helper and just printing within HTTP's client `download` method. This should be fine even when waiting for downloads, because I realised that logic won't probably need to call download at all, looking at records would be just as fine.

maciektr (approved):
🦾

maciektr (approved):

szymmis (approved):

PR #818, opened 2023-10-23
Set up machinery for testing HTTP interactions of HTTP registry

mkaput:
Signed-off-by: Marek Kaput <marek.kaput@swmansion.com>

---

**Stack**:
- #892
- #846
- #845
- #844
- #819
- #818 ⬅

### DIFF
--- a/utils/scarb-test-support/src/simple_http_server.rs
+++ b/utils/scarb-test-support/src/simple_http_server.rs
@@ -1,3 +1,5 @@
+use std::collections::BTreeMap;
+use std::fmt;
 use std::net::{SocketAddr, TcpListener};
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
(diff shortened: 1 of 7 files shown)


utils/scarb-test-support/src/simple_http_server.rs
```diff
@@ -104,6 +140,22 @@ async fn logger<B>(
...
+    let log = HttpLog {
+        req_method,
+        req_uri,
+        req_headers,
+        res_status,
+        res_headers,
+    };
+
+    {
+        let mut logs = logs.lock().await;
```
maciektr, line 155: Would it make sense, to have some flag that disables logging entirely, so we can avoid the performance hit? (I know this server is primarily a test helper, but might be useful for some quick demo / test 🤔 )

maciektr (approved):

utils/scarb-test-support/src/simple_http_server.rs
mkaput: I doubt this will be a performance issue ever, there are more serious things going on here, like buffering entire response in the middleware. Let's avoid optimizing this part, it's really just meant for very low traffic tests, and it doesn't make sense to add more code here.

If you want a demo, you'd better be with `python3 -m http.server` or Caddy 😉

PR #824, opened 2023-10-24
Allow ignoring example project by name in test_for_each_example macro

maciektr:
We fail to run cairo-test on alexandria with cairo main. This results in failed tests on https://github.com/software-mansion/scarb/pull/821

This PR adds an ignore argument on  test_for_each_example, so we can skip running cairo-test on dependencies example.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ cargo_metadata = "0.18"
 clap = { version = "4", features = ["derive", "env", "string"] }
 clap-verbosity-flag = "2"
 console = "0.15"
+darling = "0.20.3"
 data-encoding = "2"
 deno_task_shell = "0.13"
 derive_builder = "0.12"
--- a/utils/test-for-each-example/src/lib.rs
+++ b/utils/test-for-each-example/src/lib.rs
@@ -2,13 +2,46 @@ use proc_macro::TokenStream;
 use std::path::Path;
 
 use camino::Utf8PathBuf;
+use darling::ast::NestedMeta;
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -46,6 +46,7 @@ cargo_metadata = "0.18"
 clap = { version = "4", features = ["derive", "env", "string"] }
 clap-verbosity-flag = "2"
 console = "0.15"
+darling = "0.20.3"
```
maciektr, line 49: Not sure if we really need this, but I would need help from someone more fluent in rust macros to implement it without this helper.

mkaput: Let's go with it. Otherwise, you would have to parse tokens manually, which would be a nightmare.

utils/test-for-each-example/src/lib.rs
```diff
@@ -2,14 +2,47 @@ use proc_macro::TokenStream;
...
+use darling::ast::NestedMeta;
+use darling::{Error, FromMeta};
 use proc_macro2::{Ident, Span};
 use quote::quote;
 use syn::{parse_macro_input, parse_quote, ItemFn};
 
+#[derive(Debug, FromMeta)]
+struct TestForEachArgs {
+    #[darling(default)]
+    /// Comma-separated list of example project dir names to ignore.
```
mkaput, line 14:
```suggestion
    /// Comma-separated list of example project dir names to ignore.
    #[darling(default)]
```

mkaput (approved):

PR #825, opened 2023-10-25
Make profile args take precedence over env

### LINKED ISSUE #823: Cannot use `--dev` and `--release` flags in run scripts
image

szymmis:
Fixes #823

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -3,6 +3,7 @@
 //! CLI arguments datastructures.
 
 use std::collections::BTreeMap;
+use std::env;
 use std::ffi::OsString;
 
 use anyhow::Result;
@@ -346,7 +347,7 @@ pub struct GitRefGroup {
 #[group(multiple = false)]
 pub struct ProfileSpec {
     /// Specify profile to use by name.
-    #[arg(short = 'P', long, env = "SCARB_PROFILE")]
+    #[arg(short = 'P', long)]
     pub profile: Option<SmolStr>,
     /// Use release profile.
     #[arg(long, hide_short_help = true)]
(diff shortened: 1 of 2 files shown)


scarb/src/bin/scarb/args.rs
```diff
@@ -345,7 +346,7 @@ pub struct GitRefGroup {
 #[group(multiple = false)]
 pub struct ProfileSpec {
     /// Specify profile to use by name.
-    #[arg(short = 'P', long, env = "SCARB_PROFILE")]
+    #[arg(short = 'P', long)]
```
mkaput, line 349: There is a little problem here, we remove information about the environment variable from `--help` :/

Zrzut ekranu 2023-10-25 o 08 53 56

mkaput (changes requested):
I wonder if it wouldn't be better to allow multiple arguments in `ProfileSpec` group and employ overriding logic (`--dev` takes over `--release` which takes over `--profile`). Because we essentialy also have this problem if the user sets `export SCARB_PROFILE=...` in their `.rc` file etc.

scarb/tests/profiles.rs
```diff
@@ -240,6 +240,33 @@ fn cannot_choose_not_existing_profile() {
...
+        .assert()
+        .success();
+
+    Scarb::quick_snapbox()
+        .args(["run", "build-release"])
+        .current_dir(&t)
+        .assert()
+        .success();
+}
+
```
maciektr, line 269: Can we check which profile has been used? 
You can read it from Scarb metadata: https://docs.rs/scarb-metadata/latest/scarb_metadata/struct.Metadata.html#structfield.current_profile

Can we check if cli args take precedence over env var?

maciektr (approved):

mkaput (approved):
👏🏻

PR #827, opened 2023-10-27
Prevent packaging with uncommited Git changes

### LINKED ISSUE #643: Prevent packaging packages with uncommited Git changes, and include VCS information in package tarballs
Implement this: https://github.com/rust-lang/cargo/blob/4820624ce76b5c047566f62d615a3b39eed91501/src/cargo/ops/cargo_package.rs#L498-L637

Add this file as `VCS.json` to the tarball. Use the same format as Cargo has.

Add `--allow-dirty` behaviour.

szymmis:
Resolves #643

### DIFF
--- a/scarb/src/bin/scarb/commands/package.rs
+++ b/scarb/src/bin/scarb/commands/package.rs
@@ -23,7 +23,7 @@ pub fn run(args: PackageArgs, config: &Config) -> Result<()> {
 
     let opts = PackageOpts {
         /// Disable dirty repository checks when printing package files.
-        allow_dirty: if args.list { true } else { args.allow_dirty },
+        allow_dirty: args.list || args.allow_dirty,
     };
 
     if args.list {
--- a/scarb/src/bin/scarb/commands/package.rs
+++ b/scarb/src/bin/scarb/commands/package.rs
@@ -21,7 +21,10 @@ pub fn run(args: PackageArgs, config: &Config) -> Result<()> {
         .map(|p| p.id)
         .collect::<Vec<_>>();
 
-    let opts = PackageOpts;
+    let opts = PackageOpts {
+        /// Disable dirty repository checks when printing package files.
+        allow_dirty: if args.list { true } else { args.allow_dirty },
+    };
 
     if args.list {
         let result = ops::package_list(&packages, &opts, &ws)?;
--- a/scarb/src/ops/package.rs
+++ b/scarb/src/ops/package.rs
@@ -1,6 +1,7 @@
 use std::collections::BTreeMap;
 use std::fs::File;
 use std::io::{Seek, SeekFrom, Write};
+use std::process::Command;
 
 use anyhow::{bail, ensure, Context, Result};
 use camino::Utf8PathBuf;
@@ -8,21 +9,28 @@ use indoc::writedoc;
 
 use scarb_ui::components::Status;
 use scarb_ui::{HumanBytes, HumanCount};
+use serde::Serialize;
 
 use crate::core::publishing::manifest_normalization::prepare_manifest_for_publish;
 use crate::core::publishing::source::list_source_files;
 use crate::core::{Package, PackageId, PackageName, Workspace};
 use crate::flock::FileLockGuard;
 use crate::internal::restricted_names;
-use crate::{ops, MANIFEST_FILE_NAME};
+use crate::{ops, MANIFEST_FILE_NAME, VCS_INFO_FILE_NAME};
 
 const VERSION: u8 = 1;
 const VERSION_FILE_NAME: &str = "VERSION";
 const ORIGINAL_MANIFEST_FILE_NAME: &str = "Scarb.orig.toml";
 
-const RESERVED_FILES: &[&str] = &[VERSION_FILE_NAME, ORIGINAL_MANIFEST_FILE_NAME];
+const RESERVED_FILES: &[&str] = &[
+    VERSION_FILE_NAME,
+    ORIGINAL_MANIFEST_FILE_NAME,
+    VCS_INFO_FILE_NAME,
+];
 
-pub struct PackageOpts;
+pub struct PackageOpts {
+    pub allow_dirty: bool,
+}
 
 /// A listing of files to include in the archive, without actually building it yet.
 ///
@@ -88,10 +96,47 @@ fn before_package(ws: &Workspace<'_>) -> Result<()> {
     Ok(())
 }
 
-#[tracing::instrument(level = "trace", skip(_opts, ws))]
+#[derive(Serialize)]
+struct VcsInfo {
+    sha1: String,
+    path_in_vcs: String,
+}
+
+#[allow(clippy::dbg_macro)]
+fn extract_vcs_info(pkg: &Package, opts: &PackageOpts) -> Result<Option<VcsInfo>> {
+    if let Ok(repo) = gix::discover(pkg.root()) {
+        // There is nothing to parse if repository is empty, so HEAD reference needs to exist.
+        if let Ok(sha1) = repo.rev_parse_single("HEAD") {
+            // `git status -s` output is empty only if there are no changes at all, but always returns success.
+            // `git diff-index --quiet HEAD` returns status code correctly, but doesn't take untracked files into account.
+            let output = Command::new("git")
+                .current_dir(pkg.root())
+                .arg("status")
+                .arg("-s")
+                .output()?;
+
+            if !opts.allow_dirty && !output.stdout.is_empty() {
+                bail!("cannot package a repository containing uncommited changes. Use `--allow-dirty` to ignore this check.")
+            }
+
+            return Ok(Some(VcsInfo {
+                sha1: sha1.to_string(),
+                // Relative path from the repository root to the package root is calculated.
+                path_in_vcs: pkg
+                    .root()
+                    .strip_prefix(repo.work_dir().unwrap())?
+                    .to_string(),
+            }));
+        }
+    }
+
+    Ok(None)
+}
+
+#[tracing::instrument(level = "trace", skip(opts, ws))]
 fn package_one_impl(
     pkg_id: PackageId,
-    _opts: &PackageOpts,
+    opts: &PackageOpts,
     ws: &Workspace<'_>,
 ) -> Result<FileLockGuard> {
     let pkg = ws.fetch_package(&pkg_id)?;
@@ -102,9 +147,9 @@ fn package_one_impl(
 
     // TODO(mkaput): Check metadata
 
-    // TODO(#643): Check dirty in VCS (but do not do it when listing!).
+    let vcs_info = extract_vcs_info(pkg, opts)?;
 
-    let recipe = prepare_archive_recipe(pkg)?;
+    let recipe = prepare_archive_recipe(pkg, vcs_info)?;
     let num_files = recipe.len();
 
     // Package up and test a temporary tarball and only move it to the final location if it actually
@@ -147,15 +192,16 @@ fn package_one_impl(
 
 fn list_one_impl(
     pkg_id: PackageId,
-    _opts: &PackageOpts,
+    opts: &PackageOpts,
     ws: &Workspace<'_>,
 ) -> Result<Vec<Utf8PathBuf>> {
     let pkg = ws.fetch_package(&pkg_id)?;
-    let recipe = prepare_archive_recipe(pkg)?;
+    let vcs_info = extract_vcs_info(pkg, opts)?;
+    let recipe = prepare_archive_recipe(pkg, vcs_info)?;
     Ok(recipe.into_iter().map(|f| f.path).collect())
 }
 
-fn prepare_archive_recipe(pkg: &Package) -> Result<ArchiveRecipe> {
+fn prepare_archive_recipe(pkg: &Package, vcs_info: Option<VcsInfo>) -> Result<ArchiveRecipe> {
     let mut recipe = source_files(pkg)?;
 
     // Sort the recipe before any checks, to ensure generated errors are reproducible.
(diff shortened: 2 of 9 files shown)


scarb/src/bin/scarb/commands/package.rs
```diff
@@ -21,7 +21,10 @@ pub fn run(args: PackageArgs, config: &Config) -> Result<()> {
         .map(|p| p.id)
         .collect::<Vec<_>>();
 
-    let opts = PackageOpts;
+    let opts = PackageOpts {
+        /// Disable dirty repository checks when printing package files.
+        allow_dirty: if args.list { true } else { args.allow_dirty },
```
maciektr, line 26:
```suggestion
        allow_dirty: args.list || args.allow_dirty,
```

scarb/src/ops/package.rs
```diff
@@ -88,10 +96,47 @@ fn before_package(ws: &Workspace<'_>) -> Result<()> {
     Ok(())
 }
 
-#[tracing::instrument(level = "trace", skip(_opts, ws))]
+#[derive(Serialize)]
+struct VcsInfo {
+    sha1: String,
+    path_in_vcs: String,
+}
```
maciektr, line 103: Is there a reason to diverge from Cargo in regards of file format? 🤔 

I.e.
```
#[derive(Serialize)]
struct VcsInfo {
    git: GitVcsInfo,
    /// Path to the package within repo (empty string if root). / not \
    path_in_vcs: String,
}

#[derive(Serialize)]
struct GitVcsInfo {
    sha1: String,
}
```
Serializes to 
```
{
 "git": {
   "sha1": "aac20b6e7e543e6dd4118b246c77225e3a3a1302"
 },
 "path_in_vcs": ""
}
```
While we do
```
{
 "sha1": "aac20b6e7e543e6dd4118b246c77225e3a3a1302",
 "path_in_vcs": ""
}
```

scarb/src/ops/package.rs
```diff
@@ -88,10 +96,47 @@ fn before_package(ws: &Workspace<'_>) -> Result<()> {
...
+            // `git diff-index --quiet HEAD` returns status code correctly, but doesn't take untracked files into account.
+            let output = Command::new("git")
+                .current_dir(pkg.root())
+                .arg("status")
+                .arg("-s")
+                .output()?;
+
+            if !opts.allow_dirty && !output.stdout.is_empty() {
+                bail!("cannot package a repository containing uncommited changes. Use `--allow-dirty` to ignore this check.")
+            }
```
maciektr, line 120:
```suggestion
             ensure!(
                opts.allow_dirty || output.stdout.is_empty(),
                indoc!(
                    r#"
                        cannot package a repository containing uncommited changes
                        help: `--allow-dirty` to ignore this check
                    "#
                )
            );
```

scarb/tests/package.rs
```diff
@@ -458,6 +458,189 @@ fn workspace() {
         .name_and_version("workspace_dep", "1.0.0");
 }
 
+#[test]
```
maciektr, line 461: What would happen, if the package repo has a submodule? 🤔

maciektr (commented):
Minor questions attached :D

scarb/src/ops/package.rs
szymmis: It seems like Cargo want to be prepared for supporting other VCS but we don't, that's why I've decided that this additional key doesn't really bring much onto the table.

scarb/tests/package.rs
szymmis: Could you elaborate on that question? I don't know if that's what you are asking for, but even if the package has submodule, the main one would be packaged.

maciektr: I just wondered whether we are taking into account changes of git submodules, but I see now that it would be reported by `git status -m`, so we should be good to go :)

szymmis: Yeah, at the end of the day it is one big repository. That's why we also discover repository outwards instead of just trying to look for it into the subpackage root, so that CWD doesn't matter when packaging.

scarb/src/ops/package.rs
mkaput: We don't support other VCS, but this doesn't mean we never will, so I'm siding with @maciektr.

scarb/src/ops/package.rs
```diff
@@ -147,15 +198,16 @@ fn package_one_impl(
 
 fn list_one_impl(
     pkg_id: PackageId,
-    _opts: &PackageOpts,
+    opts: &PackageOpts,
     ws: &Workspace<'_>,
 ) -> Result<Vec<Utf8PathBuf>> {
     let pkg = ws.fetch_package(&pkg_id)?;
-    let recipe = prepare_archive_recipe(pkg)?;
+    let vcs_info = extract_vcs_info(pkg, opts)?;
```
mkaput, line 205: you can put this line directly within `prepare_archive_recipe`; no need to put it outside

scarb/src/ops/package.rs
```diff
@@ -88,10 +96,53 @@ fn before_package(ws: &Workspace<'_>) -> Result<()> {
...
 }
 
-#[tracing::instrument(level = "trace", skip(_opts, ws))]
+#[derive(Serialize)]
+struct VcsInfo {
+    sha1: String,
+    path_in_vcs: String,
+}
+
+#[allow(clippy::dbg_macro)]
```
mkaput, line 105:
```suggestion
```

scarb/src/ops/package.rs
```diff
@@ -88,10 +96,53 @@ fn before_package(ws: &Workspace<'_>) -> Result<()> {
...
+                .output()?;
+
+            ensure!(
+                opts.allow_dirty || output.stdout.is_empty(),
+                indoc!(
+                    r#"
+                        cannot package a repository containing uncommited changes
+                        help: `--allow-dirty` to ignore this check
+                    "#
+                )
```
mkaput, line 125:
```suggestion
                indoc! {r#"
                    cannot package a repository containing uncommited changes
                    help: to proceed despite this and include the uncommitted changes, pass the `--allow-dirty` flag
                "#}
```

we need to emphasize that doing `--allow-dirty` is probably a bad idea!

scarb/src/ops/package.rs
```diff
@@ -88,10 +96,53 @@ fn before_package(ws: &Workspace<'_>) -> Result<()> {
...
+    if let Ok(repo) = gix::discover(pkg.root()) {
+        // There is nothing to parse if repository is empty, so HEAD reference needs to exist.
+        if let Ok(sha1) = repo.rev_parse_single("HEAD") {
+            // `git status -s` output is empty only if there are no changes at all, but always returns success.
+            // `git diff-index --quiet HEAD` returns status code correctly, but doesn't take untracked files into account.
+            let output = Command::new("git")
+                .current_dir(pkg.root())
+                .arg("status")
+                .arg("-s")
+                .output()?;
```
mkaput, line 116: This piece of code mixes two stages that the general logic of packaging does:
1. First it tries to **detect** what files are to be included.
2. Second, **only if doing packaging proper**, it computes values of generated files.

This function immediately detects and then computes the value; and then in the `if let Some(vcs_info) = ...` block later on even serializes this file, even in the **first** stage; which is unnecessary.

The boundary between this stages is: 1. is what happens when preparing the recipe, 2. is what happens within the closure passed to `Generated` variant in the recipe file.

Computing the file (i.e. calling `git`) and serializing it should happen in `generated` closure.

scarb/src/ops/package.rs
```diff
@@ -88,10 +96,53 @@ fn before_package(ws: &Workspace<'_>) -> Result<()> {
...
+}
+
+#[allow(clippy::dbg_macro)]
+fn extract_vcs_info(pkg: &Package, opts: &PackageOpts) -> Result<Option<VcsInfo>> {
+    if let Ok(repo) = gix::discover(pkg.root()) {
+        // There is nothing to parse if repository is empty, so HEAD reference needs to exist.
+        if let Ok(sha1) = repo.rev_parse_single("HEAD") {
+            // `git status -s` output is empty only if there are no changes at all, but always returns success.
+            // `git diff-index --quiet HEAD` returns status code correctly, but doesn't take untracked files into account.
+            let output = Command::new("git")
```
mkaput, line 112: Please try to reuse & extend the code used in `GitSource`, it tries to tame user's `git` installation, for example with this envs: https://github.com/software-mansion/scarb/blob/95885ad4913862c7c120cff4ca09a32d75f61bc0/scarb/src/sources/git/client.rs#L333-L347

Don't build `Command` here manually; instead let's maybe try to construct a `GitDatabase` object we have and call its method?

scarb/src/ops/package.rs
```diff
@@ -88,10 +96,53 @@ fn before_package(ws: &Workspace<'_>) -> Result<()> {
...
-#[tracing::instrument(level = "trace", skip(_opts, ws))]
+#[derive(Serialize)]
+struct VcsInfo {
+    sha1: String,
+    path_in_vcs: String,
+}
+
+#[allow(clippy::dbg_macro)]
+fn extract_vcs_info(pkg: &Package, opts: &PackageOpts) -> Result<Option<VcsInfo>> {
+    if let Ok(repo) = gix::discover(pkg.root()) {
```
mkaput, line 107: I don't feel safe with the fact that this code is ignoring all kinds of errors coming from these calls. Wdyt?

mkaput (changes requested):

scarb/src/ops/package.rs
```diff
@@ -88,10 +96,53 @@ fn before_package(ws: &Workspace<'_>) -> Result<()> {
     Ok(())
 }
 
-#[tracing::instrument(level = "trace", skip(_opts, ws))]
+#[derive(Serialize)]
+struct VcsInfo {
+    sha1: String,
+    path_in_vcs: String,
+}
```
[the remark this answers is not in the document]
maciektr: It's seems cargo uses snake_case 😛  https://doc.rust-lang.org/cargo/commands/cargo-package.html#cargo_vcs_infojson-format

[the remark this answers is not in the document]
mkaput: Yes I know, but we generally avoid it like, well, snakes; so maybe let's depart here?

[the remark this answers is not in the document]
mkaput: Ah, it would make it more sense to use kebabs in TOML, but indeed in JSON snake is more palatable

scarb/src/sources/git/client.rs
```diff
@@ -330,6 +331,51 @@ fn collect_refspecs(reference: &GitReference) -> (Vec<String>, bool) {
     }
 }
 
+/// A wrapper over [`scarb::core::Package`] that provides functionality used to gather VCS info.
+pub struct PackageRepository<'a> {
+    pkg: &'a Package,
```
mkaput, line 336: `Package` has inner `Arc`. You can freely store cloned instance inside this.

scarb/src/sources/git/client.rs
```diff
@@ -330,6 +331,51 @@ fn collect_refspecs(reference: &GitReference) -> (Vec<String>, bool) {
     }
 }
 
+/// A wrapper over [`scarb::core::Package`] that provides functionality used to gather VCS info.
+pub struct PackageRepository<'a> {
```
mkaput, line 335: :fire:

scarb/src/sources/git/client.rs
```diff
@@ -330,6 +331,51 @@ fn collect_refspecs(reference: &GitReference) -> (Vec<String>, bool) {
...
+
+        Ok(output.stdout.is_empty())
+    }
+
+    pub fn head_rev_hash(&self) -> Result<String> {
+        Ok(self.repo.rev_parse_single("HEAD")?.to_string())
+    }
+
+    /// Calculate relative path from the repository root to the package root.
+    pub fn path_in_vcs(&self) -> Result<String> {
```
mkaput, line 369:
```suggestion
    pub fn path_in_vcs(&self) -> Result<Utf8PathBuf> {
```

this would make more sense imo

scarb/src/ops/package.rs
```diff
@@ -185,6 +231,21 @@ fn prepare_archive_recipe(pkg: &Package) -> Result<ArchiveRecipe> {
         contents: ArchiveFileContents::Generated(Box::new(|| Ok(VERSION.to_string().into_bytes()))),
     });
 
+    // Add VCS info file.
+    if has_vcs(pkg) {
```
mkaput, line 235: if you `Clone` `pkg`, you can do this:

```suggestion
    if let Ok(repo) = PackageRepository::open(pkg) {
```

and then you can _move_ repo into the closure

this saves you a `gix::discover` call.

scarb/src/ops/package.rs
```diff
@@ -88,10 +98,48 @@ fn before_package(ws: &Workspace<'_>) -> Result<()> {
...
+}
+
+#[derive(Serialize)]
+struct VcsInfo {
+    git: GitVcsInfo,
+    path_in_vcs: String,
+}
+
+fn has_vcs(pkg: &Package) -> bool {
+    gix::discover(pkg.root().to_path_buf()).is_ok()
```
mkaput, line 113: Hmm, I think I'd print the error to logs. Wdyt?

scarb/src/ops/package.rs
```diff
@@ -88,10 +98,48 @@ fn before_package(ws: &Workspace<'_>) -> Result<()> {
...
+fn extract_vcs_info(pkg: &Package, opts: &PackageOpts) -> Result<Option<VcsInfo>> {
+    if let Ok(repo) = PackageRepository::open(pkg) {
+        ensure!(
+            opts.allow_dirty || repo.is_clean()?,
+            indoc!(
+                r#"
+                cannot package a repository containing uncommited changes
+                help: to proceed despite this and include the uncommitted changes, pass the `--allow-dirty` flag
+                "#
+            )
```
mkaput, line 125:
```suggestion
            indoc! {r#"
                cannot package a repository containing uncommited changes
                help: to proceed despite this and include the uncommitted changes, pass the `--allow-dirty` flag
            "#}
```

using `{}` in this context is more idiomatic. one indent less

we use `()` only in the following cases: `formatdoc!("pattern", arg, ...)`

mkaput (changes requested):
This is shaping very nicely 😊

mkaput (approved):
LGTM

PR #843, opened 2023-10-30
Major dependencies update

mkaput:
- `deno_task_shell`
- ran `cargo update`

PR #809, opened 2023-10-20
Implement caching of records returned from registry clients

mkaput:
NOTE: Nothing is cached actually yet, because no registry client uses hooks provided by this code. This will come in subsequent PRs.

---

**Stack**:
- #892
- #846
- #845
- #844
- #819
- #818
- #809 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #846, opened 2023-10-30
Implement caching of downloaded tarballs

mkaput:
Signed-off-by: Marek Kaput <marek.kaput@swmansion.com>

---

**Stack**:
- #846 ⬅
- #845
- #844
- #843
- #819
- #818
- #809
- #808
- #807


scarb/src/core/registry/client/cache.rs
```diff
@@ -154,6 +175,28 @@ impl<'c> RegistryClientCache<'c> {
             })
             .await
     }
+
+    /// Check whether the tarball has already been downloaded in the _dl_ directory.
+    /// If user modifies downloaded archive, it's their problem.
```
maciektr, line 180: > If user modifies downloaded archive

Wouldn't checksum verification fail in this instance?

maciektr (approved):

scarb/src/core/registry/client/cache.rs
mkaput: Exactly

szymmis (approved):

scarb/src/core/registry/client/cache.rs
mkaput: #906 changes the code, so that checksums are not verified for already downloaded packages, so this comment now will hold exactly as it says

PR #851, opened 2023-10-31
Use rustls-tls with native roots on unix targets

maciektr:
commit-id:01e6ae9b

---

**Stack**:
- #852
- #851 ⬅

### DIFF
--- a/scarb/Cargo.toml
+++ b/scarb/Cargo.toml
@@ -76,6 +76,12 @@ xxhash-rust.workspace = true
 zip.workspace = true
 zstd.workspace = true
 
+[target.'cfg(not(unix))'.dependencies]
+reqwest = { workspace = true, default-features = true}
+
+[target.'cfg(unix)'.dependencies]
+reqwest = { workspace = true, features = ["rustls-tls-native-roots"] }
+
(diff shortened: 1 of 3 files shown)


scarb/Cargo.toml
```diff
@@ -76,6 +76,12 @@ xxhash-rust.workspace = true
 zip.workspace = true
 zstd.workspace = true
 
+[target.'cfg(not(unix))'.dependencies]
+reqwest = { workspace = true, default-features = true}
+
+[target.'cfg(unix)'.dependencies]
+reqwest = { workspace = true, features = ["rustls-tls-native-roots"] }
```
maciektr, line 83: Unfortunately, afaik we cannot switch based on target arch from workspace dependencies. 
Note: features are additive, so we inherit all specified in workspace manifest as well as those added here.

maciektr:
I will merge it for now, so we can build nightlies during the _all saints day_ break.

scarb/Cargo.toml
mkaput, line 83: This worked well on macOS, which is also captured by `unix` cfg. Please send a PR which changes this to be Linux-specific

PR #852, opened 2023-10-31
Add dry_run arg to nightly workflow

maciektr:
commit-id:707851ae

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -2,6 +2,11 @@ name: Nightly
 
 on:
   workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "Dry run"
+        type: boolean
+        default: true
   schedule:
     - cron: '0 0 * * 3,6'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -163,7 +170,8 @@ jobs:
 
   notify_failed:
     runs-on: ubuntu-latest
-    if: always() && contains(needs.*.result, 'failure')
+    # Do not run on dry_run or success
+    if: ${{ !(inputs.dry_run) }} && contains(needs.*.result, 'failure')
```
mkaput, line 174: Don't you have to put `always` anyway here so that GH won't skip this job if dependencies fail? I'm not knowledgeable enough to be certain here.

mkaput (approved):

.github/workflows/nightly.yml
maciektr: I will add it then

PR #856, opened 2023-11-03
Throw an error in `scarb package` with non-`[lib]` packages

### LINKED ISSUE #735: Non-`[lib]` packages should not be possible to package
### Problem

Right now, `scarb package` happily processes packages that do not provide a `lib` target.

### Steps

_No response_

### Possible Solution(s)

[`ensure!`](https://docs.rs/anyhow/latest/anyhow/macro.ensure.html) in [`ops::package` and `ops::package_list`](https://github.com/software-mansion/scarb/blob/f307c4c6f794617446161edccc5459f78c8995b7/scarb/src/ops/package.rs#L52-L74), that the `Package` does indeed have a `lib` target. Try to provide help to the user what has to be changed in `Scarb.toml`, you can see an example how such help could look like [here](https://github.com/software-mansion/scarb/blob/45b936aa7cbedb07ea436e196f994f5965222b0a/scarb/src/ops/compile.rs#L70-L83).

### Notes

_No response_

### Version

_No response_

tomek0123456789:
Closes #735

### DIFF
--- a/scarb/src/ops/package.rs
+++ b/scarb/src/ops/package.rs
@@ -4,7 +4,7 @@ use std::io::{Seek, SeekFrom, Write};
 
 use anyhow::{bail, ensure, Context, Result};
 use camino::Utf8PathBuf;
-use indoc::{indoc, writedoc};
+use indoc::{formatdoc, indoc, writedoc};
 
 use crate::sources::client::PackageRepository;
 
@@ -192,6 +192,19 @@ fn list_one_impl(
 }
 
 fn prepare_archive_recipe(pkg: &Package, opts: &PackageOpts) -> Result<ArchiveRecipe> {
+    ensure!(
+        pkg.manifest.targets.iter().any(|x| x.is_lib()),
+        formatdoc!(
+            r#"
+            cannot archive package `{package_name}` without a `lib` target
+            help: add `[lib]` section to package manifest
+             --> Scarb.toml
+            +   [lib]
+            "#,
+            package_name = pkg.id.name,
+        )
+    );
+
     let mut recipe = source_files(pkg)?;
 
     // Sort the recipe before any checks, to ensure generated errors are reproducible.
--- a/scarb/tests/package.rs
+++ b/scarb/tests/package.rs
@@ -1039,3 +1039,28 @@ fn ignore_whitelist_pattern() {
             src/lib.cairo
         "#}));
 }
+
+#[test]
+fn no_lib_target() {
+    let t = TempDir::new().unwrap();
+    ProjectBuilder::start()
+        .name("foo")
+        .version("1.0.0")
+        .manifest_extra(indoc! {r#"
+        [[target.starknet-contract]]
+        "#})
+        .build(&t);
+
+    Scarb::quick_snapbox()
+        .arg("package")
+        .current_dir(&t)
+        .assert()
+        .failure()
+        .stdout_matches(indoc! {r#"
+        [..] Packaging foo v1.0.0 [..]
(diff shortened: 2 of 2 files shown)


scarb/src/ops/package.rs
```diff
@@ -156,6 +156,20 @@ fn list_one_impl(
...
 fn prepare_archive_recipe(pkg: &Package) -> Result<ArchiveRecipe> {
+    ensure!(
+        pkg.manifest.targets.iter().any(|x| x.is_lib()),
+        formatdoc! {
+            r#"
+            package `{package_name}` does not provide a `[lib]` target.
+            if a package does not provide a library target, it cannot be used as a dependency.
+            help: add `[lib]` section to package manifest
+             --> Scarb.toml
+            +   [lib]
```
maciektr, line 167: - By convention, we do not use dots in error messages
- The `dependency` is confusing - we are only packaging local code
- We need to emphasize why we need to return an error - not defining `lib` is not an error per se, without the context. 
Maybe something like this? Not sure on exact wording.
```suggestion
            cannot archive package `{package_name}` without a `lib` target
            help: add `[lib]` section to package manifest
             --> Scarb.toml
            +   [lib]
```

maciektr (commented):
Great job!!! 🔥

scarb/src/ops/package.rs
```diff
@@ -156,6 +156,20 @@ fn list_one_impl(
 }
 
 fn prepare_archive_recipe(pkg: &Package) -> Result<ArchiveRecipe> {
+    ensure!(
+        pkg.manifest.targets.iter().any(|x| x.is_lib()),
+        formatdoc! {
+            r#"
+            package `{package_name}` does not provide a `[lib]` target.
+            if a package does not provide a library target, it cannot be used as a dependency.
```
mkaput, line 164:
```suggestion
            note: if a package does not provide a library target, it cannot be used as a dependency.
```

scarb/src/ops/package.rs
```diff
@@ -156,6 +156,20 @@ fn list_one_impl(
 }
 
 fn prepare_archive_recipe(pkg: &Package) -> Result<ArchiveRecipe> {
+    ensure!(
+        pkg.manifest.targets.iter().any(|x| x.is_lib()),
+        formatdoc! {
```
mkaput, line 161:
```suggestion
        formatdoc!(
```

we use `{` for `indoc!` et all when there is a single argument. then, rustfmt glues the brace and string literal together instead of making an extra indentation

here we indent anyway, so let's go with `(` which is better handled by rustfmt

scarb/tests/package.rs
```diff
@@ -837,3 +837,32 @@ fn ignore_whitelist_pattern() {
...
+        .failure()
+        .stdout_matches(formatdoc!(
+            r#"
+        [..] Packaging {package_name} v{version} [..]
+        error: cannot archive package `{package_name}` without a `lib` target
+        help: add `[lib]` section to package manifest
+         --> Scarb.toml
+        +   [lib]
+        "#
+        ));
```
mkaput, line 867:
```suggestion
        .stdout_matches(indoc! {r#"
        [..] Packaging foo v1.0.0 [..]
        error: cannot archive package `foo` without a `lib` target
        help: add `[lib]` section to package manifest
         --> Scarb.toml
        +   [lib]
        "#});
```

😆

mkaput (approved):

mkaput (approved):
LGTM

maciektr (approved):
❤️

PR #858, opened 2023-11-05
Add package edition field

maciektr:


scarb-metadata/src/lib.rs
```diff
@@ -208,6 +208,9 @@ pub struct PackageMetadata {
     /// Package version as given in `Scarb.toml`.
     pub version: Version,
 
+    /// Package edition as given in `Scarb.toml`.
+    pub edition: String,
```
mkaput, line 212: Adding new fields without serde falling back is a breaking change here. I think we should use `Option<String>` here and skip serializing if field is `None`. Also don't forget about deserializing from older Scarbs

Cargo.toml
```diff
@@ -82,6 +82,7 @@ serde-value = "0.7"
 serde_json = "1"
 serde_repr = "0.1"
 serde_test = "1"
+serde_variant = "0.1.2"
```
mkaput, line 85:
```suggestion
serde_variant = "0.1"
```

scarb/src/ops/metadata.rs
```diff
@@ -125,10 +126,13 @@ fn collect_package_metadata(package: &Package) -> m::PackageMetadata {
         .build()
         .unwrap();
 
+    let edition = to_variant_name(&package.manifest.edition).unwrap();
```
mkaput, line 129: hmm, why? what's the difference from just serializing the edition value?

scarb/tests/metadata.rs
```diff
@@ -1112,3 +1112,29 @@ fn infer_readme_workspace() {
...
+        .arg("--json")
+        .arg("metadata")
+        .arg("--format-version")
+        .arg("1")
+        .current_dir(&t)
+        .stdout_json::<Metadata>();
+
+    for package in metadata.packages {
+        if package.name == "hello" {
+            assert_eq!(package.edition, "2023_10".to_string());
```
mkaput, line 1135: nitpick, no need to change

```suggestion
            assert_eq!(&package.edition, "2023_10");
```

scarb/tests/build.rs
```diff
@@ -635,3 +636,54 @@ fn workspace_as_dep() {
...
+            snapbox::assert_matches(
+                r#"[..]ProjectConfig { base_path: "[..]", corelib: Some(Real("[..]")), content: ProjectConfigContent { crate_roots: OrderedHashMap({"pkg0": "[..]"}), crates_config: AllCratesConfig { global: SingleCrateConfig { edition: V2023_01 }, override_map: OrderedHashMap({"pkg0": SingleCrateConfig { edition: V2023_01 }, "core": SingleCrateConfig { edition: V2023_01 }}) } } }[..]"#,
+                line.as_ref(),
+            );
+            return;
+        }
+    }
+
+    panic!("Failed to find project_config in stderr");
+}
```
mkaput, line 667: I find this test overly complicated and prone to errors if we change something in trace logs. It's smart, I can't say it isn't, but I don't feel easy merging it.

Why not simply check if a piece of code using edition-specific feature is just compiling successfully?

scarb/src/compiler/db.rs
```diff
@@ -92,13 +96,31 @@ fn build_project_config(unit: &CompilationUnit) -> Result<ProjectConfig> {
...
+                    edition: component.package.manifest.edition,
+                },
+            )
+        })
+        .collect();
+
+    let crates_config = AllCratesConfig {
+        global: SingleCrateConfig::default(),
+        override_map: crates_config,
+    };
```
mkaput, line 115:
```suggestion
    let crates_config = AllCratesConfig {
        override_map: crates_config,
        ..Default::default()
    };
```

scarb/src/compiler/db.rs
mkaput, line 1: Don't we have to set edition for `core` upstream in its Scarb.toml?

mkaput (changes requested):

scarb/src/ops/metadata.rs
maciektr:
```
#[test]
fn test() {
    let e = Edition::V2023_10;
    assert_eq!(to_variant_name(&e).unwrap(), "2023_10");
    assert_eq!(serde_json::to_string(&e).unwrap(), "2023_10");
}
```
The second assert fails with 
```
assertion `left == right` failed
  left: "\"2023_10\""
 right: "2023_10"
```

scarb/src/compiler/db.rs
maciektr: Yes, we probably should

scarb/src/ops/metadata.rs
mkaput: I think we could avoid this dependency by serializing to `serde_json::Value` and then deserializing as `String`. This way we could also better handle cases `Edition` serialization produces something unexpected—current code would produce broken output, while with my suggestion, we would early panic instead.

mkaput (approved):
LGTM 🔥

PR #860, opened 2023-11-06
Support plugin suites

maciektr:


scarb/src/compiler/plugin/builtin.rs
```diff
@@ -1,25 +1,19 @@
...
-use cairo_lang_test_plugin::TestPlugin;
+use cairo_lang_defs::plugin::{MacroPlugin, PluginSuite};
+use cairo_lang_starknet::starknet_plugin_suite;
+use cairo_lang_test_plugin::test_plugin_suite;
 
 use crate::compiler::plugin::{CairoPlugin, CairoPluginInstance};
 use crate::core::{PackageId, PackageName, SourceId};
 use crate::internal::to_version::ToVersion;
 
 impl CairoPluginInstance for Arc<dyn MacroPlugin> {
```
mkaput, line 12: Are we using this impl anywhere anymore? Maybe it'll be worth dropping it

maciektr: I don't think so!

PR #863, opened 2023-11-06
feat: add emit option

Eikix:
- add an emit option to stdout to match some linter specs
- modify FmtAction to have an exclusive option to: fix, check or emit formatted file.
- not modify the existing flag --check for clarity, though --check and --emit flags are mutually exclusive (--check overrides --emit)

Requesting help for: getting familiar with the repository: 
- what should the clap argument for emit be? is this fine?
```rust
pub struct FmtArgs {
    // ...
    #[arg(short, long)]
    pub emit: Option<EmitTarget>,
    // ...
}
```
- How should the `emit_formatted_file` function behave? Should it output a specific per-path message like the diff printing function, i.e.: `ws.config().ui().print(format!("{file}"))` versus `ws.config().ui().print(format!("Diff in file {}:\n {}", path.display(), diff))`

- How should I test the feature?

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -8,6 +8,7 @@ use std::ffi::OsString;
 use anyhow::Result;
 use camino::Utf8PathBuf;
 use clap::{CommandFactory, Parser, Subcommand};
+use scarb::ops::EmitTarget;
 use smol_str::SmolStr;
 use tracing::level_filters::LevelFilter;
 use tracing_log::AsTrace;
@@ -244,6 +245,9 @@ pub struct FmtArgs {
     /// Only check if files are formatted, do not write the changes to disk.
     #[arg(short, long, default_value_t = false)]
     pub check: bool,
+    /// Emit the formatted file to stdout
+    #[arg(short, long)]
+    pub emit: Option<EmitTarget>,
     /// Do not color output.
     #[arg(long, default_value_t = false)]
     pub no_color: bool,
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -243,7 +243,7 @@ pub struct NewArgs {
 #[derive(Parser, Clone, Debug)]
 pub struct FmtArgs {
     /// Only check if files are formatted, do not write the changes to disk.
-    #[arg(short, long, default_value_t = false)]
+    #[arg(short, long, default_value_t = false, conflicts_with = "emit")]
     pub check: bool,
     /// Emit the formatted file to stdout
     #[arg(short, long)]
--- a/scarb/src/bin/scarb/commands/fmt.rs
+++ b/scarb/src/bin/scarb/commands/fmt.rs
@@ -10,7 +10,6 @@ pub fn run(args: FmtArgs, config: &Config) -> Result<()> {
     // The action the formatted should perform,
     // e.g. check formatting, format in place, or emit formatted file to stdout.
     let action = if args.check {
-        // adding the `--check` flag will shortcircuit the ability to emit the formatted file
         FmtAction::Check
     } else if let Some(emit_target) = args.emit {
         FmtAction::Emit(emit_target)
--- a/scarb/src/bin/scarb/commands/fmt.rs
+++ b/scarb/src/bin/scarb/commands/fmt.rs
@@ -3,10 +3,21 @@ use anyhow::Result;
 use crate::args::FmtArgs;
 use crate::errors::error_with_exit_code;
 use scarb::core::Config;
-use scarb::ops;
+use scarb::ops::{self, FmtAction};
 
 #[tracing::instrument(skip_all, level = "info")]
 pub fn run(args: FmtArgs, config: &Config) -> Result<()> {
+    // The action the formatted should perform,
+    // e.g. check formatting, format in place, or emit formatted file to stdout.
+    let action = if args.check {
+        // adding the `--check` flag will shortcircuit the ability to emit the formatted file
+        FmtAction::Check
+    } else if let Some(emit_target) = args.emit {
+        FmtAction::Emit(emit_target)
+    } else {
+        // Format in place is the default option
+        FmtAction::Fix
+    };
     let ws = ops::read_workspace(config.manifest_path(), config)?;
     let packages = args
         .packages_filter
@@ -17,7 +28,7 @@ pub fn run(args: FmtArgs, config: &Config) -> Result<()> {
     if ops::format(
         ops::FmtOptions {
             packages,
-            check: args.check,
+            action,
             color: !args.no_color,
         },
         &ws,
--- a/scarb/src/ops/fmt.rs
+++ b/scarb/src/ops/fmt.rs
@@ -4,6 +4,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 
 use anyhow::Result;
 use cairo_lang_formatter::{CairoFormatter, FormatOutcome, FormatterConfig};
+use clap::ValueEnum;
 use ignore::WalkState::{Continue, Skip};
 use ignore::{DirEntry, Error, ParallelVisitor, ParallelVisitorBuilder, WalkState};
 use tracing::{info, warn};
@@ -12,9 +13,24 @@ use crate::core::workspace::Workspace;
 use crate::core::PackageId;
 use crate::internal::serdex::toml_merge;
 
+#[derive(Debug, Clone, ValueEnum)]
+pub enum EmitTarget {
+    Stdout,
+}
+
+/// Format option to display the output formatted file
+/// Example: scarb fmt --emit stdout
+#[derive(Debug, Default, Clone)]
+pub enum FmtAction {
+    #[default]
+    Fix,
+    Check,
+    Emit(EmitTarget),
+}
+
 #[derive(Debug)]
 pub struct FmtOptions {
-    pub check: bool,
+    pub action: FmtAction,
     pub packages: Vec<PackageId>,
     pub color: bool,
 }
@@ -92,6 +108,10 @@ fn print_diff(ws: &Workspace<'_>, path: &Path, diff: impl Display) {
         .print(format!("Diff in file {}:\n {}", path.display(), diff));
 }
 
+fn print_formatted_file(ws: &Workspace<'_>, path: &Path, file: impl Display) {
+    ws.config().ui().print(format!("{file}"));
+}
+
 fn print_error(ws: &Workspace<'_>, path: &Path, error: anyhow::Error) {
     let error_msg = error.to_string();
     ws.config().ui().error(format!(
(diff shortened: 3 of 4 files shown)


scarb/src/bin/scarb/commands/fmt.rs
```diff
@@ -3,10 +3,21 @@ use anyhow::Result;
...
 use crate::errors::error_with_exit_code;
 use scarb::core::Config;
-use scarb::ops;
+use scarb::ops::{self, FmtAction};
 
 #[tracing::instrument(skip_all, level = "info")]
 pub fn run(args: FmtArgs, config: &Config) -> Result<()> {
+    // The action the formatted should perform,
+    // e.g. check formatting, format in place, or emit formatted file to stdout.
+    let action = if args.check {
```
Eikix, line 12: this is meant to allow people to do `scarb fmt --check`, `scarb fmt --emit stdout` as well as `scarb fmt --check --emit stdout` and still obtain coherent results (check has priority over emit)

scarb/src/ops/fmt.rs
```diff
@@ -12,9 +13,24 @@ use crate::core::workspace::Workspace;
 use crate::core::PackageId;
 use crate::internal::serdex::toml_merge;
 
+#[derive(Debug, Clone, ValueEnum)]
+pub enum EmitTarget {
```
Eikix, line 17: In case someone wants to implement other emit targets

scarb/src/ops/fmt.rs
```diff
@@ -12,9 +13,24 @@ use crate::core::workspace::Workspace;
...
+    Stdout,
+}
+
+/// Format option to display the output formatted file
+/// Example: scarb fmt --emit stdout
+#[derive(Debug, Default, Clone)]
+pub enum FmtAction {
+    #[default]
+    Fix,
```
Eikix, line 26: Can rename `Fix` into `FormatInPlace`

scarb/src/ops/fmt.rs
```diff
@@ -12,9 +13,24 @@ use crate::core::workspace::Workspace;
...
+    Check,
+    Emit(EmitTarget),
+}
+
 #[derive(Debug)]
 pub struct FmtOptions {
-    pub check: bool,
+    pub action: FmtAction,
```
Eikix, line 33: I think this is cleaner with regards to business logic now that there is a third option (fix, check or emit)

scarb/src/ops/fmt.rs
```diff
@@ -92,6 +108,10 @@ fn print_diff(ws: &Workspace<'_>, path: &Path, diff: impl Display) {
         .print(format!("Diff in file {}:\n {}", path.display(), diff));
 }
 
+fn print_formatted_file(ws: &Workspace<'_>, path: &Path, file: impl Display) {
```
Eikix, line 111: not so sure about this

maciektr: One way to avoid this, would be to create some trait like `Emittable`, which would have an interface for emitting results of formatting. Then we could pass a zero sized struct implementing this method to `emit_formatted_file` instead of doing match at `159`.

We have done something similar in the compiler repo  https://github.com/starkware-libs/cairo/blob/f0c9a677fcea54f567c707f25f22a507560af5e4/crates/cairo-lang-formatter/src/cairo_formatter.rs#L87C3-L141

scarb/src/ops/fmt.rs
maciektr: 👍

scarb/src/bin/scarb/commands/fmt.rs
maciektr: Why not disallow using both `--check` and `--emit` simultaneously in clap then?

maciektr:
Please add some e2e tests for this functionality. Take a look at https://github.com/software-mansion/scarb/blob/main/scarb/tests/fmt.rs

maciektr (commented):
Incredible job!

scarb/src/bin/scarb/commands/fmt.rs
Eikix: I don't know how to do it, but I'll investigate!

maciektr: You can take a look at https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with

scarb/src/ops/fmt.rs
Eikix: Ok let me try this

scarb/tests/fmt.rs
```diff
@@ -270,3 +303,77 @@ fn workspace_with_root() {
...
+
+    Scarb::quick_snapbox()
+        .arg("fmt")
+        .arg("--emit")
+        .arg("stdout")
+        .current_dir(&t)
+        .assert()
+        .failure()
+        .stdout_matches(indoc! {"\
+            fn main() -> felt252 {
```
Eikix, line 340: Added a test,
It's unclear to me why I can't pass `SIMPLE_FORMATTED` here, for lack of a newline at end of line

scarb/tests/fmt.rs
```diff
@@ -270,3 +303,77 @@ fn workspace_with_root() {
...
+
+    let content = fs::read_to_string(t.child("src/lib.cairo")).unwrap();
+    assert_eq!(content, SIMPLE_ORIGINAL);
+    let content = fs::read_to_string(t.child("first/src/lib.cairo")).unwrap();
+    assert_eq!(content, SIMPLE_ORIGINAL);
+    let content = fs::read_to_string(t.child("second/src/lib.cairo")).unwrap();
+    assert_eq!(content, SIMPLE_ORIGINAL);
+
+    Scarb::quick_snapbox()
+        .args(["fmt", "--workspace", "--emit", "stdout"])
```
Eikix, line 354: added a test for workspace,
it's unclear to me what behaviour the stdout should have for formatting multiple files at once,
i will search what the linters expect, maybe that some separator symbol and file path be introduced?

scarb/src/ops/fmt.rs
```diff
@@ -127,6 +143,37 @@ fn check_file_formatting(
     }
 }
 
+pub trait Emittable {
+    fn emit(&self, ws: &Workspace<'_>, formatted: &str);
+}
+
+impl Emittable for EmitTarget {
```
Eikix, line 150: didn't go the empty struct route, 
this felt more natural, but open to challenges

scarb/src/bin/scarb/args.rs
```diff
@@ -242,8 +243,11 @@ pub struct NewArgs {
 #[derive(Parser, Clone, Debug)]
 pub struct FmtArgs {
     /// Only check if files are formatted, do not write the changes to disk.
-    #[arg(short, long, default_value_t = false)]
+    #[arg(short, long, default_value_t = false, conflicts_with = "emit")]
```
Eikix, line 246: added conflicts with

scarb/tests/fmt.rs
maciektr: It just stems from usage of `println` internally by Scarb ui https://github.com/software-mansion/scarb/blob/333cdb048eab98d5d331aadc16fcd78c25a4985e/utils/scarb-ui/src/message.rs#L33C13-L33C20

You can use `format!("{SIMPLE_FORMATTED}\n")` to account for that here.

scarb/tests/fmt.rs
maciektr: I would prefer if we followed what Cargo does in this regard (i.e. `cargo fmt -- --emit stdout`), as Scarb is mostly inspired by Cargo's approach. This means printing first full path to file, terminated with `:`, then from the next line file contents. 
I have not looked into tools that may use this format - please do! 🙇

maciektr (commented):
Great work! 🦾 🦾 🦾 

We need to decide on output format for multiple files before moving forward though.

scarb/tests/fmt.rs
Eikix: Thanks! Fixed

scarb/tests/fmt.rs
Eikix: Great idea! I did that

scarb/tests/fmt.rs
```diff
@@ -58,6 +58,40 @@ fn simple_check_invalid() {
...
+        .arg("fmt")
+        .arg("--emit")
+        .arg("stdout")
+        .arg("--no-color")
+        .current_dir(&t)
+        .assert()
+        .failure()
+        .stdout_matches(format!(
+            "{}:\n{}\n",
+            fs::canonicalize(t.child("src/lib.cairo"))
```
maciektr, line 74:
```suggestion
            fsx::canonicalize(t.child("src/lib.cairo"))
```
Please use `fsx` instead, it's our internal extension of `fs`.

scarb/tests/fmt.rs
```diff
@@ -270,3 +304,79 @@ fn workspace_with_root() {
...
+    Scarb::quick_snapbox()
+        .arg("fmt")
+        .arg("--emit")
+        .arg("stdout")
+        .current_dir(&t)
+        .assert()
+        .failure()
+        .stdout_matches(format!(
+            "{}:\n{}\n",
+            fs::canonicalize(t.child("src/lib.cairo"))
```
maciektr, line 342:
```suggestion
            fsx::canonicalize(t.child("src/lib.cairo"))
```
Please use `fsx` instead, it's our internal extension of `fs`.

scarb/tests/fmt.rs
```diff
@@ -270,3 +304,79 @@ fn workspace_with_root() {
...
+            "{}:\n{}\n{}:\n{}\n{}:\n{}\n",
+            fs::canonicalize(t.child("first/src/lib.cairo"))
+                .unwrap()
+                .display(),
+            SIMPLE_FORMATTED,
+            fs::canonicalize(t.child("second/src/lib.cairo"))
+                .unwrap()
+                .display(),
+            SIMPLE_FORMATTED,
+            fs::canonicalize(t.child("src/lib.cairo"))
```
maciektr, line 370:
```suggestion
            fsx::canonicalize(t.child("first/src/lib.cairo"))
                .unwrap()
                .display(),
            SIMPLE_FORMATTED,
            fsx::canonicalize(t.child("second/src/lib.cairo"))
                .unwrap()
                .display(),
            SIMPLE_FORMATTED,
            fsx::canonicalize(t.child("src/lib.cairo"))
```
Please use `fsx` instead, it's our internal extension of `fs`.

maciektr (commented):
This is incredible! Great job! ❤️ 
The last tiny bit before we can merge it is to change `fs::canonicalize` to `fsx::canonicalize`, as per linter suggestion.

scarb/tests/fmt.rs
Eikix: Nice! Thx:)

Eikix:
> ❤️ ❤️ ❤️ Thank you very much!

Thank you 🫶🙏👑
image

Issue #883, opened 2023-11-08
Publish tests produce different checksums on windows


maciektr:

mkaput:
This is caused by https://github.com/alexcrichton/tar-rs/issues/341. We'll work around this on our side, and I'll submit a patch upstream.

PR #884, opened 2023-11-08
Ignore tests failing as of #883

maciektr:
commit-id:1bcbe058

### DIFF
--- a/scarb/tests/local_registry.rs
+++ b/scarb/tests/local_registry.rs
@@ -123,7 +123,9 @@ fn url_pointing_to_file() {
     drop(registry_t);
 }
 
+// FIXME(#883): Unignore these tests.
 #[test]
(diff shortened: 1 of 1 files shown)


maciektr:
Only those two are failing: https://github.com/software-mansion/scarb/actions/runs/6798423689/job/18482546239

scarb/tests/local_registry.rs
```diff
@@ -123,7 +123,9 @@ fn url_pointing_to_file() {
     drop(registry_t);
 }
 
+// FIXME(#883)
```
mkaput, line 126:
```suggestion
// FIXME(#883): Unignore these tests.
```

otherwise editors may not catch this

mkaput (approved):

scarb/tests/local_registry.rs
maciektr: Thank you!

PR #886, opened 2023-11-08
Add snforge test collector

piotmag769:
Tests will be added to ensure bumping cairo versions doesn't break anything. Before introducing it in snforge will be tested on snforge side locally at first, then with the nightly build.

Example usage:
```console
scarb --manifest-path simple_package/Scarb.toml test-collector --output-path simple_package/target/snforge/simple_package_tests.sierra.json
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "extensions/scarb-cairo-language-server",
     "extensions/scarb-cairo-run",
     "extensions/scarb-cairo-test",
+    "extensions/scarb-snforge-test-collector",
     "utils/create-output-dir",
     "utils/scarb-build-metadata",
     "utils/scarb-test-support",
@@ -29,19 +30,27 @@ anyhow = "1"
 assert_fs = "1"
 async-trait = "0.1"
 axum = { version = "0.6", features = ["http2"] }
-cairo-lang-compiler = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-defs = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-filesystem = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-formatter = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-language-server = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-runner = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-semantic = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-sierra = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-sierra-to-casm = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-starknet = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-test-plugin = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-test-runner = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
-cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo", rev = "2850a160c348b72d020d1464544323c0a772c6d4", features = ["env_logger"] }
+cairo-felt = "0.9"
+cairo-lang-casm = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-compiler = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-diagnostics = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-debug = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-defs = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-filesystem = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-formatter = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-language-server = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-lowering = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-project = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-runner = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-semantic = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-sierra = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-sierra-generator = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-sierra-to-casm = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-starknet = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-syntax = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-test-plugin = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-test-runner = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4" }
+cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo.git", rev = "2850a160c348b72d020d1464544323c0a772c6d4", features = ["env_logger"] }
 camino = { version = "1", features = ["serde1"] }
 cargo_metadata = ">=0.18"
 clap = { version = "4", features = ["derive", "env", "string"] }
@@ -69,6 +78,8 @@ itertools = "0.11"
 libc = "0.2"
 log = "0.4"
 ntest = "0.9"
+num-bigint = { version = "0.4", features = ["rand"] }
+num-traits = "0.2"
 once_cell = "1"
 pathdiff = { version = "0.2", features = ["camino"] }
 petgraph = "0.6"
@@ -76,6 +87,7 @@ predicates = "3"
 proc-macro2 = "1"
 quote = "1"
 redb = "1.3"
+rayon = "1.8"
 reqwest = { version = "0.11", features = ["gzip", "brotli", "deflate", "json", "stream"], default-features = false }
 semver = { version = "1", features = ["serde"] }
 serde = { version = "1", features = ["serde_derive"] }
@@ -89,6 +101,7 @@ similar-asserts = { version = "1", features = ["serde"] }
 smallvec = "1"
 smol_str = { version = "0.2", features = ["serde"] }
 snapbox = { version = "0.4", features = ["cmd", "path"] }
+starknet = "0.6"
 syn = "2"
 tar = "0.4"
 tempfile = "3"
--- a/extensions/scarb-snforge-test-collector/src/compilation.rs
+++ b/extensions/scarb-snforge-test-collector/src/compilation.rs
@@ -1,0 +1,44 @@
+use anyhow::Result;
+use cairo_lang_sierra::program::{ProgramArtifact, VersionedProgram};
+use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
+use serde::Serialize;
+
+use crate::compilation::test_collector::{collect_tests, TestCaseRaw};
+use crate::crate_collection::{CrateLocation, TestCompilationTarget};
+use crate::metadata::CompilationUnit;
+
+pub mod test_collector;
+
+#[derive(Debug, Clone, Serialize, PartialEq)]
+pub struct CompiledTestCrateRaw {
+    pub sierra_program: VersionedProgram,
+    pub test_cases: Vec<TestCaseRaw>,
+    pub tests_location: CrateLocation,
+}
+
+pub fn compile_tests(
+    targets: &Vec<TestCompilationTarget>,
+    compilation_unit: &CompilationUnit,
+) -> Result<Vec<CompiledTestCrateRaw>> {
+    targets
+        .par_iter()
+        .map(|target| target.compile_tests(compilation_unit))
+        .collect()
+}
+
+impl TestCompilationTarget {
+    fn compile_tests(&self, compilation_unit: &CompilationUnit) -> Result<CompiledTestCrateRaw> {
+        let (sierra_program, test_cases) = collect_tests(
+            &self.crate_name,
+            self.crate_root.as_std_path(),
+            &self.lib_content,
+            compilation_unit,
+        )?;
+
+        Ok(CompiledTestCrateRaw {
+            sierra_program: VersionedProgram::v1(ProgramArtifact::stripped(sierra_program)),
+            test_cases,
+            tests_location: self.crate_location.clone(),
+        })
+    }
+}
--- a/extensions/scarb-snforge-test-collector/src/compilation/test_collector.rs
+++ b/extensions/scarb-snforge-test-collector/src/compilation/test_collector.rs
@@ -1,0 +1,241 @@
+use anyhow::{anyhow, bail, Context, Result};
+use cairo_lang_compiler::db::RootDatabase;
+use cairo_lang_compiler::diagnostics::DiagnosticsReporter;
+use cairo_lang_debug::DebugWithDb;
+use cairo_lang_defs::db::DefsGroup;
+use cairo_lang_defs::ids::{FreeFunctionId, FunctionWithBodyId, ModuleId, ModuleItemId};
+use cairo_lang_diagnostics::ToOption;
+use cairo_lang_filesystem::cfg::{Cfg, CfgSet};
+use cairo_lang_filesystem::db::{AsFilesGroupMut, CrateConfiguration, FilesGroup, FilesGroupEx};
+use cairo_lang_filesystem::ids::{CrateId, CrateLongId, Directory};
+use cairo_lang_lowering::ids::ConcreteFunctionWithBodyId;
+use cairo_lang_project::{ProjectConfig, ProjectConfigContent};
+use cairo_lang_semantic::db::SemanticGroup;
+use cairo_lang_semantic::items::functions::GenericFunctionId;
+use cairo_lang_semantic::{ConcreteFunction, FunctionLongId};
+use cairo_lang_sierra::extensions::enm::EnumType;
+use cairo_lang_sierra::extensions::NamedType;
+use cairo_lang_sierra::program::{GenericArg, Program};
+use cairo_lang_sierra_generator::db::SierraGenGroup;
+use cairo_lang_sierra_generator::replace_ids::replace_sierra_ids_in_program;
+use cairo_lang_starknet::starknet_plugin_suite;
+use cairo_lang_test_plugin::test_plugin_suite;
+use cairo_lang_utils::ordered_hash_map::OrderedHashMap;
+use itertools::Itertools;
+use serde::Serialize;
+use smol_str::SmolStr;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+use crate::compilation::test_collector::config::ExpectedTestResult;
+use crate::metadata::CompilationUnit;
+pub use config::RawForkConfig;
+use config::{forge_try_extract_test_config, FuzzerConfig, SingleTestConfig};
+use plugin::snforge_test_plugin_suite;
+use sierra_casm_generator::FunctionFinder;
+
+mod config;
+mod plugin;
+mod sierra_casm_generator;
+
+fn find_all_tests(
+    db: &dyn SemanticGroup,
+    main_crate: CrateId,
+) -> Vec<(FreeFunctionId, SingleTestConfig)> {
+    let mut tests = vec![];
+    let modules = db.crate_modules(main_crate);
+    for module_id in modules.iter() {
+        let Ok(module_items) = db.module_items(*module_id) else {
+            continue;
+        };
+        tests.extend(module_items.iter().filter_map(|item| {
+            let ModuleItemId::FreeFunction(func_id) = item else {
+                return None;
+            };
+            let Ok(attrs) = db.function_with_body_attributes(FunctionWithBodyId::Free(*func_id))
+            else {
+                return None;
+            };
+            Some((
+                *func_id,
+                forge_try_extract_test_config(db.upcast(), &attrs).unwrap()?,
+            ))
+        }));
+    }
+    tests
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize)]
+pub struct TestCaseRaw {
+    pub name: String,
+    pub available_gas: Option<usize>,
+    pub ignored: bool,
+    pub expected_result: ExpectedTestResult,
+    pub fork_config: Option<RawForkConfig>,
+    pub fuzzer_config: Option<FuzzerConfig>,
+}
+
+pub fn collect_tests(
+    crate_name: &str,
+    crate_root: &Path,
+    lib_content: &str,
+    compilation_unit: &CompilationUnit,
+) -> Result<(Program, Vec<TestCaseRaw>)> {
+    let crate_roots: OrderedHashMap<SmolStr, PathBuf> = compilation_unit
+        .dependencies()
+        .iter()
+        .cloned()
+        .map(|source_root| (source_root.name.into(), source_root.path))
+        .collect();
+
+    let project_config = ProjectConfig {
+        base_path: crate_root.into(),
+        corelib: Some(Directory::Real(compilation_unit.corelib_path()?)),
+        content: ProjectConfigContent {
+            crate_roots,
+            crates_config: compilation_unit.crates_config_for_compilation_unit(),
+        },
+    };
+
+    // code taken from crates/cairo-lang-test-runner/src/lib.rs
+    let db = &mut {
+        let mut b = RootDatabase::builder();
+        b.with_cfg(CfgSet::from_iter([Cfg::name("test")]));
+        b.with_plugin_suite(snforge_test_plugin_suite());
+        b.with_plugin_suite(test_plugin_suite());
+        b.with_plugin_suite(starknet_plugin_suite());
+        b.with_project_config(project_config);
+        b.build()?
+    };
+
+    let main_crate_id =
+        insert_lib_entrypoint_content_into_db(db, crate_name, crate_root, lib_content);
+
+    if DiagnosticsReporter::stderr().check(db) {
+        return Err(anyhow!(
+            "Failed to compile test artifact, for detailed information go through the logs above"
+        ));
+    }
+    let all_tests = find_all_tests(db, main_crate_id);
+
+    let z: Vec<ConcreteFunctionWithBodyId> = all_tests
+        .iter()
+        .filter_map(|(func_id, _cfg)| {
+            ConcreteFunctionWithBodyId::from_no_generics_free(db, *func_id)
+        })
+        .collect();
+
+    let sierra_program = db
+        .get_sierra_program_for_functions(z)
+        .to_option()
+        .context("Compilation failed without any diagnostics")
+        .context("Failed to get sierra program")?;
+
+    let collected_tests = all_tests
+        .into_iter()
+        .map(|(func_id, test)| {
+            (
+                format!(
+                    "{:?}",
+                    FunctionLongId {
+                        function: ConcreteFunction {
+                            generic_function: GenericFunctionId::Free(func_id),
+                            generic_args: vec![]
+                        }
+                    }
+                    .debug(db)
+                ),
+                test,
+            )
+        })
+        .collect_vec()
+        .into_iter()
+        .map(|(test_name, config)| TestCaseRaw {
+            name: test_name,
+            available_gas: config.available_gas,
+            ignored: config.ignored,
+            expected_result: config.expected_result,
+            fork_config: config.fork_config,
+            fuzzer_config: config.fuzzer_config,
+        })
+        .collect();
+
+    let sierra_program = replace_sierra_ids_in_program(db, &sierra_program);
+
+    validate_tests(sierra_program.clone(), &collected_tests)?;
+
+    Ok((sierra_program, collected_tests))
+}
+
+// inspired with cairo-lang-compiler/src/project.rs:49 (part of setup_single_project_file)
+fn insert_lib_entrypoint_content_into_db(
+    db: &mut RootDatabase,
+    crate_name: &str,
+    crate_root: &Path,
+    lib_content: &str,
+) -> CrateId {
+    let main_crate_id = db.intern_crate(CrateLongId::Real(SmolStr::from(crate_name)));
+    db.set_crate_config(
+        main_crate_id,
+        Some(CrateConfiguration::default_for_root(Directory::Real(
+            crate_root.to_path_buf(),
+        ))),
+    );
+
+    let module_id = ModuleId::CrateRoot(main_crate_id);
+    let file_id = db.module_main_file(module_id).unwrap();
+    db.as_files_group_mut()
+        .override_file_content(file_id, Some(Arc::new(lib_content.to_string())));
+
+    main_crate_id
+}
+
+fn validate_tests(
+    sierra_program: Program,
+    collected_tests: &Vec<TestCaseRaw>,
+) -> Result<(), anyhow::Error> {
+    let function_finder = match FunctionFinder::new(sierra_program) {
+        Ok(casm_generator) => casm_generator,
+        Err(e) => panic!("{}", e),
+    };
+    for test in collected_tests {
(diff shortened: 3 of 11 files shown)


Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
     "extensions/scarb-cairo-language-server",
     "extensions/scarb-cairo-run",
     "extensions/scarb-cairo-test",
+    "extensions/scarb-test-collector",
```
mkaput, line 8:
```suggestion
    "extensions/scarb-snforge-test-collector",
```

`test-collector` by itself was pretty ambiguous IMO

extensions/scarb-test-collector/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+once_cell.workspace = true
+rayon.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+smol_str.workspace = true
+starknet.workspace = true
+thiserror.workspace = true
+walkdir.workspace = true
+
+scarb-metadata = { path = "../../scarb-metadata" }
```
mkaput, line 48:
```suggestion
rayon.workspace = true
scarb-metadata = { path = "../../scarb-metadata" }
serde.workspace = true
serde_json.workspace = true
smol_str.workspace = true
starknet.workspace = true
thiserror.workspace = true
walkdir.workspace = true
```

Cargo.toml
```diff
@@ -68,12 +81,16 @@ itertools = "0.11"
 libc = "0.2"
 log = "0.4"
 ntest = "0.9"
+num-bigint = { version = "0.4.4", features = ["rand"] }
+num-integer = "0.1"
+num-traits = "0.2.17"
```
mkaput, line 86:
```suggestion
num-traits = "0.2"
```

Cargo.toml
```diff
@@ -68,12 +81,16 @@ itertools = "0.11"
 libc = "0.2"
 log = "0.4"
 ntest = "0.9"
+num-bigint = { version = "0.4.4", features = ["rand"] }
```
mkaput, line 84:
```suggestion
num-bigint = { version = "0.4", features = ["rand"] }
```

Cargo.toml
```diff
@@ -29,19 +30,31 @@ anyhow = "1"
...
-cairo-lang-runner = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-semantic = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-sierra = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-sierra-to-casm = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-starknet = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-test-plugin = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-test-runner = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f", features = ["env_logger"] }
+cairo-felt = "0.8.2"
```
mkaput, line 33:
```suggestion
cairo-felt = "0.8"
```

Cargo.toml
```diff
@@ -29,19 +30,31 @@ anyhow = "1"
...
+cairo-lang-sierra-ap-change = { git = "https://github.com/starkware-libs/cairo", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
+cairo-lang-sierra-gas = { git = "https://github.com/starkware-libs/cairo", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
+cairo-lang-sierra-generator = { git = "https://github.com/starkware-libs/cairo", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
+cairo-lang-sierra-to-casm = { git = "https://github.com/starkware-libs/cairo", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
+cairo-lang-starknet = { git = "https://github.com/starkware-libs/cairo", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
+cairo-lang-syntax = { git = "https://github.com/starkware-libs/cairo", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
+cairo-lang-test-plugin = { git = "https://github.com/starkware-libs/cairo", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
+cairo-lang-test-runner = { git = "https://github.com/starkware-libs/cairo", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
+cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f", features = ["env_logger"] }
+cairo-vm = "0.8.2"
```
mkaput, line 57:
```suggestion
cairo-vm = "0.8"
```

Cargo.toml
```diff
@@ -87,6 +104,7 @@ similar-asserts = { version = "1", features = ["serde"] }
 smallvec = "1"
 smol_str = { version = "0.2", features = ["serde"] }
 snapbox = { version = "0.4", features = ["cmd", "path"] }
+starknet = "0.6.0"
```
mkaput, line 107:
```suggestion
starknet = "0.6"
```

mkaput (changes requested):

Cargo.toml
mkaput, line 1: I'm not sure that we need all of these dependencies.  I see some `#[allow(dead_code)]` lines. Could you please remove that dead code and thus remove unnecessary dependencies?

Scarb is already super slow to compile, I don't like adding more big stuff for this reason.

mkaput (changes requested):

extensions/scarb-test-collector/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+
+mod compilation;
+mod crate_collection;
+mod metadata;
+
+#[derive(Parser)]
+#[command(version)]
+struct Args {
+    #[arg(long)]
+    output_path: String,
```
mkaput, line 22:
```suggestion
    output_path: Utf8PathBuf,
```

extensions/scarb-test-collector/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+use crate::metadata::{
+    corelib_for_package, dependencies_for_package, id_for_package, paths_for_package,
+};
+
+mod compilation;
+mod crate_collection;
+mod metadata;
+
+#[derive(Parser)]
+#[command(version)]
```
mkaput, line 19: Add a help line saying that this is Starknet Foundry private extension and user should not call it manually.

extensions/scarb-test-collector/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+struct Args {
+    #[arg(long)]
+    output_path: String,
+}
+
+fn main() -> Result<()> {
+    let args = Args::parse();
+    let scarb_metadata = MetadataCommand::new().inherit_stderr().exec()?;
+
+    let manifest_path = Utf8PathBuf::from_str(&std::env::var("SCARB_MANIFEST_PATH")?)?;
```
mkaput, line 29: Wouldn't it make more sense for you  to use `SCARB_PACKAGE` env, handled by scarb-ui's PackagesFilter?

mkaput (changes requested):

extensions/scarb-test-collector/src/main.rs
piotmag769: Is it passable to scarb when calling subcommand? Or should I just call pass it directly?

Cargo.toml
maciektr: Adding on this, maybe we could skip deserializing those values into starknet-rs types here, and do this in starknet-foundry instead? 🤔 This way we could possibly drop starknet-rs. 
@mkaput @piotmag769 
https://github.com/software-mansion/scarb/blob/37eeb1a515deba0b773661af7ad6989847dd034b/extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs#L249

piotmag769: Good idea, will do. We had it this way since it was convenient before

extensions/scarb-snforge-test-collector/src/compilation.rs
```diff
@@ -0,0 +1,68 @@
...
+
+pub(crate) type CompiledTestCrateRaw = CompiledTestCrate<RawForkConfig>;
+
+pub fn compile_tests(
+    targets: &Vec<TestCompilationTarget>,
+    linked_libraries: &[LinkedLibrary],
+    corelib_path: &Utf8Path,
+) -> Result<Vec<CompiledTestCrateRaw>> {
+    targets
+        .par_iter()
```
cptartur, line 42: Just to make sure, we are okay with using Rayon in scarb?

extensions/scarb-snforge-test-collector/src/crate_collection.rs
```diff
@@ -0,0 +1,82 @@
...
+    Lib,
+    /// Crate in the `tests/` directory
+    Tests,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct TestCompilationTarget {
+    pub crate_root: Utf8PathBuf,
+    pub crate_name: String,
+    pub crate_location: CrateLocation,
```
cptartur, line 18: Is this necessary in the collector at all? We use this information in forge, but it doesn't seem like it is needed here.

extensions/scarb-snforge-test-collector/src/crate_collection.rs
```diff
@@ -0,0 +1,82 @@
...
+        compilation_targets.push(TestCompilationTarget {
+            crate_name: "tests".to_string(),
+            crate_location: CrateLocation::Tests,
+            lib_content: get_or_create_test_lib_content(tests_dir_path.as_path())?,
+            crate_root: tests_dir_path,
+        });
+    }
+
+    Ok(compilation_targets)
+}
```
cptartur, line 45: Are we sure this should be handled by scarb? I understand moving the compilation there, but shouldn't forge tell it what we want to compile? This way we have control on this targets.

cc @mkaput @maciektr

extensions/scarb-snforge-test-collector/src/metadata.rs
cptartur, line 1: Should we still need to read this values from metadata? If collection is part of scarb, wouldn't it be possible to resolve this informations in a more direct way? Or is this the preferred way?

Cargo.toml
```diff
@@ -29,19 +30,27 @@ anyhow = "1"
...
-cairo-lang-semantic = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-sierra = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-sierra-to-casm = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-starknet = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-test-plugin = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-test-runner = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
-cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo.git", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f", features = ["env_logger"] }
+cairo-felt = "0.8"
+cairo-lang-casm = { git = "https://github.com/starkware-libs/cairo", rev = "dd8bd1f43eff37ba747dcd92e5c821bdfa8c007f" }
```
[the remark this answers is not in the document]
maciektr: @cptartur You will need to use the same revision of the compiler as we do. We often have to pin to the main for development, then release with version from crates.

extensions/scarb-snforge-test-collector/src/compilation.rs
```diff
@@ -0,0 +1,68 @@
...
+use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
+use serde::Serialize;
+
+use crate::compilation::test_collector::{collect_tests, ForkConfig, RawForkConfig, TestCase};
+use crate::crate_collection::{CrateLocation, TestCompilationTarget};
+use crate::metadata::LinkedLibrary;
+
+pub mod test_collector;
+
+static BUILTINS: Lazy<Vec<&str>> = Lazy::new(|| {
```
maciektr, line 14: Is there a reason why it is not public in compiler api?

piotmag769: No idea tbh, can make a contribution to the compiler tomorrow

https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-starknet/src/plugin/consts.rs#L53C51-L53C51

extensions/scarb-snforge-test-collector/src/metadata.rs
piotmag769: See the comment above, I think we want to do it this way - `snforge-test-collector` as a sort of separate entity that handles generating test artifacts by itself

extensions/scarb-snforge-test-collector/src/crate_collection.rs
piotmag769: See the comments above and below. If there is a need we can set up a call to discuss it

piotmag769: See the comment above and below. If there is a need we can set up a call to discuss it

extensions/scarb-snforge-test-collector/src/compilation.rs
piotmag769: @mkaput @maciektr actually I don't think it should be here anymore, it is a leftover from times where fuzzing isn't a thing. Would still like to merge the PR to cairo repo since we use it in forge for fuzzing. Sorry for the confusion

cc @cptartur

extensions/scarb-snforge-test-collector/src/compilation/test_collector.rs
```diff
@@ -0,0 +1,253 @@
...
+    };
+    for test in collected_tests {
+        let func = casm_generator.find_function(&test.name)?;
+        let mut filtered_params: Vec<String> = Vec::new();
+        for param in &func.params {
+            let param_str = &param.ty.debug_name.as_ref().unwrap().to_string();
+            if !ignored_params.contains(&param_str.as_str()) {
+                filtered_params.push(param_str.to_string());
+            }
+        }
```
piotmag769, line 213: Removing it, the filtered params were not used anywhere 🤷 
cc @MaksymilianDemitraszek @cptartur

extensions/scarb-snforge-test-collector/src/crate_collection.rs
cptartur: I mean `crate_location` specifically. This is only useful in forge and only for pretty printing.

extensions/scarb-snforge-test-collector/src/compilation.rs
cptartur: These are also used to filter builtin arguments from user defined ones in fuzzing. We'll still need it in forge.

extensions/scarb-snforge-test-collector/src/crate_collection.rs
piotmag769: I don't expect us to make breaking changes in test collection. Otherwise they will be breaking anyways, so I think it's fine to not be backwards compatible then

extensions/scarb-snforge-test-collector/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+fn main() -> Result<()> {
+    let args = Args::parse();
+    let scarb_metadata = MetadataCommand::new().inherit_stderr().exec()?;
+
+    let manifest_path = Utf8PathBuf::from_str(&std::env::var("SCARB_MANIFEST_PATH")?)?;
+    let package_id = id_for_package(&scarb_metadata, &manifest_path)?;
+
+    let (package_path, package_source_dir_path, package_name) =
```
maciektr, line 34: I feel like this flattening of project abstraction is actually more confusing?
You could pass `PackageMetadata` (extracted with packages_filter) downwards, and extract info you need, where you actually need it.

extensions/scarb-snforge-test-collector/src/crate_collection.rs
```diff
@@ -0,0 +1,82 @@
...
+pub fn collect_test_compilation_targets(
+    package_path: &Utf8Path,
+    package_name: &str,
+    package_source_dir_path: &Utf8Path,
+) -> Result<Vec<TestCompilationTarget>> {
+    let mut compilation_targets = vec![TestCompilationTarget {
+        crate_root: package_source_dir_path.to_path_buf(),
+        crate_name: package_name.to_string(),
+        crate_location: CrateLocation::Lib,
+        lib_content: std::fs::read_to_string(package_source_dir_path.join("lib.cairo"))?,
```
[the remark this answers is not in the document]
maciektr: Also, please add context with file path to read_to_string result, e.g. ```.with_context(|| format!("failed to read `{}`", path.display()))```

extensions/scarb-snforge-test-collector/src/crate_collection.rs
```diff
@@ -0,0 +1,82 @@
...
+        });
+    }
+
+    Ok(compilation_targets)
+}
+
+fn get_or_create_test_lib_content(tests_folder_path: &Utf8Path) -> Result<String> {
+    let tests_lib_path = tests_folder_path.join("lib.cairo");
+    if tests_lib_path.try_exists()? {
+        return Ok(std::fs::read_to_string(tests_lib_path)?);
```
maciektr, line 50: vide supra

extensions/scarb-snforge-test-collector/src/crate_collection.rs
```diff
@@ -0,0 +1,82 @@
...
+
+fn module_name_from_path<'a>(tests_folder_path: &Utf8Path, path: &'a Utf8Path) -> &'a str {
+    path.strip_prefix(tests_folder_path)
+        .unwrap_or_else(|_| {
+            panic!(
+                "Path to test = {path} does not start with test_folder_path = {tests_folder_path}"
+            )
+        })
+        .as_str()
+        .strip_suffix(".cairo")
```
maciektr, line 80: Isn't this roughly equivalent to `path.file_stem()`?

extensions/scarb-snforge-test-collector/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+        paths_for_package(&scarb_metadata, &package_id)?;
+    let dependencies = dependencies_for_package(&scarb_metadata, &package_id)?;
+    let corelib_path = corelib_for_package(&scarb_metadata, &package_id)?;
+
+    let compilation_targets =
+        collect_test_compilation_targets(&package_path, &package_name, &package_source_dir_path)?;
+    let test_crates = compile_tests(&compilation_targets, &dependencies, &corelib_path)?;
+
+    let serialized_crates = serde_json::to_string(&test_crates)?;
+    std::fs::write(args.output_path, serialized_crates)?;
```
maciektr, line 44:
```suggestion
    let file = std::fs::File::create(args.output_path.clone())?;
    let file = BufWriter::new(&file);
    serde_json::to_writer(file, &test_crates)
        .context("failed to serialize test crates")?;
```

maciektr: Will foundry make sure that the containing directory of output_path exists? Both `std::fs` functions declare `Depending on the platform, this function may fail if the full directory path does not exist`

maciektr: Shouldn't we lock here? cc'ing @mkaput as our lock expert 👑

extensions/scarb-snforge-test-collector/src/compilation/test_collector.rs
```diff
@@ -0,0 +1,244 @@
...
+        .cloned()
+        .map(|source_root| (source_root.name.into(), source_root.path))
+        .collect();
+
+    let project_config = ProjectConfig {
+        base_path: crate_root.into(),
+        corelib: Some(Directory::Real(corelib_path)),
+        content: ProjectConfigContent {
+            crate_roots,
+            crates_config: Default::default(),
```
maciektr, line 99: Crate editions should be read from scarb metadata (available since 1.9.0 https://docs.rs/scarb-metadata/latest/scarb_metadata/struct.PackageMetadata.html#structfield.edition) and passed here as in https://github.com/software-mansion/scarb/blob/fcc3307b4b3f59b7abb16e07cf4048dea780e371/scarb/src/compiler/db.rs#L94

maciektr (commented):
🦾 🦾 🦾 
Some pain in the neck comments below :D

extensions/scarb-snforge-test-collector/src/main.rs
piotmag769: Yup, it will be created by `snforge`. I thought about `workspace_root/snforge_target/{package_name}.sierra.json` as the output path. I chose a separate directory to make sure Scarb doesn't overwrite anything in `target` directory

extensions/scarb-snforge-test-collector/src/crate_collection.rs
piotmag769: We can enforce the path here as it is `tests/lib.cairo`, it has nothing to do with package source path. It is an entrypoint to our test target

maciektr: Oh yes, but please add context as FS::read error will not be easy to pin to specific file.

extensions/scarb-snforge-test-collector/src/main.rs
mkaput: Having a custom `snforge_target` directory, you basically introduce a new concept that users will have to put in their `gitignore` and also it will hurt their eyes IMHO.

I'd definitely keep this in the `target/<profile name>` hierarchy. You can freely use the `target/<profile name>/snforge` directory.

That said, it'd be safer to use our file locking machinery here, for safety purposes. Who knows, when some user will attempt to run `snforge build` twice 😅 For this, you have to use our [`Filesystem`](https://github.com/software-mansion/scarb/blob/b78262ec74991546bc8c3d2383abb2b6f57f5260/scarb/src/flock.rs#L130) data structure and its [`open_rw`](https://github.com/software-mansion/scarb/blob/b78262ec74991546bc8c3d2383abb2b6f57f5260/scarb/src/flock.rs#L187) method. Here you have a snippet from our compilers that writes JSON to such files:

https://github.com/software-mansion/scarb/blob/a5d8783c9936db66be138704ffdcf7e01457ff02/scarb/src/compiler/helpers.rs#L46-L58

An extra benefit is that `scarb clean` will handle it for free for you.

extensions/scarb-snforge-test-collector/src/compilation.rs
mkaput: yes, that's ok

mkaput (commented):
Looks OK for me, let's resolve all comments and I think this should be good to go 🚀 

I hope we'll manage to get this released in 2.4.0 🤞🏻

extensions/scarb-snforge-test-collector/src/metadata.rs
```diff
@@ -0,0 +1,95 @@
...
+
+pub fn compilation_unit_for_package<'a>(
+    metadata: &'a Metadata,
+    package_metadata: &PackageMetadata,
+) -> Result<&'a CompilationUnitMetadata> {
+    metadata
+        .compilation_units
+        .iter()
+        .filter(|unit| unit.package == package_metadata.id)
+        .min_by_key(|unit| match unit.target.kind.as_str() {
```
piotmag769, line 84: Just to make sure - we should use `kind` here? `name` was used before 🤔

Cargo.toml
piotmag769: Is it urgent? Or can we do it in the next PR? To make it anyhow readable for the forge team I would have to first move this parsing a layer up on our side; that means additional PR, review etc.

extensions/scarb-snforge-test-collector/src/metadata.rs
maciektr: Yes, makes sense to use kind.

Cargo.toml
```diff
@@ -113,7 +126,7 @@ zip = { version = "0.6", default-features = false, features = ["deflate"] }
 zstd = "0.13"
 
 [profile.release]
-lto = true
+#lto = true
```
mkaput, line 129:
```suggestion
lto = true
```

Cargo.toml
mkaput: We can merge this here, but please submit next PR asap.

piotmag769: https://github.com/foundry-rs/starknet-foundry/issues/1119

extensions/scarb-snforge-test-collector/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    let metadata = MetadataCommand::new().inherit_stderr().exec()?;
+    let selected_packages_metadata = args.packages_filter.match_many(&metadata)?;
+
+    let target_dir = metadata
+        .target_dir
+        .clone()
+        .unwrap_or_else(|| metadata.workspace.root.join("target"));
+    let snforge_target_dir = target_dir.join(&metadata.current_profile).join("snforge");
+    create_dir_all(&snforge_target_dir)?;
```
maciektr, line 41:
```suggestion
   create_output_dir::create_output_dir(   path_to_target   )?;
    create_dir_all(&snforge_target_dir)?;
```
Please ensure that target dir is created with https://github.com/software-mansion/scarb/tree/main/utils/create-output-dir first.

extensions/scarb-snforge-test-collector/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+            &dependencies,
+            &corelib_path,
+            &crates_config,
+        )?;
+
+        let serialized_crates = serde_json::to_string(&test_crates)?;
+
+        // artifact saved to `{target_dir}/{profile_name}/{package_name}.sierra.json`
+        let output_path =
+            snforge_target_dir.join(&format!("{}.sierra.json", package_metadata.name));
```
maciektr, line 69: Maybe it would make sense to use a different extensions, as this format differs from `.sierra.json` output of lib compilation?

extensions/scarb-snforge-test-collector/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+            &package_metadata.name,
+            &package_metadata.root,
+            compilation_unit_metadata.target.source_root(),
+            &compilation_unit_metadata.target.source_path,
+        )?;
+
+        let dependencies = dependencies_for_compilation_unit(compilation_unit_metadata);
+        let corelib_path = corelib_for_compilation_unit(compilation_unit_metadata)?;
+        let crates_config =
+            crates_config_for_compilation_unit(&metadata, compilation_unit_metadata);
```
maciektr, line 56: I still don't think this destructuring of `CompilationUnit` is beneficial in any way. I think you would get more flexible api if for instance you wrap `CompilationUnit` into some impl and extract raw paths where needed only, instead of having this series of function calls with long signatures. 

E.g. 
```
#[derive(Clone)]
pub struct CompilationUnit<'a>(&'a CompilationUnitMetadata);

impl CompilationUnit<'_> {
    pub fn dependencies(
        &self,
    ) -> Vec<LinkedLibrary> {
        let dependencies = self.0
            .components
            .iter()
            .filter(|du| &du.name != "core")
            .map(|cu| LinkedLibrary {
                name: cu.name.clone(),
                path: cu.source_root().to_owned().into_std_path_buf(),
            })
            .collect();

        dependencies
    }

    pub fn corelib_path(&self) -> Result<Utf8PathBuf> {
        let corelib = self.0
            .components
            .iter()
            .find(|du| du.name == "core")
            .context("corelib could not be found")?;
        Ok(Utf8PathBuf::from(corelib.source_root()))
    }
}
```

```
pub fn collect_tests(
    crate_name: &str,
    crate_root: &Path,
    lib_content: &str,
    crates_config: AllCratesConfig,
   compilation_unit: CompilationUnit
) -> Result<(Program, Vec<TestCaseRaw>)> {
    let crate_roots: OrderedHashMap<SmolStr, PathBuf> = compilation_unit.dependencies()
        .iter()
        .cloned()
        .map(|source_root| (source_root.name.into(), source_root.path))
        .collect();

    let project_config = ProjectConfig {
        base_path: crate_root.into(),
        corelib: Some(Directory::Real(compilation_unit.corelib_path()?.into())),
```

extensions/scarb-snforge-test-collector/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+            .write(true)
+            .truncate(true)
+            .open(&output_path)?;
+
+        output_file
+            .lock_exclusive()
+            .with_context(|| format!("Couldn't lock the output file = {output_path}"))?;
+        output_file
+            .write_all(serialized_crates.as_ref())
+            .with_context(|| format!("Couldn't lock the output file = {output_path}"))?;
```
maciektr, line 81:
```suggestion
    let file = BufWriter::new(&* output_file);
    serde_json::to_writer(file, & test_crates)
        .with_context(|| format!("failed to serialize {file_name}"))?;
```

extensions/scarb-snforge-test-collector/src/main.rs
piotmag769: `snforge_sierra.json`?

piotmag769:
Any approves? 🙏  @mkaput (will add tests in the next PR, we have to think about how to approach it)

MaksymilianDemitraszek:
It would be very useful in case of such PR if you marked which code is not copied & requires review

extensions/scarb-snforge-test-collector/src/compilation/test_collector/sierra_casm_generator.rs
```diff
@@ -0,0 +1,62 @@
...
+    #[error(transparent)]
+    ProgramRegistryError(#[from] Box<ProgramRegistryError>),
+}
+
+pub struct SierraCasmGenerator {
+    /// The sierra program.
+    sierra_program: cairo_lang_sierra::program::Program,
+    /// Program registry for the Sierra program.
+    sierra_program_registry: ProgramRegistry<CoreType, CoreLibfunc>,
+}
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: I think it generates nothing, it is used to query function types to validate them. 
Can we adjust the name so it is consistent with the purpose?

MaksymilianDemitraszek (commented):
OK, with merging it but please adjust the SierraCasmGenerator name, because it does not make much sense.

extensions/scarb-snforge-test-collector/src/metadata.rs
```diff
@@ -0,0 +1,112 @@
...
+use cairo_lang_project::{AllCratesConfig, SingleCrateConfig};
+use cairo_lang_utils::ordered_hash_map::OrderedHashMap;
+use camino::{Utf8Path, Utf8PathBuf};
+use scarb_metadata::{CompilationUnitMetadata, Metadata, PackageMetadata};
+use smol_str::SmolStr;
+use std::path::PathBuf;
+
+/// Represents a dependency of a Cairo project
+#[derive(Debug, Clone)]
+pub struct LinkedLibrary {
```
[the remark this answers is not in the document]
maciektr: Umm, this would probably translate to compilation unit components, but since you are dropping scarb project abstraction I am ok with this struct having a different name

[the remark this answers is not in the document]
MaksymilianDemitraszek: For me the value here is, we are going to use the same name when discussing thins between teams :p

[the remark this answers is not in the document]
maciektr: I see your point and it is generally true. However, this is just a subset of component metadata, and a very primitive one. In fact, when looking at usages I would be in favour of dropping the struct completely - it could be passed as a plain tuple without any damage to readability.

extensions/scarb-snforge-test-collector/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+            snforge_target_dir.join(&format!("{}.snforge_sierra.json", package_metadata.name));
+        let output_file = File::options()
+            .create(true)
+            .write(true)
+            .truncate(true)
+            .open(&output_path)?;
+
+        output_file
+            .lock_exclusive()
+            .with_context(|| format!("Couldn't lock the output file = {output_path}"))?;
```
mkaput, line 63: FYI I'm not requesting changing this, but we use a different error messages style in Scarb. Lower-case, not ending with a dot, no English abbreviations, and surrounding variables with backslashes, except the last one goes like this `: {var}` (instead of `=`)

So this would be more Scarb-ish:

```suggestion
            .with_context(|| format!("failed to lock the output file: {output_path}"))?;
```

extensions/scarb-snforge-test-collector/src/compilation/test_collector.rs
```diff
@@ -0,0 +1,248 @@
...
+    };
+
+    let main_crate_id =
+        insert_lib_entrypoint_content_into_db(db, crate_name, crate_root, lib_content);
+
+    if DiagnosticsReporter::stderr().check(db) {
+        return Err(anyhow!(
+            "Failed to compile test artifact, for detailed information go through the logs above"
+        ));
+    }
```
mkaput, line 120: FYI https://docs.rs/anyhow/latest/anyhow/macro.ensure.html

extensions/scarb-snforge-test-collector/src/metadata.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: Ok, your call

mkaput (approved):
Looks okayish, I haven't dived deeper into copy-pasted logic; will wait for tests with this.

extensions/scarb-snforge-test-collector/src/main.rs
piotmag769: It is consistent with our error message format, the only place this extension will be used in is forge

extensions/scarb-snforge-test-collector/src/compilation/test_collector.rs
piotmag769: I think it would make it less readable

extensions/scarb-snforge-test-collector/src/metadata.rs
```diff
@@ -0,0 +1,112 @@
...
+        dependencies
+    }
+
+    pub fn corelib_path(&self) -> Result<PathBuf> {
+        let corelib = self
+            .unit_metadata
+            .components
+            .iter()
+            .find(|du| du.name == "core")
+            .context("corelib could not be found")?;
```
cptartur, line 69:
```suggestion
            .context("Corelib could not be found")?;
```

extensions/scarb-snforge-test-collector/src/metadata.rs
```diff
@@ -0,0 +1,112 @@
...
+            .components
+            .iter()
+            .map(|component| {
+                (
+                    SmolStr::from(&component.name),
+                    SingleCrateConfig {
+                        edition: if let Some(edition) = self
+                            .metadata
+                            .get_package(&component.package)
+                            .unwrap()
```
cptartur, line 85: We should probably use `expect `instead of `unwrap` so we have more descriptive panic message.

cptartur (commented):
Assuming most of the logic stayed the same as in forge, looks fine

extensions/scarb-snforge-test-collector/src/compilation.rs
```diff
@@ -0,0 +1,44 @@
...
+use serde::Serialize;
+
+use crate::compilation::test_collector::{collect_tests, TestCaseRaw};
+use crate::crate_collection::{CrateLocation, TestCompilationTarget};
+use crate::metadata::CompilationUnit;
+
+pub mod test_collector;
+
+#[derive(Debug, Clone, Serialize, PartialEq)]
+pub struct CompiledTestCrateRaw {
```
piotmag769, line 13: Got rid of the generics to match in https://github.com/foundry-rs/starknet-foundry/pull/1157

maciektr (approved):

PR #887, opened 2023-11-08
Update guide for editions

maciektr:


mkaput (approved):

website/docs/guides/creating-a-new-package.md
```diff
@@ -22,6 +22,7 @@ Let's take a closer look at `Scarb.toml`:
 [package]
 name = "hello_world"
 version = "0.1.0"
+edition = "2023_01"
```
mkaput, line 25:
```suggestion
edition = "2023_10"
```

mkaput (changes requested):

website/docs/guides/creating-a-new-package.md
maciektr: Actually, 01 is default now image

mkaput: `01` is default, because that's the edition you get when you omit the `edition` field.

we want to push people to start new projects with `10`, see Cairo rel notes :)

PR #908, opened 2023-11-15
Verify that package tarball compiles

### LINKED ISSUE #736: Verify that the package builds after packaging it into a tarball
### Problem

`scarb package` should check if the generated tarball actually builds. This should mimic Cargo's behaviour (also the `--no-verify` flag behaviour).

### Proposed Solution

Fix this [TODO comment](https://github.com/software-mansion/scarb/blob/f307c4c6f794617446161edccc5459f78c8995b7/scarb/src/ops/package.rs#L113). Basically, port Cargo's logic.

### Notes

_No response_

szymmis:
Resolves #736

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -319,6 +319,18 @@ pub struct TestArgs {
     pub args: Vec<OsString>,
 }
 
+/// Arguments accepted by both the `package` and the `publish` command.
+#[derive(Parser, Clone, Debug)]
+pub struct PackageSharedArgs {
+    /// Allow working directories with uncommitted VCS changes to be packaged.
+    #[arg(long)]
+    pub allow_dirty: bool,
+
+    /// Do not verify the contents by building them.
+    #[arg(long)]
+    pub no_verify: bool,
+}
+
 /// Arguments accepted by the `package` command.
 #[derive(Parser, Clone, Debug)]
 pub struct PackageArgs {
@@ -326,13 +338,8 @@ pub struct PackageArgs {
     #[arg(short, long)]
     pub list: bool,
 
-    /// Allow working directories with uncommitted VCS changes to be packaged.
-    #[arg(long)]
-    pub allow_dirty: bool,
-
-    /// Don't verify the contents by building them
-    #[arg(long)]
-    pub no_verify: bool,
+    #[clap(flatten)]
+    pub shared_args: PackageSharedArgs,
 
     #[command(flatten)]
     pub packages_filter: PackagesFilter,
@@ -345,13 +352,8 @@ pub struct PublishArgs {
     #[arg(long, value_name = "URL")]
     pub index: Url,
 
-    /// Allow working directories with uncommitted VCS changes to be packaged.
-    #[arg(long)]
-    pub allow_dirty: bool,
-
-    /// Don't verify the contents by building them
-    #[arg(long)]
-    pub no_verify: bool,
+    #[clap(flatten)]
+    pub shared_args: PackageSharedArgs,
 
     #[command(flatten)]
     pub packages_filter: PackagesFilter,
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -330,6 +330,10 @@ pub struct PackageArgs {
     #[arg(long)]
     pub allow_dirty: bool,
 
+    /// Don't verify the contents by building them
+    #[arg(long)]
+    pub no_verify: bool,
+
     #[command(flatten)]
     pub packages_filter: PackagesFilter,
 }
@@ -344,6 +348,10 @@ pub struct PublishArgs {
     /// Allow working directories with uncommitted VCS changes to be packaged.
     #[arg(long)]
     pub allow_dirty: bool,
+
+    /// Don't verify the contents by building them
+    #[arg(long)]
+    pub no_verify: bool,
 
     #[command(flatten)]
     pub packages_filter: PackagesFilter,
--- a/scarb/src/ops/package.rs
+++ b/scarb/src/ops/package.rs
@@ -7,7 +7,6 @@ use camino::Utf8PathBuf;
 use indoc::{formatdoc, indoc, writedoc};
 
 use crate::core::registry::package_source_store::PackageSourceStore;
-use crate::internal::fsx::remove_dir_all;
 use crate::sources::client::PackageRepository;
 
 use scarb_ui::components::Status;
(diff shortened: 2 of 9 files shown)


scarb/tests/local_registry.rs
```diff
@@ -150,6 +150,9 @@ fn publish() {
             .success()
             .stdout_matches(formatdoc! {r#"
             [..] Packaging {name} v{version} ([..])
+            [..] Verifying {name}-{version}.tar.zst
+            [..] Compiling {name} v{version} ([..])
+            [..]  Finished release target(s) in [..]
```
maciektr, line 155: Do we need to see this? 🤔

scarb/tests/package.rs
```diff
@@ -1190,3 +1205,52 @@ fn no_lib_target() {
...
+        [..] Packaging foo v1.0.0 [..]
+        [..] Verifying foo-1.0.0.tar.zst
+        [..] Compiling foo v1.0.0 ([..])
+        error: Skipped tokens. Expected: [..]
+         --> [..]
+        .
+        ^
+
+
+        error: could not compile `foo` due to previous error
```
maciektr, line 1233: We would need some error msg context to state that verification has failed.

scarb/tests/local_registry.rs
szymmis: That's how Cargo behaves

image

maciektr: Oh, ok, haven't realize!

scarb/src/bin/scarb/args.rs
```diff
@@ -330,6 +330,10 @@ pub struct PackageArgs {
     #[arg(long)]
     pub allow_dirty: bool,
 
+    /// Don't verify the contents by building them
```
mkaput, line 333:
```suggestion
    /// Do not verify the contents by building them.
```

scarb/src/bin/scarb/args.rs
```diff
@@ -345,6 +349,10 @@ pub struct PublishArgs {
...
```
mkaput, line 352:
```suggestion
    /// Do not verify the contents by building them.
```

scarb/src/ops/publish.rs
```diff
@@ -10,6 +10,7 @@ use crate::sources::RegistrySource;
 pub struct PublishOpts {
     pub index_url: Url,
     pub allow_dirty: bool,
+    pub verify: bool,
```
mkaput, line 13: hmm, a nitpick: what about embedding `PackageOpts` here instead of copy-pasting its fields?

we could also share `clap` arguments that way, thanks to `#[clap(flatten)]`

scarb/tests/package.rs
```diff
@@ -1190,3 +1205,55 @@ fn no_lib_target() {
...
+        [..] Packaging foo v1.0.0 [..]
+        [..] Verifying foo-1.0.0.tar.zst
+        [..] Compiling foo v1.0.0 ([..])
+        error: Skipped tokens. Expected: [..]
+         --> [..]
+        .
+        ^
+
+        
+        error: failed to verify package tarball
```
mkaput, line 1233: hmm, that's a lot of `\n`s. don't we happen to by mistake add more newlines than needed?

scarb/src/ops/package.rs
```diff
@@ -160,7 +163,10 @@ fn package_one_impl(
 
     let uncompressed_size = tar(pkg_id, recipe, &mut dst, ws)?;
 
-    // TODO(mkaput): Verify.
+    if opts.verify {
+        dst.seek(SeekFrom::Start(0))?;
+        run_verify(ws, pkg, &dst).with_context(|| "failed to verify package tarball")?;
```
mkaput, line 168:
```suggestion
        run_verify(ws, pkg, &dst).context("failed to verify package tarball")?;
```

scarb/src/ops/package.rs
```diff
@@ -285,6 +291,39 @@ fn prepare_archive_recipe(pkg: &Package, opts: &PackageOpts) -> Result<ArchiveRe
     Ok(recipe)
 }
 
+fn run_verify(ws: &Workspace<'_>, pkg: &Package, tar: &FileLockGuard) -> Result<()> {
```
mkaput, line 294:
```suggestion
fn run_verify(pkg: &Package, tar: &FileLockGuard, ws: &Workspace<'_>) -> Result<()> {
```

we used to always put `ws` or `config` as the last argument

scarb/src/ops/package.rs
```diff
@@ -160,7 +163,10 @@ fn package_one_impl(
 
     let uncompressed_size = tar(pkg_id, recipe, &mut dst, ws)?;
 
-    // TODO(mkaput): Verify.
+    if opts.verify {
+        dst.seek(SeekFrom::Start(0))?;
```
mkaput, line 167: It would be more defensive programming if we put this `seek` call within of `run_verify`.

scarb/src/ops/package.rs
```diff
@@ -285,6 +291,39 @@ fn prepare_archive_recipe(pkg: &Package, opts: &PackageOpts) -> Result<ArchiveRe
...
+
+    if dst.exists() {
+        remove_dir_all(&dst)?;
+    }
+
+    // From Cargo:
+    // We don't need to set the Modified Time, as it's not relevant to verification
+    // and it errors on filesystems that don't support setting a modified timestamp
+    ar.set_preserve_mtime(false);
+    ar.unpack(dst.parent().unwrap())?;
```
mkaput, line 311: What do you think about the following? I don't feel super attached to this as I see quite a bit of changes would be needed to achieve this:

---

I'd rather reuse our existing logic for extraction, that also happens to perform some additional validation: https://github.com/software-mansion/scarb/blob/858300171f3904f2525c2ba41b3d426b61ab8764/scarb/src/core/registry/package_source_store.rs#L43

Foremost: it deals with fs locking, so no concurrent `scarb package` runs will collide.

A nice benefit is that it calls `fs.recreate()` under the hood, so that's less of stuff to reason about here. (+ I wanted to comment on this anyway in the existing code). But on the other hand, we would have to somehow skip the _don't unpack if already did_ logic.

I am thinking of adding such function to `PackageSourceStore`:

```rust
impl PackageSourceStore<'_> {
    #[tracing::instrument(level = "debug", skip_all)]
    pub async fn extract_to(archive: FileLockGuard, fs: Filesystem, config: &Config) -> Result<Utf8PathBuf>
}
```

scarb/src/bin/scarb/args.rs
szymmis: Actually this is a help info, so no period at the end.

mkaput: Clap removes trailing periods :) So that you get proper documentation comments & help info at once. Take a look at other comments

szymmis: Okay, nervermind then

scarb/tests/package.rs
szymmis: Seems like that's how Cairo compiler reports it.

mkaput (changes requested):

scarb/src/ops/package.rs
```diff
@@ -285,6 +292,36 @@ fn prepare_archive_recipe(pkg: &Package, opts: &PackageOpts) -> Result<ArchiveRe
...
+    ws.config()
+        .ui()
+        .print(Status::new("Verifying", &pkg.id.tarball_name()));
+
+    let dst = tar.path().parent().unwrap().join(pkg.id.tarball_basename());
+    if dst.exists() {
+        remove_dir_all(&dst)?;
+    }
+
+    let fs = Filesystem::new(tar.path().parent().unwrap().into());
```
mkaput, line 305: Would it be possible to use `fs.recreate()` here?

mkaput (approved):
👏🏻 🔥

Issue #834, opened 2023-10-27
Handle checksums in registry packages


mkaput:
### Problem

Current registry client implementation is totally ignoring checksums stored in the registry.

### Proposed Solution

Implement these TODOs:

https://github.com/software-mansion/scarb/blob/b039be7fed9b24ee46f379d0fd51b0190d5c03e3/scarb/src/sources/registry.rs#L125

https://github.com/software-mansion/scarb/blob/b039be7fed9b24ee46f379d0fd51b0190d5c03e3/scarb/src/sources/registry.rs#L173

And include the checksum in the lock and verify the downloaded package checksum matches one stored in `Scarb.lock`

PR #910, opened 2023-11-16
Store package checksum in `Summary`

mkaput:
**Stack**:
- #912
- #911
- #910 ⬅

### DIFF
--- a/scarb/src/core/manifest/summary.rs
+++ b/scarb/src/core/manifest/summary.rs
@@ -8,7 +8,8 @@ use typed_builder::TypedBuilder;
 #[cfg(doc)]
 use crate::core::Manifest;
 use crate::core::{
-    DepKind, DependencyVersionReq, ManifestDependency, PackageId, PackageName, SourceId, TargetKind,
+    Checksum, DepKind, DependencyVersionReq, ManifestDependency, PackageId, PackageName, SourceId,
+    TargetKind,
 };
 
 /// Subset of a [`Manifest`] that contains only the most important information about a package.
@@ -17,7 +18,7 @@ use crate::core::{
 #[derive(Clone, Debug)]
 pub struct Summary(Arc<SummaryInner>);
 
-#[derive(TypedBuilder, Debug)]
+#[derive(TypedBuilder, Clone, Debug)]
 #[builder(builder_type(name = SummaryBuilder))]
 #[builder(builder_method(vis = ""))]
 #[builder(build_method(into = Summary))]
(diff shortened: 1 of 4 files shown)


scarb/src/core/manifest/summary.rs
```diff
@@ -51,6 +54,10 @@ impl Summary {
         SummaryInner::builder()
     }
 
+    pub fn set_checksum(&mut self, cksum: Checksum) {
+        Arc::make_mut(&mut self.0).checksum = Some(cksum);
```
mkaput, line 58: I stole this black magic from Cargo 🤯

scarb/src/core/registry/client/cache.rs
```diff
@@ -139,14 +139,16 @@ impl<'c> RegistryClientCache<'c> {
...
         // Skip downloading if the package already has been.
         if self.is_package_downloaded(package).await {
             trace!("found cached archive which is not empty, skipping download");
             let tarball_name = package.tarball_name();
-            return self
+            let file = self
                 .dl_fs
-                .open_rw(&tarball_name, &tarball_name, self.config);
+                .open_rw(&tarball_name, &tarball_name, self.config)?;
+            let checksum = self.get_record_maybe_uncached(package).await?.checksum;
```
maciektr, line 150: This may be personal impression, but I would find it cleaner if this has been called only once for both cases (instead of hiding alternative call in `verify_checksum` function). The `verify_checksum` wouldn't even need to take `self` then, this could be a separate struct all together.

maciektr (approved):

scarb/src/core/registry/client/cache.rs
```diff
@@ -139,14 +139,16 @@ impl<'c> RegistryClientCache<'c> {
     pub async fn download_with_cache_and_verify_checksum(
         &self,
         package: PackageId,
-    ) -> Result<FileLockGuard> {
+    ) -> Result<(FileLockGuard, Checksum)> {
```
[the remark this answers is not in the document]
mkaput: Renamed this to `download_and_verify_with_cache` as part of #834 as that's the PR which did the original change.

scarb/src/core/registry/client/cache.rs
mkaput: Note that both logic branches diverge before calling `get_record_maybe_uncached` + there are logic branches which do not need this call. This means, that extracting this value would be suboptimal, as for some branches, the computation result would be discarded.

I did a bit of rewiring in #834 though.

PR #912, opened 2023-11-16
Store checksums for registry deps in Scarb.lock

mkaput:


scarb/src/core/resolver.rs
```diff
@@ -51,6 +53,10 @@ impl Resolve {
         self.graph
             .neighbors_directed(package_id, petgraph::Direction::Outgoing)
     }
+
+    pub fn summary(&self, package_id: PackageId) -> &Summary {
+        &self.summaries[&package_id]
+    }
```
maciektr, line 59:
```suggestion
    pub fn fetch_summary(&self, package_id: PackageId) -> &Summary {
        &self.summaries.get(&package_id).expect("package has to exist in resolve graph")
    }
```

mkaput: `impl Index for HashMap` does exactly this: https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#1326-1343. The error message is a bit generic, but I wouldn't pay so much attention here to this.

As for `fetch_` prefix: we used to use it when we return `Result<T>` in getters, so this doesn't fit here.

maciektr (approved):

PR #914, opened 2023-11-17
Move validation before usage in build-external glob selectors

maciektr:


szymmis (approved):
What is this change related to? Asking about the context.

maciektr:
@szymmis Quite frankly to nothing, I've glanced over the starknet contract compiler when discussing some possibilities with sncast team and noticed this code runs validations after using the user input.

PR #920, opened 2023-11-20
Add `dev-dependencies`

### LINKED ISSUE #717: Support dev-dependencies
### Problem

Internally, manifest dependencies can define target kind. Allow reading dev-dependencies (for test targets) from manifest.  

### Proposed Solution

- Add `dev-dependencies` field to [`TomlManifest`](https://github.com/software-mansion/scarb/blob/3ba70df198efbfe5b36a7f25d1bb5a4d16ab0d71/scarb/src/core/manifest/toml_manifest.rs#L40). Note, that `TomlManifest` represents the toml structure of Scarb manifest files. The `dev-dependencies`, similarly to `dependencies` should allow inheriting dependency specification from workspace manifest via `MaybeTomlWorkspaceDependency` pattern (this translates to an enum, that accepts either dependency specification, or `workspace = true`, all under `MaybeWorkspace` type). This inheritance will be resolved later on (see `to_manifest` for reference). 
- Update [`TomlManifest::to_manifest`](https://github.com/software-mansion/scarb/blob/3ba70df198efbfe5b36a7f25d1bb5a4d16ab0d71/scarb/src/core/manifest/toml_manifest.rs#L366) function, to take `dev-dependencies` into account. Note, that this function is responsible for translating toml-based structure of manifest files, to our internal representation - `Manifest`. 
- To do that, you will need to parse `dev-dependencies` similarly to [`dependencies`](https://github.com/software-mansion/scarb/blob/3ba70df198efbfe5b36a7f25d1bb5a4d16ab0d71/scarb/src/core/manifest/toml_manifest.rs#L409) field. First, workspace inheritance is resolved, then the `TomlDependency`, which represents dependency spec in toml format is translated to `ManifestDependency`, which is our internal dependency specification representation, via `to_dependency` (note this is analogous to `TomlManifest`, `Manifest` and `to_manifest` flow described before). 
- Parsed `dev-dependencies` spec can be passed to a single vector of dependencies, together with `dependencies`.  The `ManifestDependency` struct (i.e. the parsed, internal representation of dependency specification) has a `kind` field, storing an enum called `DepKind`. This enum can be either `Normal` - meaning dependencies that are always included in the compilation unit, or `Target(TargetKind)`, which specifies that this dependency should only be included in compilation of a single target kind. 
- Please modify `to_dependency` so that you can define a dependency kind and set it to `test` target for all the `dev-dependencies`. Note, that you can use `kind` method on `ManifestDependency` builder.   
- When you pass a vector of dependencies, with appropriate dependency kind specification, the resolver will take care of plugin them in for appropriate builds - you should be good to go!

### Notes

Take a look at [Scarb core data model graph](https://github.com/software-mansion/scarb/blob/main/ARCHITECTURE.md#core-data-model) for more insight on relations between Scarb internals.

tomek0123456789:
Closes #717

### DIFF
--- a/scarb/src/core/manifest/toml_manifest.rs
+++ b/scarb/src/core/manifest/toml_manifest.rs
@@ -412,8 +412,7 @@ impl TomlManifest {
             PackageId::new(name, version, source_id)
         };
 
-        let mut dependencies =
-            Vec::with_capacity(self.dependencies.iter().len() + self.dev_dependencies.iter().len());
+        let mut dependencies = Vec::new();
         let toml_deps = zip(self.dependencies.iter().flatten(), repeat(DepKind::Normal));
         let toml_dev_deps = zip(
             self.dev_dependencies.iter().flatten(),
--- a/scarb/src/core/manifest/toml_manifest.rs
+++ b/scarb/src/core/manifest/toml_manifest.rs
@@ -22,8 +22,8 @@ use crate::core::manifest::{ManifestDependency, ManifestMetadata, Summary, Targe
 use crate::core::package::PackageId;
 use crate::core::source::{GitReference, SourceId};
 use crate::core::{
-    DependencyVersionReq, ManifestBuilder, ManifestCompilerConfig, PackageName, TargetKind,
-    TestTargetProps, TestTargetType,
+    DepKind, DependencyVersionReq, ManifestBuilder, ManifestCompilerConfig, PackageName,
+    TargetKind, TestTargetProps, TestTargetType,
 };
 use crate::internal::fsx;
 use crate::internal::fsx::PathBufUtf8Ext;
@@ -42,6 +42,7 @@ pub struct TomlManifest {
     pub package: Option<Box<TomlPackage>>,
     pub workspace: Option<TomlWorkspace>,
     pub dependencies: Option<BTreeMap<PackageName, MaybeTomlWorkspaceDependency>>,
+    pub dev_dependencies: Option<BTreeMap<PackageName, MaybeTomlWorkspaceDependency>>,
     pub lib: Option<TomlTarget<TomlLibTargetParams>>,
     pub cairo_plugin: Option<TomlTarget<TomlExternalTargetParams>>,
     pub test: Option<Vec<TomlTarget<TomlExternalTargetParams>>>,
@@ -96,6 +97,7 @@ pub struct TomlWorkspace {
     pub members: Option<Vec<String>>,
     pub package: Option<PackageInheritableFields>,
     pub dependencies: Option<BTreeMap<PackageName, TomlDependency>>,
+    pub dev_dependencies: Option<BTreeMap<PackageName, TomlDependency>>,
     pub scripts: Option<BTreeMap<SmolStr, ScriptDefinition>>,
     pub tool: Option<TomlToolsDefinition>,
 }
@@ -418,14 +420,44 @@ impl TomlManifest {
                     .and_then(|deps| deps.get(name.as_str()))
                     .cloned()
                     .ok_or_else(|| anyhow!("dependency `{}` not found in workspace", name.clone()))?
-                    .to_dependency(name.clone(), workspace_manifest_path)
+                    .to_dependency(name.clone(), workspace_manifest_path, DepKind::Normal)
             };
             let toml_dep = toml_dep
                 .clone()
-                .map(|dep| dep.to_dependency(name.clone(), manifest_path))?
+                .map(|dep| dep.to_dependency(name.clone(), manifest_path, DepKind::Normal))?
                 .resolve(name.as_str(), inherit_ws)?;
             dependencies.push(toml_dep);
         }
+
+        let mut dev_dependencies = Vec::new();
+        for (name, toml_dep) in self.dev_dependencies.iter().flatten() {
+            let inherit_ws = || {
+                workspace
+                    .dependencies
+                    .as_ref()
+                    .and_then(|deps| deps.get(name.as_str()))
+                    .cloned()
+                    .ok_or_else(|| anyhow!("dependency `{}` not found in workspace", name.clone()))?
+                    .to_dependency(
+                        name.clone(),
+                        workspace_manifest_path,
+                        DepKind::Target(TargetKind::TEST),
+                    )
+            };
+            let toml_dep = toml_dep
+                .clone()
+                .map(|dep| {
+                    dep.to_dependency(
+                        name.clone(),
+                        manifest_path,
+                        DepKind::Target(TargetKind::TEST),
+                    )
+                })?
+                .resolve(name.as_str(), inherit_ws)?;
+            dev_dependencies.push(toml_dep);
+        }
+
+        dependencies.extend(dev_dependencies);
 
         let no_core = package.no_core.unwrap_or(false);
 
@@ -871,8 +903,9 @@ impl TomlDependency {
         &self,
         name: PackageName,
         manifest_path: &Utf8Path,
+        dep_kind: DepKind,
     ) -> Result<ManifestDependency> {
-        self.resolve().to_dependency(name, manifest_path)
+        self.resolve().to_dependency(name, manifest_path, dep_kind)
     }
 }
 
@@ -881,6 +914,7 @@ impl DetailedTomlDependency {
         &self,
         name: PackageName,
         manifest_path: &Utf8Path,
+        dep_kind: DepKind,
     ) -> Result<ManifestDependency> {
         let version_req = self
             .version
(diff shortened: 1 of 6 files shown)


scarb/src/core/publishing/manifest_normalization.rs
```diff
@@ -38,6 +38,7 @@ pub fn prepare_manifest_for_publish(pkg: &Package) -> Result<TomlManifest> {
         package,
         workspace: None,
         dependencies,
+        dev_dependencies: None,
```
mkaput, line 41: Hmm, interestingly, Cargo seems to include dev deps in its packages: https://github.com/rust-lang/cargo/blob/3eb246cd875964b7ee33acbcf30e146ade7a11b9/src/cargo/util/toml/mod.rs#L276-L280

Not sure how we should do in Scarb though 🤔 I don't understand the rationale behind this in Cargo.

scarb/tests/build.rs
```diff
@@ -687,3 +687,49 @@ fn edition_must_exist() {
...
+               Compiling x v1.0.0 ([..]/Scarb.toml)
+            error: Identifier not found.
+             --> [..]/src/foo.cairo:2:13
+                    use q::f0;
+                        ^
+
+
+            error: could not compile `x` due to previous error
+        "#});
+}
```
mkaput, line 735: Add trailing end lines to files :)

I recommend configuring your editor to do this automatically for you on save.

scarb/tests/build.rs
```diff
@@ -687,3 +687,49 @@ fn edition_must_exist() {
...
+    let t = TempDir::new().unwrap();
+    let q = t.child("q");
+    ProjectBuilder::start()
+        .name("q")
+        .build(&q);
+    ProjectBuilder::start()
+        .name("x")
+        .dev_dep("q", Dep.path("./q"))
+        .src("src/foo.cairo", r#"
+        use q::f0;
```
mkaput, line 702: This will break, the number `0` is non-deterministic and will be different if you run the full test suite. You have to provide library source code manually. ;) Reason? To enforce test developers to not rely on default values assumed by `ProjectBuilder`.

scarb/tests/build.rs
```diff
@@ -687,3 +687,49 @@ fn edition_must_exist() {
...
+
+    t.child("src/lib.cairo")
+        .write_str( r#"
+        mod foo;
+
+        fn not_working() {
+            foo::foo_not_working();
+        }
+        "#)
+        .unwrap();
```
mkaput, line 718: You should just call `.lib_cairo` on `ProjectBuilder`

scarb/tests/build.rs
```diff
@@ -687,3 +687,49 @@ fn edition_must_exist() {
...
+#[test]
+fn dev_dep_used_outside_tests() {
+    let t = TempDir::new().unwrap();
+    let q = t.child("q");
+    ProjectBuilder::start()
+        .name("q")
+        .build(&q);
+    ProjectBuilder::start()
+        .name("x")
+        .dev_dep("q", Dep.path("./q"))
```
mkaput, line 700:
```suggestion
        .dev_dep("q", &q)
```

there's `impl DepBuilder for &ChildPath`.

mkaput (commented):
Minor stylistic comments. I haven't looked yet at the meat.

scarb/src/core/publishing/manifest_normalization.rs
maciektr: @mkaput @tomek0123456789 Isn't this important to have it though?
What if pkg A has a test in `src/`, that depends on `alexandria` wired as `dev-dep`. Then pkg A can be published to registry, and downloaded as (normal) dependency of pkg B. When we want to run tests of B, the `cfg(test)` is enabled for all packages. Wouldn't we fail to compile tests of pkg B? 🤔 
Maybe this would be worth exploring with some experiment?

mkaput: Hmm, but we somehow do not run tests of dependencies anyway? How is this accomplished?

Because what I'd expect here is that no tests are compiled in alexandria 🤔

maciektr: @mkaput 

The other crates are not searched for tests. If they are compiled, is something I suggest we should verify. 

https://github.com/starkware-libs/cairo/blob/d4edafc66ead03ee67c86d53aad39fa48b525577/crates/cairo-lang-test-plugin/src/lib.rs#L52

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -409,20 +412,28 @@ impl TomlManifest {
             PackageId::new(name, version, source_id)
         };
 
-        let mut dependencies = Vec::new();
-        for (name, toml_dep) in self.dependencies.iter().flatten() {
+        let mut dependencies =
+            Vec::with_capacity(self.dependencies.iter().len() + self.dev_dependencies.iter().len());
```
mkaput, line 416:
```suggestion
        let mut dependencies = Vec::new();
```

Nitpick: a negligible micro-optimization (this vector won't grow beyond 100ish items in sensible projects), which hinders readability a bit ;)

mkaput (commented):
This looks very good! Let's resolve the topic of packaging dev deps, and then I'll ✅ this.

scarb/src/core/publishing/manifest_normalization.rs
maciektr: I generally agree, but I still think we should first verify if this very PR does not break something this works now, especially since it's not hard to verify with a quick test.

mkaput: It does not break anything because nobody is using this feature. If that's going to be a problem, users will just still use `[dependencies]` section.

maciektr: FYI Tomek has created an issue to tackle this later :D 
https://github.com/software-mansion/scarb/issues/940

scarb/tests/build.rs
```diff
@@ -687,3 +687,58 @@ fn edition_must_exist() {
...
+        .lib_cairo(
+            r#"
+        mod foo;
+
+        fn not_working() {
+            foo::foo_not_working();
+        }
+        "#,
+        )
+        .build(&t);
```
maciektr, line 727: Nit: the second file does not really add a lot here. You can simplify with 
```suggestion
        .dev_dep("q", &q)
        .lib_cairo(indoc! {r#"
            use q::dev_dep_function;

            fn foo_not_working() {
                dev_dep_function();
            }
        "#})
        .build(&t);
```

scarb/tests/build.rs
```diff
@@ -687,3 +687,58 @@ fn edition_must_exist() {
...
+    let q = t.child("q");
+    ProjectBuilder::start()
+        .name("q")
+        .lib_cairo(
+            r#"
+            fn dev_dep_function() -> felt252 {
+                42
+            }
+            "#,
+        )
```
maciektr, line 703: Nit: can fit to a single line
```suggestion
        .lib_cairo("fn dev_dep_function() -> felt252 { 42 }")
```

scarb/tests/local_registry.rs
```diff
@@ -186,7 +186,7 @@ fn publish() {
             {
                 "v": "1.0.0",
                 "deps": [],
-                "cksum": "sha256:3771f51595f7df697a822823989714b3c9a83764a67fa517ac6d8df3cfc642bf",
+                "cksum": "sha256:8c612d3a5ec513259f786b692c91aa0f37d9d30e24428ae4d72dc7d39d00b45c",
```
mkaput, line 189: hmm why did checksums change?

mkaput (changes requested):

scarb/tests/build.rs
```diff
@@ -687,3 +687,58 @@ fn edition_must_exist() {
...
+            [..] Compiling x v1.0.0 ([..])
+            error: Identifier not found.
+             --> [..]/src/foo.cairo:2:13
+                    use q::dev_dep_function;
+                        ^
+
+
+            error: could not compile `x` due to previous error
+        "#});
+}
```
maciektr, line 744: Can you add a second test case, where this actually works?

scarb/tests/local_registry.rs
tomek0123456789: I think it's because of the `[dev-dependencies]` field now added to each `ProjectBuilder`

maciektr: I guess
https://github.com/software-mansion/scarb/blob/5cec6ef3de89d537880103c80856ae46fdec8540/scarb/src/core/publishing/manifest_normalization.rs#L41

mkaput: Can we avoid adding `[dev-dependencies]` in `ProjectBuilder` if not needed, then?

maciektr (approved):
👏 👏 👏

PR #937, opened 2023-11-24
Implement lockfiles checksums verification

### LINKED ISSUE #907: Support checksums in lockfiles for registry sources

mkaput:
fix #907

### DIFF
--- a/scarb/src/core/resolver.rs
+++ b/scarb/src/core/resolver.rs
@@ -164,8 +164,8 @@ impl Resolve {
                 }
 
                 // If our checksum has not been calculated, then it could mean that future Scarb
-                // figured out how to do it or more realistically we were overridden with a source
-                // that does not have checksums.
+                // figured out how to do it or the source has been shadowed by with
+                // a different one thanks to some unknown future logic.
                 (None, Some(_)) => {
                     bail!(formatdoc! {"
                         checksum for `{id}` could not be calculated, but a checksum is listed in \
--- a/scarb/src/core/resolver.rs
+++ b/scarb/src/core/resolver.rs
@@ -155,7 +155,7 @@ impl Resolve {
 
                         this could be indicative of a few possible situations:
 
-                            * the source `{source_id}` did not previously support checksums,
+                            * the source `{source_id}` did not previously support checksums, \
                               but was replaced with one that does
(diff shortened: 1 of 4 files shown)


szymmis (approved):

scarb/src/core/resolver.rs
```diff
@@ -96,3 +100,104 @@ impl From<TargetKind> for DependencyEdge {
...
+                              but was replaced with one that does
+                            * newer Scarb implementations know how to checksum this source, \
+                              but this older implementation does not
+                            * the lock file is corrupt
+                    "});
+                }
+
+                // If our checksum has not been calculated, then it could mean that future Scarb
+                // figured out how to do it or more realistically we were overridden with a source
+                // that does not have checksums.
```
maciektr, line 168: > we were overridden with a source that does not have checksums

I think the `source_id` will differ then. For `core` we get `std` source not `registry` source at this point in the resolver logic.

maciektr (approved):

scarb/tests/lockfile_checksums.rs
```diff
@@ -0,0 +1,183 @@
...
+        .current_dir(&t)
+        .assert()
+        .failure()
+        .stdout_matches(indoc! {r#"
+        error: checksum for `bar v1.0.0 ([..])` was not previously calculated, but now it could be
+
+        this could be indicative of a few possible situations:
+
+            * the source `[..]` did not previously support checksums,
+              but was replaced with one that does
```
mkaput, line 179:
```suggestion
            * the source `[..]` did not previously support checksums, but was replaced with one that does
```

scarb/src/core/resolver.rs
```diff
@@ -96,3 +100,104 @@ impl From<TargetKind> for DependencyEdge {
...
```
mkaput, line 168:
```suggestion
                // If our checksum has not been calculated, then it could mean that future Scarb
                // figured out how to do it or the source has been shadowed by with
                // a different one thanks to some unknown future logic.
```

scarb/src/core/resolver.rs
mkaput: [68370cd](https://github.com/software-mansion/scarb/pull/937/commits/68370cd36ea514ac01d38731404c3aa7b6e8967e)

PR #939, opened 2023-11-24
Do not compile packages when publishing them in tests

mkaput:


maciektr (approved):

utils/scarb-test-support/src/registry/local.rs
```diff
@@ -22,6 +22,7 @@ impl LocalRegistry {
         f(&t);
         Scarb::quick_snapbox()
             .arg("publish")
+            .arg("--no-verify")
```
szymmis, line 25: There were problems related to that that were caught by that verification step, so I'm wondering why we want to disable it?

maciektr: Performance - compiling cairo is by the order of magnitude slower than anything scarb does

szymmis: To be honest test performance over coverage isn't a convincing argument for me 😅.

maciektr: Well, you always need to balance those two. Is it beneficial to run the verification multiple times here, or maybe it would be enough to have a single test case that runs it?

maciektr: If you think this is beneficial here, maybe we should create test levels - skip this in ci and run with scheduled testing? 
I don't think the very example here justifies that, but wanted to share this as an idea 😛

szymmis: Writing even a one case that tests it is for sure a lot better than disabling it altogether, so I thing that is a good idea. Definitely don't want to overcompilate the whole process with test layers.

szymmis: It's only for packaging, and what I meant was a situation where you verify while uploading. That's where the errors related to Windows and file locks popped up when I was working on package validation.

maciektr: Hmm, isn't the verification step exactly identical in both cases though? 🤔 
We have https://github.com/software-mansion/scarb/blob/6868661e06beba100555edc184b1e4a7ed9b2380/scarb/tests/local_registry.rs#L153 as well, but if you have an idea for another one, then have a go at it!

szymmis: The verification step is completely the same, but the problem was somehow related to reading the archive twice, once for verification, second for uploading. That's something package test cases don't check because for me it would couple verification logic tests to registry implementation. That's why it should be checked in the registries tests, as it is now. However I do understand that this verification is run on every test case because "the call is inside" a helper function, so we might want to disable it here, and just have one case when we run the publish command by hand.

The only thing I care is to have anything that tests this special case. Hope that's clear because this message is quite a wall of text 😅.

maciektr: Ok, thanks for clarifying!  
Would you mind adding such test case when you have a moment then?

mkaput: Mind that `LocalRegistry` is something used to set up a registry fixture. We're not using it for testing local registries itself, see that this struct is not used in local registry publish test. We always trust the input put there because it's all hardcoded in tests. There's no benefit in verifying it.

This patch causes some tests speed-ups from 17s to 1s. That's a big benefit.

PR #941, opened 2023-11-24
Parse `BlockId` on forge's side -> remove starknet dependency

piotmag769:
https://github.com/foundry-rs/starknet-foundry/pull/1193

Changes:
- pass strings instead of `BlockId` structure
- remove `starknet-rs` dependency from Scarb
- fix `BlockId::Tag` bug https://github.com/software-mansion/scarb/pull/941#discussion_r1404539329


extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
```diff
@@ -284,60 +287,56 @@ fn extract_fork_config_from_args(db: &dyn SyntaxGroup, attr: &Attribute) -> Opti
...
+                return Some(
+                    u64::try_from(value.numeric_value(db).unwrap())
+                        .ok()?
+                        .to_string(),
+                );
             }
-            None
         }
         "Hash" => {
+            // TODO #1179: add range check
```
piotmag769, line 318: https://github.com/foundry-rs/starknet-foundry/issues/1179

extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
```diff
@@ -55,19 +54,20 @@ impl From<TestExpectation> for ExpectedTestResult {
...
 }
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub struct RawForkParams {
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub struct ForkParamsString {
     pub url: String,
-    pub block_id: BlockId,
+    pub block_id_type: String,
+    pub block_id_value: String,
```
piotmag769, line 67: Maybe not the best type to use, but I think it's better than copying structs from `starknet-rs`

extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
maciektr: Why do we need to have it here? Wouldn't be enough if foundry did validations on your side?

piotmag769: Not really since we wouldn't be able to show the exact place where the error occured using Cairo diagnostics

extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
```diff
@@ -55,19 +54,20 @@ impl From<TestExpectation> for ExpectedTestResult {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub enum RawForkConfig {
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub enum ForkConfigString {
```
cptartur, line 58: Why this name? I think `RawForkConfig` was okay for unparsed form.

extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
```diff
@@ -284,60 +287,56 @@ fn extract_fork_config_from_args(db: &dyn SyntaxGroup, attr: &Attribute) -> Opti
...
-    Some(RawForkConfig::Params(RawForkParams { url, block_id }))
+    Some(ForkConfigString::Params(ForkParamsString {
+        url,
+        block_id_type,
+        block_id_value,
+    }))
 }
 
-fn try_get_block_id(db: &dyn SyntaxGroup, block_id_type: &str, expr: &Expr) -> Option<BlockId> {
+fn try_get_block_id(db: &dyn SyntaxGroup, block_id_type: &str, expr: &Expr) -> Option<String> {
```
cptartur, line 306: If we are "parsing" it here anyway, why won't we pass this as enum defined by us? Seems counter-productive to first parse the block id and the return it as two strings anway.

extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
mkaput: Perhaps you could add `struct BlockId(String)` to at least increase type safety?

extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
piotmag769: Actually you are right, idk why I changed it

extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
piotmag769: I feel like that would be pretending that it is type safe, you can still put anything inside the `BlockId`. I don't think we gain anything by doing it, but maybe I'm missing sth

extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
piotmag769: That would be basically copy pasting `starknet-rs` structs, not sure if that's what we want. It's not costless to add those structs, additional thing to maintain. I guess that's up to debate, I think both of the solutions are bad

piotmag769: Discussed offline, leaving it as it is

cptartur (approved):

PR #956, opened 2023-11-28
Handle case when packaged repo does not have any commits

### LINKED ISSUE #869: Handle case when packaged repository does not have any commits

szymmis:
Resolves #869.

Unlike cargo which completely ignores VCS when using `--allow-dirty`, we still want to package the info because that seems reasonably. 

There is an edge case when `--allow-dirty` flag which reads a hash of HEAD commit fails due to repository not having any commits. In that case we want to treat it like there was no VCS information to include at all and just ignore it.

### DIFF
--- a/scarb/src/ops/package.rs
+++ b/scarb/src/ops/package.rs
@@ -123,12 +123,16 @@ fn extract_vcs_info(repo: PackageRepository, opts: &PackageOpts) -> Result<Optio
         "#}
     );
 
-    Ok(Some(VcsInfo {
-        path_in_vcs: repo.path_in_vcs()?,
-        git: GitVcsInfo {
-            sha1: repo.head_rev_hash()?,
-        },
-    }))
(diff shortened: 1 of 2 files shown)


scarb/src/ops/package.rs
```diff
@@ -123,12 +123,16 @@ fn extract_vcs_info(repo: PackageRepository, opts: &PackageOpts) -> Result<Optio
         "#}
     );
 
-    Ok(Some(VcsInfo {
-        path_in_vcs: repo.path_in_vcs()?,
-        git: GitVcsInfo {
-            sha1: repo.head_rev_hash()?,
-        },
-    }))
+    // If the HEAD commit has cannot be determined, we assume the repository is empty.
```
maciektr, line 126:
```suggestion
    // If the HEAD commit cannot be determined, we assume the repository is empty.
```

maciektr (approved):

PR #961, opened 2023-11-29
adds snforge test collector test

tkumor3:


piotmag769 (commented):
The idea is fine, but preferably I would add a few tests with different combinations of attributes (that's the main thing we want to test) and in different locations (src/ vs tests/). 

I just realised that I may have removed some unit tests testing the code/file collection mechanism... I guess we could restore them too
cc @cptartur 

When it comes to smoke tests I don't really have an opinion, I feel like there is a really fine line between too much tests and not enough tests and I'm not sure where it is there xD

extensions/scarb-snforge-test-collector/tests/test.rs
cptartur, line 1: Following the @piotmag769 suggestion, it'd make sense to have a test case where tests are collected from somewhere in the `src/` directory as well. Perhaps maybe a combination of that and tests in `tests/`.

Also, unit tests from there https://github.com/foundry-rs/starknet-foundry/blob/a1faecc30f70f2cc7e19073fa9cc9eb37edb2974/crates/forge/src/collecting.rs#L134 would be worth restoring here.

extensions/scarb-snforge-test-collector/tests/test.rs
```diff
@@ -0,0 +1,113 @@
...
+    assert_eq!(&json[0]["test_cases"][0]["available_gas"], &Value::Null);
+    assert_eq!(&json[0]["test_cases"][0]["expected_result"], "Success");
+    assert_eq!(&json[0]["test_cases"][0]["fork_config"], &Value::Null);
+    assert_eq!(&json[0]["test_cases"][0]["fuzzer_config"], &Value::Null);
+    assert_eq!(&json[0]["test_cases"][0]["ignored"], false);
+}
+
+const WITH_MANY_ATTRIBUTES_TEST: &str = indoc! {r#"
+    #[ignore]
+    #[fork(url: "http://your.rpc.url", block_id: BlockId::Number(123))]
```
cptartur, line 55: We should have a test fro where fork is referenced just by name https://foundry-rs.github.io/starknet-foundry/testing/fork-testing.html#configure-fork-in-the-scarbtoml as well as tests for `BlockId::Tag` and `BlockId::Hash`.

extensions/scarb-snforge-test-collector/tests/test.rs
```diff
@@ -0,0 +1,113 @@
...
+    #[should_panic]
```
cptartur, line 56: A test for something like `#[should_panic(expected: ('panic message', 'eventual second message',))]` should be added

extensions/scarb-snforge-test-collector/tests/test.rs
```diff
@@ -0,0 +1,113 @@
...
+        .current_dir(&pkg1)
+        .assert()
+        .success();
+
+    let snforge_sierra = pkg1
+        .child("target/dev/snforge/forge_test.snforge_sierra.json")
+        .read_to_string();
+
+    let json: Value = serde_json::from_str(&snforge_sierra).unwrap();
+    dbg!(&json[0]["test_cases"]);
```
maciektr, line 85:
```suggestion
```

extensions/scarb-snforge-test-collector/tests/test.rs
```diff
@@ -0,0 +1,113 @@
...
+
+use indoc::indoc;
+
+use scarb_test_support::command::Scarb;
+
+use scarb_test_support::project_builder::ProjectBuilder;
+use serde_json::Value;
+
+const SIMPLE_TEST: &str = indoc! {r#"
+    #[test]
```
maciektr, line 14: Pls a test case enclosed in cfg block

extensions/scarb-snforge-test-collector/tests/test.rs
piotmag769: +1

piotmag769 (commented):
LGTM, but I agree with Artur's suggestions

extensions/scarb-snforge-test-collector/tests/test.rs
piotmag769: Discussed offline, it's hard to unit test with signature changes (we would need to mock metadata). If it's covered in e2e I'm fine with letting it go

extensions/scarb-snforge-test-collector/tests/test.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    let snforge_sierra = pkg1
+        .child("target/dev/snforge/forge_test.snforge_sierra.json")
+        .read_to_string();
+
+    let json: Value = serde_json::from_str(&snforge_sierra).unwrap();
+    dbg!();
+    assert_eq!(
+        &json[0]["test_cases"][0]["expected_result"]["Panics"].to_string(),
+        "{\"Exact\":[{\"value\":{\"val\":[1935763301,544040307,1634625891,112]}},{\"value\":{\"val\":[1935763301,544040307,1668247140,1814066021,1853125985,6649445]}}]}"
```
cptartur, line 279: I wonder why two shortstrings get serialized to so many values instead of just two. But if it works I guess it's fine.

cptartur (approved):

PR #976, opened 2023-12-01
validate git urls

tomek0123456789:
commit-id:d8d2f542

formatting

better validation, test added

### DIFF
--- a/scarb/src/manifest_editor/add.rs
+++ b/scarb/src/manifest_editor/add.rs
@@ -1,7 +1,8 @@
 use std::mem;
 
-use anyhow::{anyhow, ensure, Result};
+use anyhow::{anyhow, ensure, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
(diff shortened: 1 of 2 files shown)


scarb/src/manifest_editor/add.rs
```diff
@@ -136,6 +136,17 @@ impl Dep {
                     .unwrap_or(git),
                 Err(_) => git,
             };
+            let git_url = Url::parse(&git)
```
maciektr, line 139: We do not actually need to add anything here. 
Please note, that we already call `Url::parse` at `133`. We only need to stop ignoring the errors it produces.
As of now we do `Err(_) => git`, but we could just drop whole match, and do `Url::parse(&git)?` instead, to bubble the error up.

maciektr (commented):
Please add a test case that triggers this

maciektr (approved):

szymmis (approved):

PR #979, opened 2023-12-04
adds max steps attribute

tkumor3:


extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
```diff
@@ -87,6 +88,8 @@ pub struct SingleTestConfig {
     pub fork_config: Option<RawForkConfig>,
     /// Custom fuzzing configuration
     pub fuzzer_config: Option<FuzzerConfig>,
+    /// custom max steps
```
cptartur, line 91:
```suggestion
    /// Custom max steps
```

cptartur (approved):

extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
```diff
@@ -140,6 +145,23 @@ pub fn forge_try_extract_test_config(
         None
     };
 
+    let max_steps = if let Some(attr) = max_steps_attr {
+        if attr.args.is_empty() {
+            None
+        } else {
```
piotmag769, line 151: Just now realised this approach is just wrong (I know it was used above, it has to be fixed too, sooner or later). If there are no args here we should just throw an error, it is incorrect to have sth like `#[max_steps]` there (same with `#[fork]` and `#[fuzzer]`). I would leave it as it is and create an issue for that.

extensions/scarb-snforge-test-collector/src/compilation/test_collector/config.rs
```diff
@@ -140,6 +145,23 @@ pub fn forge_try_extract_test_config(
...
+            extract_max_steps(attr, db).on_none(|| {
+                diagnostics.push(PluginDiagnostic {
+                    severity: Severity::Error,
+                    stable_ptr: attr.args_stable_ptr.untyped(),
+                    message: "Expected max steps must be <u32>"
```
piotmag769, line 156: Doesn't align exactly with the other error messages

tkumor3: @piotmag769 updated let me know what do you think?

piotmag769: I was thinking of "Expected max steps config must be of the form `max_steps: <u32>`"

tkumor3: Ok, but imo it sound like `#[max_steps({max_steps: 3232})]`

piotmag769: Up to you

piotmag769 (approved):

PR #985, opened 2023-12-06
Add metadata completeness validation

### LINKED ISSUE #738: Check package metadata for completeness
### Problem

Port package metadata checking logic from cargo. Start here: https://github.com/rust-lang/cargo/blob/4820624ce76b5c047566f62d615a3b39eed91501/src/cargo/ops/cargo_package.rs#L93-L95

Implement it here: https://github.com/software-mansion/scarb/blob/f307c4c6f794617446161edccc5459f78c8995b7/scarb/src/ops/package.rs#L92

### Proposed Solution

_No response_

### Notes

_No response_

tomek0123456789:
Closes #738


scarb/src/ops/package.rs
```diff
@@ -486,3 +490,54 @@ fn tar(
...
+    if !missing.is_empty() {
+        let mut things = missing[..missing.len() - 1].join(", ");
+        // `things` will be empty if and only if its length is 1 (i.e., the only case
+        // to have no `or`).
+        if !things.is_empty() {
+            things.push_str(" or ");
+        }
+        things.push_str(missing.last().unwrap());
+
+        config.ui().warn(formatdoc!(
```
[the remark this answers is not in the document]
tomek0123456789: Yup, thought the same about the message. I don't know how it should be worded, maybe something like this:
```
warn: manifest has no (description), (license OR license-file), (documentation OR homepage OR repository) under [package] section.
```
wdyt?

Will add `README` check, thanks!

scarb/src/bin/scarb/args.rs
```diff
@@ -383,6 +383,10 @@ pub struct PackageArgs {
     #[arg(short, long)]
     pub list: bool,
 
+    /// Ignore warnings about a lack of human-usable metadata
+    #[arg(long)]
+    pub no_metadata: bool,
+
```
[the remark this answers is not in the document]
tomek0123456789: Cargo actually doesn't have `no_metadata` argument in `cargo publish` (which I understand, because publishing a package with no information about it would be a bit misleading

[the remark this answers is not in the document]
szymmis: Yeah you are right, it's not an error, that's why I'm wondering now what is the use case for `--no-metadata` flag at all.

The thing is that all of these validations are only client-side and registry doesn't verify it at all. I guess its more of a friendly notice to the package creator to include info that might help others discover his package properly.

Also, the reason I've suggested it, is that I consider `package` args to be a subset of `publish` args - primarily because `package` is one of the steps of `publish` process. That's why for me, if there's an option to silence that kind of output in `package`, there should be an option to do that in `publish` too.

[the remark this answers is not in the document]
maciektr: Personally, it makes sense the way it is in Cargo (with the `--no-metadata` on package only). The package command is more generic - it only creates a tarball for distribution, but the "distribution" itself can be whatever. For some distribution methods it may make sense to omit the package metadata. On the other hand, it always makes sense to include metadata when publishing to registry.

scarb/src/ops/package.rs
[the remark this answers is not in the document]
maciektr: 1. Please do not use dots at the end of warning 
2. As much as I agree it's logically correct, I don't like the `(A), (B or C)` format (it reminds me of boolean algebra class to much). Personally, I think the cargo approach is fine - it's just a docs pointer anyway. If anything, I would just print groups as separate warnings, i.e.:
``` 
warn: manifest has no A
warn: manifest has no B or C
see: https://scarb.pretty.docs for context

File: utils/scarb-test-support/src/project_builder.rs [hunk 3]
```diff
@@ -111,6 +139,18 @@ impl ProjectBuilder {
...
+        }
+        if let Some(license) = self.license.as_ref() {
+            doc["package"]["license"] = Item::Value(Value::from(license.to_string()));
+        }
+        if let Some(readme) = self.readme.as_ref() {
+            doc["package"]["readme"] = Item::Value(Value::from(readme.to_string()));
+        }
+        if let Some(homepage) = self.homepage.as_ref() {
+            doc["package"]["homepage"] = Item::Value(Value::from(homepage.to_string()));
+        }
```
maciektr, line 153: Repetitions could be simplified with a macro

scarb/tests/local_registry.rs
```diff
@@ -150,6 +150,12 @@ fn publish() {
             .success()
             .stdout_matches(formatdoc! {r#"
             [..] Packaging {name} v{version} ([..])
+            warn: manifest has no [..]
+            warn: manifest has no [..]
+            warn: manifest has no [..]
+            warn: manifest has no [..]
+            [..]
```
maciektr, line 157:
```suggestion
             see [..] for more info
```

scarb/src/ops/package.rs
```diff
@@ -486,3 +489,48 @@ fn tar(
...
+    if !missing.is_empty() {
+        let messages = missing
+            .split(|elem| elem == " ")
+            .filter(|vec| !vec.is_empty())
+            .map(|vec| String::from("manifest has no ") + &*vec.join(" or "))
+            .collect::<Vec<String>>();
+        for message in messages {
+            config.ui().warn(message);
+        }
+        config.ui().print("see https://docs.swmansion.com/scarb/docs/reference/manifest.html#package for more info\n");
```
maciektr, line 532:
```suggestion
        config.ui().print("see https://docs.swmansion.com/scarb/docs/reference/manifest.html#package for more info");
```
Unless you feel we need it

utils/scarb-test-support/src/project_builder.rs
```diff
@@ -95,6 +103,26 @@ impl ProjectBuilder {
         self.dep("starknet", Dep.version(CAIRO_VERSION))
     }
 
+    pub fn readme(mut self, readme: impl ToString) -> Self {
```
maciektr, line 106: Did we end up using those anywhere?

maciektr (commented):
👏 👏 👏 
very minor comments attached

scarb/src/ops/package.rs
tomek0123456789: Hmm, felt like newline makes message clearer - without it in the example the error and the warn seem to be a one message instead of two separate ones, wdyt?
```
warn: manifest has no description
warn: manifest has no license or license-file
warn: manifest has no documentation or homepage or repository
see https://docs.swmansion.com/scarb/docs/reference/manifest.html#package for more info

error: cannot archive package `workspaces` without a `lib` target
help: add `[lib]` section to package manifest
 --> Scarb.toml
+   [lib]
```

or 

```
warn: manifest has no description
warn: manifest has no license or license-file
warn: manifest has no documentation or homepage or repository
see https://docs.swmansion.com/scarb/docs/reference/manifest.html#package for more info
error: cannot archive package `workspaces` without a `lib` target
help: add `[lib]` section to package manifest
 --> Scarb.toml
+   [lib]
```

utils/scarb-test-support/src/project_builder.rs
tomek0123456789: My bad, not used anywhere since we're using `--no-metadata` flag, thanks!

maciektr (approved):
❤️

PR #986, opened 2023-12-07
Use diagnostic severity in cairo formatter

maciektr:


scarb/src/ops/fmt.rs
```diff
@@ -108,15 +110,25 @@ fn print_diff(ws: &Workspace<'_>, path: &Path, diff: impl Display) {
...
-        path.display()
-    ));
+fn print_error(ws: &Workspace<'_>, path: &Path, error: FormattingError) {
+    match error {
+        FormattingError::ParsingError(error) => {
+            for entry in error.iter() {
+                let msg = entry
+                    .message()
+                    .strip_suffix('\n')
+                    .unwrap_or(entry.message());
```
szymmis, line 120: Maybe [`trim_end_matches`](https://doc.rust-lang.org/std/string/struct.String.html#method.trim_end_matches) can be used here instead to avoid this `unwrap_or` duplication, or do we want to strip exactly one `\n`?

maciektr: Exactly one - error/warn uses `println` which adds one line break at the end, so I have added this specifically to account for that.

PR #987, opened 2023-12-07
Update cairo-run docs, add no-build argument

### LINKED ISSUE #983: Cairo-run should not build the package
### Problem

Docs: 
> The scarb cairo-run command executes the main function of a local package. It does not compile any cairo code within the package so using scarb build beforehand is necessary.

(https://docs.swmansion.com/scarb/docs/extensions/cairo-run.html#using-cairo-run-in-scarb-projects)

Actual code:
https://github.com/software-mansion/scarb/blob/dc2e88b36e9109803cff222cf6197d4be0b33ff1/extensions/scarb-cairo-run/src/main.rs#L44

### Steps

_No response_

### Possible Solution(s)

I suggest removing the build from cairo-run as it would allow to run code compiled by custom plugins. Currently you have to reimplement both build and cairo-run to do that.

### Notes

_No response_

### Version

```text
2.3.1, also current main
```

maciektr:
resolves #983

### DIFF
--- a/extensions/scarb-cairo-run/src/main.rs
+++ b/extensions/scarb-cairo-run/src/main.rs
@@ -10,7 +10,7 @@ use clap::Parser;
 use indoc::formatdoc;
 use serde::Serializer;
 
-use scarb_metadata::{MetadataCommand, ScarbCommand};
+use scarb_metadata::{Metadata, MetadataCommand, ScarbCommand};
 use scarb_ui::args::PackagesFilter;
 use scarb_ui::components::Status;
 use scarb_ui::{Message, OutputFormat, Ui, Verbosity};
@@ -30,6 +30,10 @@ struct Args {
     /// Print more items in memory.
     #[arg(long, default_value_t = false)]
     print_full_memory: bool,
+
+    /// Do not rebuild the package.
+    #[arg(long, default_value_t = false)]
+    no_build: bool,
 }
 
 fn main() -> Result<()> {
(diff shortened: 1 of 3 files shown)


extensions/scarb-cairo-run/src/main.rs
```diff
@@ -41,7 +45,13 @@ fn main() -> Result<()> {
 
     let package = args.packages_filter.match_one(&metadata)?;
 
-    ScarbCommand::new().arg("build").run()?;
+    if !args.no_build {
+        let filter = PackagesFilter::generate_for::<Metadata>(vec![package.clone()].iter());
+        ScarbCommand::new()
+            .arg("build")
+            .env("SCARB_PACKAGES_FILTER", filter.to_env())
```
tomek0123456789, line 52: Why is this env now set?

tomek0123456789 (approved):
image

extensions/scarb-cairo-run/src/main.rs
maciektr: Just as we do in cairo test here: https://github.com/software-mansion/scarb/blob/31b92d4cd940539fa47b800571d2e9c8b329d97d/extensions/scarb-cairo-test/src/main.rs#L38-L43
It's used to override the PackagesFilter in the subsequent call to scarb binary. 
https://github.com/software-mansion/scarb/blob/31b92d4cd940539fa47b800571d2e9c8b329d97d/utils/scarb-ui/src/args/packages_filter.rs#L32

The PackagesFilter controls which package is build, via both `--package/--workspace` arguments, current working directory and some logic for workspace package selection. 
User can run the cairo-run extension from some package in the workspace, passing another package's name to `--package` argument. In this case, we want to build the package that has been selected by name, but simply calling `ScarbCommand` would result in building package from cwd instead!

As the subcommands are responsible for package selection on their own (in contrary to scripts, which are run in a select package / group of packages) (which cairo-run does but consuming the --package/--workspace flags, which are ignored by Scarb itself), the responsibility for this also needs to lay here.

PR #997, opened 2023-12-13
fix: scarb new edition

enitrat:
Fixed the prelude edition used to `2023_10` - as `2023_11` is not a valid edition (maybe a bug in the compiler itself, but I figured out they might have reasons to use it internally and not have it released).

### DIFF
--- a/scarb/src/ops/new.rs
+++ b/scarb/src/ops/new.rs
@@ -133,7 +133,7 @@ fn mk(
 
     // Create the `Scarb.toml` file.
     let manifest_path = canonical_path.join(MANIFEST_FILE_NAME);
-    let edition = edition_variant(Edition::latest());
+    let edition = edition_variant(Edition::V2023_10);
     fsx::write(
         &manifest_path,
         formatdoc! {r#"


maciektr:
What problem does this PR intend to solve? 🤔 The`2023_11` edition is available in some releases of Scarb.

enitrat:
The prelude 2023_11 doesn't exist inside the Cairo repository https://github.com/starkware-libs/cairo/tree/v2.4.0/corelib/src/prelude

If you use edition 2023_11 in your Scarb project, the LS crashes repeatedly. Perhaps the problem is something else? But I was told previously that 2023_11 was not a valid edition version.

maciektr:
@enitrat But the edition does exist, it only uses `*_10` prelude https://github.com/starkware-libs/cairo/blob/8f44de9ae56163ac86d158164a4c74ac454bf76c/crates/cairo-lang-filesystem/src/db.rs#L74
If LS crashes on `2023_11` edition in a version that supports it, then there is a bug in the LS. What Scarb version do you see the error on?

enitrat:
Ah, sorry! My bad. I'm using scarb 2.4. Must be something related to scarb metadata:
```
❯ scarb metadata --format-version 2
error: failed to parse manifest at: /Users/msaug/workspace/sandbox/temp/Scarb.toml

Caused by:
    TOML parse error at line 4, column 11
      |
    4 | edition = "2023_11"
      |           ^^^^^^^^^
    unknown variant `2023_11`, expected `2023_01` or `2023_10`
❯ scarb metadata --format-version

```

maciektr:
Oh yes, a version that does not support 2023_11 will reject it when parsing the manifest file. This is the expected behaviour.

enitrat:
> 
Oh yes, a version that does not support 2023_11 will reject it when parsing the manifest file. This is the expected behaviour.

I don't get it: If 2.4.0 doesn't support 2023_11, why does creating a project with `scarb new` set the edition to `2023_11`?

maciektr:
On `2.4.0`? I don't think it does. 
```
➜  Downloads ./scarb-v2.4.0-aarch64-apple-darwin/bin/scarb --version
scarb 2.4.0 (cba988e68 2023-12-06)
cairo: 2.4.0 (https://crates.io/crates/cairo-lang-compiler/2.4.0)
sierra: 1.4.0

➜  Downloads ./scarb-v2.4.0-aarch64-apple-darwin/bin/scarb new tmp
Created `tmp` package.
➜  Downloads cat tmp/Scarb.toml
[package]
name = "tmp"
version = "0.1.0"
edition = "2023_10"

# See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest.html

[dependencies]
➜  Downloads
```
`2023_11` will be set by versions that support it.

enitrat:
well I don't understand what happened 😓 
It happened to me 3 times this afternoon: scarb new -> edition = "2023_11".
I was using 2.4.0, no nightlies, from asdf 😵

maciektr:
Hi @enitrat !
It seems to me that it works as expected for `2.4.0` 
image
I will close this PR now, please open a bug report if you are able to reproduce this.

PR #1001, opened 2023-12-16
fix: mistaken error message

BerzanOrg:
When `Scarb.toml` file is like below:
```toml
[package]
name = "token"
version = "0.1.0"
cairo-version = "4.3.5"
```

`scarb build` command prints the message below:
```
error: Package token. Required Cairo version isn't compatible with current version. Should be: ^4.3.5 is: 2.4.0
error: For each package, the required Cairo version must match the current Cairo version.
```

This part is mistaken: *"... Should be: ^4.3.5 is: 2.4.0"*

The reason is that `package_version` and `current_version` is in wrong order.
```rs
ws.config().ui().error(format!(
    "Package {}. Required Cairo version isn't compatible with current version. Should be: {} is: {}",
    pkg.id.name, package_version, current_version
));
```

This will fix it by changing the code to:
```rs
ws.config().ui().error(format!(
    "Package {}. Required Cairo version isn't compatible with current version. Should be: {} is: {}",
    pkg.id.name, current_version, package_version
));
```

### DIFF
--- a/scarb/src/ops/resolve.rs
+++ b/scarb/src/ops/resolve.rs
@@ -364,7 +364,7 @@ fn check_cairo_version_compatibility(packages: &[Package], ws: &Workspace<'_>) -
             Some(package_version) if !package_version.matches(&current_version) => {
                 ws.config().ui().error(format!(
                     "Package {}. Required Cairo version isn't compatible with current version. Should be: {} is: {}",
-                    pkg.id.name, package_version, current_version
+                    pkg.id.name, current_version, package_version
                 ));
                 false
             }


maciektr:
Hi @BerzanOrg !
Thanks for input! 

I do agree that the message is not clear, but I am not sure about the change suggested 🤔 
`Should be: 2.4.0 is: ^4.3.5` Would suggest changing the `cairo-version` requirement, but what actually should be changed is the compiler used. 
Maybe we should reword to make the message more clear? (e.g. `The current Cairo version isn't compatible with the required Cairo version. The Cairo version should be: ^4.3.5 is: 2.4.0` or something similar)

BerzanOrg:
Yeah, my bad. I think the best thing to do is to make the error message more clear.
Maybe something like below:
```
Scarb's Cairo version isn't compatible with the project's required Cairo version. 
Cairo version required by the project: ^4.3.5
Cairo version of Scarb: 2.4.0
```

maciektr:
Sound great to me!

maciektr:
superseded with https://github.com/software-mansion/scarb/pull/1009

PR #1007, opened 2023-12-19
Add ignored test case for conflicting deps in unrelated ws members

maciektr:
**Stack**:
- #1007 ⬅
- #1030

### DIFF
--- a/scarb/tests/metadata.rs
+++ b/scarb/tests/metadata.rs
@@ -646,6 +646,69 @@ fn workspace_with_root() {
 }
 
 #[test]
+#[ignore = "not implemented yet"]
+fn workspace_with_conflicting_dep() {
+    // TODO(#2): Allow using conflicting dependencies in unrelated workspace members.
+    let t = assert_fs::TempDir::new().unwrap().child("test_workspace");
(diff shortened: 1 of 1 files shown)


szymmis:
What are those tests related to, especially in the context that they are ignored?

maciektr:
@szymmis basically bugs that we should sort out eventually

mkaput (commented):
I do understand the reasoning. I am afraid how this will play out with PubGrub, as I believe the canonical algorithm will attempt to unify `some_dep` in this scenario.

Did some user request this behaviour? Do we need to commit to it?

maciektr:
> Did some user request this behaviour? Do we need to commit to it?

Nope, just a thought experiment

mkaput:
> Though it seems that having same dep with different versions may actually make sense even in cairo - when the graph is disconnected.

That's a highly philosophical argument. One could want instead to use a single version of `dep` throughout the workspace because it's well -- a workspace. Given a workspace of multiple contracts, I'm not sure the author will want the contract A to have dependency v0, while contract B -- v1.

I think we have to ask users on Telegram.

PR #1011, opened 2023-12-22
Add support for compiler experimental features

tomek0123456789:
**Stack**:
- #1012
- #1011 ⬅

### DIFF
--- a/scarb-metadata/src/lib.rs
+++ b/scarb-metadata/src/lib.rs
@@ -231,6 +231,9 @@ pub struct PackageMetadata {
     #[serde(flatten)]
     pub manifest_metadata: ManifestMetadata,
 
+    /// Compiler experimental features allowed for this package.
+    pub experimental_features: Vec<String>,
+
     /// Additional data not captured by deserializer.
     #[cfg_attr(feature = "builder", builder(default))]
     #[serde(flatten)]
--- a/scarb/src/compiler/db.rs
+++ b/scarb/src/compiler/db.rs
@@ -93,16 +93,21 @@ fn build_project_config(unit: &CompilationUnit) -> Result<ProjectConfig> {
         .components
         .iter()
         .map(|component| {
+            let experimental_features = component.package.manifest.experimental_features.clone();
             (
                 component.cairo_package_name(),
                 CrateSettings {
                     edition: component.package.manifest.edition,
-                    experimental_features: Default::default(),
+                    // TODO (#1040): replace this with a macro
+                    experimental_features: cairo_lang_filesystem::db::ExperimentalFeaturesConfig {
+                        negative_impls: experimental_features
+                            .unwrap_or_default()
+                            .contains(&SmolStr::new_inline("negative_impls")),
+                    },
                 },
             )
         })
         .collect();
-
     let crates_config = AllCratesConfig {
         override_map: crates_config,
         ..Default::default()
--- a/scarb/src/core/manifest/mod.rs
+++ b/scarb/src/core/manifest/mod.rs
@@ -49,6 +49,9 @@ pub struct Manifest {
     pub compiler_config: ManifestCompilerConfig,
     #[builder(default)]
     pub scripts: BTreeMap<SmolStr, ScriptDefinition>,
+    /// Allow experimental features.
+    #[builder(default)]
+    pub experimental_features: Option<Vec<SmolStr>>,
 }
 
 /// Subset of a [`Manifest`] that contains package metadata.
--- a/scarb/src/core/manifest/toml_manifest.rs
+++ b/scarb/src/core/manifest/toml_manifest.rs
@@ -198,6 +198,7 @@ pub struct TomlPackage {
     /// **UNSTABLE** This package does not depend on Cairo's `core`.
     pub no_core: Option<bool>,
     pub cairo_version: Option<MaybeWorkspaceField<VersionReq>>,
+    pub experimental_features: Option<Vec<SmolStr>>,
 }
 
 #[derive(Clone, Debug, Serialize, Eq, PartialEq)]
@@ -556,6 +557,8 @@ impl TomlManifest {
             .transpose()?
             .unwrap_or_default();
 
+        // TODO (#1040): add checking for fields that are not present in ExperimentalFeaturesConfig
+        let experimental_features = package.experimental_features.clone();
         let manifest = ManifestBuilder::default()
             .summary(summary)
             .targets(targets)
@@ -563,8 +566,8 @@ impl TomlManifest {
             .metadata(metadata)
             .compiler_config(compiler_config)
             .scripts(scripts)
+            .experimental_features(experimental_features)
             .build()?;
-
         Ok(manifest)
     }
 
--- a/scarb/src/ops/metadata.rs
+++ b/scarb/src/ops/metadata.rs
@@ -135,6 +135,15 @@ fn collect_package_metadata(package: &Package) -> m::PackageMetadata {
 
     let edition = edition_variant(package.manifest.edition);
 
+    let experimental_features: Vec<String> = package
+        .manifest
+        .experimental_features
+        .clone()
+        .unwrap_or_default()
+        .iter()
+        .map(|x| x.to_string())
+        .collect();
+
     m::PackageMetadataBuilder::default()
         .id(wrap_package_id(package.id))
         .name(package.id.name.clone())
@@ -146,6 +155,7 @@ fn collect_package_metadata(package: &Package) -> m::PackageMetadata {
         .dependencies(dependencies)
         .targets(targets)
         .manifest_metadata(manifest_metadata)
+        .experimental_features(experimental_features)
         .build()
         .unwrap()
 }
--- a/utils/scarb-test-support/src/project_builder.rs
+++ b/utils/scarb-test-support/src/project_builder.rs
@@ -24,6 +24,7 @@ pub struct ProjectBuilder {
     src: HashMap<Utf8PathBuf, String>,
     deps: Vec<(String, Value)>,
     dev_deps: Vec<(String, Value)>,
+    manifest_package_extra: String,
     manifest_extra: String,
 }
 
@@ -44,6 +45,7 @@ impl ProjectBuilder {
             )]),
             deps: Vec::new(),
             dev_deps: Vec::new(),
+            manifest_package_extra: String::new(),
             manifest_extra: String::new(),
         }
     }
@@ -95,6 +97,11 @@ impl ProjectBuilder {
         self.dep("starknet", Dep.version(CAIRO_VERSION))
     }
 
+    pub fn manifest_package_extra(mut self, extra: impl ToString) -> Self {
+        self.manifest_package_extra = extra.to_string();
+        self
+    }
+
     pub fn manifest_extra(mut self, extra: impl ToString) -> Self {
         self.manifest_extra = extra.to_string();
         self
@@ -111,6 +118,12 @@ impl ProjectBuilder {
         if let Some(cairo_version) = self.cairo_version.as_ref() {
             doc["package"]["cairo-version"] = Item::Value(Value::from(cairo_version.to_string()));
         }
+        let mut manifest = doc.to_string();
+        if !self.manifest_package_extra.is_empty() {
+            manifest.push_str(&self.manifest_package_extra);
+        }
+
+        let mut doc = manifest.parse::<Document>().unwrap();
         doc["dependencies"] = toml_edit::table();
         for (name, dep) in &self.deps {
             doc["dependencies"][name.clone()] = Item::Value(dep.clone());
--- a/website/docs/reference/manifest.md
+++ b/website/docs/reference/manifest.md
@@ -209,6 +209,15 @@ Keys are human-readable link names, and values are URLs.
 "We're hiring" = "https://swmansion.com/careers/"
 ```
 
+### `experimental-features`
+
+This field is responsible for setting experimental flags to be used on the package for the compiler.
+
+```toml
+[package]
+experimental-features = ["negative_impls"]
+```
+
 ## `[dependencies]`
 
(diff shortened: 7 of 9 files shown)


scarb/src/core/manifest/mod.rs
```diff
@@ -67,6 +67,8 @@ pub struct ManifestMetadata {
     #[serde(rename = "tool")]
     pub tool_metadata: Option<BTreeMap<SmolStr, Value>>,
     pub cairo_version: Option<VersionReq>,
+    /// Allow experimental features.
+    pub experimental: Option<ScarbExperimentalFeaturesConfig>,
```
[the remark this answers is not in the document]
maciektr: Why `Option` here? This has defaults that would be applied later on anyway - we can `unwrap_or_default` when deserializing.

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -198,6 +200,13 @@ pub struct TomlPackage {
...
     pub no_core: Option<bool>,
     pub cairo_version: Option<MaybeWorkspaceField<VersionReq>>,
+    pub experimental: Option<MaybeWorkspaceField<ScarbExperimentalFeaturesConfig>>,
+}
+
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct ScarbExperimentalFeaturesConfig {
+    pub negative_impls: Option<bool>,
 }
```
maciektr, line 210: My suggestion would be:
- Implement `From` trait for both `ScarbExperimentalFeaturesConfig` -> `ExperimentalFeaturesConfig` conversion and vice versa. 
- Use `ScarbExperimentalFeaturesConfig {  }` when converting to make compiler ensure we use all fields.
- Use `ScarbExperimentalFeaturesConfig ` in the `toml` structs and  `ExperimentalFeaturesConfig` everywhere else.
- Use `Into::into` to convert where needed.
- Hack: you can create second struct called `ExperimentalFeaturesConfig` (instead of calling it `ScarbExperimentalFeaturesConfig`) by enclosing it in some `mod { }`. E.g. 
```
mod serde_ext {
    struct ExperimentalFeaturesConfig {}
}
``` 
and use it by `serde_ext::ExperimentalFeaturesConfig`.

scarb-metadata/src/lib.rs
```diff
@@ -231,6 +232,9 @@ pub struct PackageMetadata {
     #[serde(flatten)]
     pub manifest_metadata: ManifestMetadata,
 
+    /// Compiler experimental features allowed for this package.
+    pub experimental_features: ExperimentalFeaturesConfig,
```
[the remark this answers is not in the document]
mkaput: Agree, this will also avoid a risky `cairo-lang-*` dependency in this crate.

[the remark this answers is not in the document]
mkaput: Also, you must use `Option<_>` here because only Scarbs since a particular release will emit this field, and this crate must have both forward and backward compatibility.

mkaput (changes requested):
Have we discussed this change with Cairo team? Do they want us to make this available to end users?

maciektr:
> Have we discussed this change with Cairo team? Do they want us to make this available to end users?

@mkaput this has been directly requested by the compiler team

scarb/src/core/manifest/mod.rs
mkaput, line 71:
```suggestion
    pub experiments: Option<ScarbExperimentalFeaturesConfig>,
```

? noun reads better here than an adjective imo

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -198,6 +200,13 @@ pub struct TomlPackage {
     /// **UNSTABLE** This package does not depend on Cairo's `core`.
     pub no_core: Option<bool>,
     pub cairo_version: Option<MaybeWorkspaceField<VersionReq>>,
+    pub experimental: Option<MaybeWorkspaceField<ScarbExperimentalFeaturesConfig>>,
```
mkaput, line 203: Suggestion: what do you think about the following alternative interface:

```toml
[package]
name = "foobar"
experiments = ["negative_impls"]
```

1. It's much more concise.
2. I do not expect to have to pass any parameters here, actually. Like, we are turning on/off flags here only.

website/docs/reference/workspaces.md
```diff
@@ -100,6 +100,7 @@ Keys that are supported:
 - `readme`
 - `repository`
 - `cairo-version`
+- `experimental`
```
mkaput, line 103: I would rather **not** make it easy for users to enable experimental features for entire workspace. We should expect users to avoid using them in production code 99% of cases. Especially, I'd prohibit publishing packages with experimental features turned on.

mkaput (changes requested):

scarb/src/core/manifest/toml_manifest.rs
tomek0123456789: Changed that to match your suggestion, but I still don't like the fact that with each new compiler flag we'll need to add support for it instead of doing it automatically - or maybe I'm missing something.

scarb/src/core/manifest/mod.rs
tomek0123456789: Would `experimental_features` be too long?

scarb/src/core/manifest/toml_manifest.rs
tomek0123456789: Changed implementation to match [Marek's suggestion](https://github.com/software-mansion/scarb/pull/1011#discussion_r1440251944)

scarb/src/core/manifest/mod.rs
mkaput: Kind of? To my understanding, the word `experiments` means basically the same thing, and is more catchy.

On the other hand, if the compiler is going to produce any diagnostics with the term _experimental features_, then it doesn't make sense to change this on our side :/

scarb/src/core/manifest/toml_manifest.rs
mkaput: We can write unit tests that will ensure everything is covered -- this should reduce the pain of maintaining this.

scarb-metadata/src/lib.rs
```diff
@@ -231,6 +231,9 @@ pub struct PackageMetadata {
     #[serde(flatten)]
     pub manifest_metadata: ManifestMetadata,
 
+    /// Compiler experimental features allowed for this package.
+    pub experimental_features: Option<serde_json::Value>,
```
mkaput, line 235:
```suggestion
    pub experimental_features: Option<Vec<String>>,
```

I used `Vec` here instead of `HashSet`, to preserve ordering.

scarb/src/ops/metadata.rs
```diff
@@ -135,6 +137,11 @@ fn collect_package_metadata(package: &Package) -> m::PackageMetadata {
 
     let edition = edition_variant(package.manifest.edition);
 
+    let experiments = package.manifest.experiments.clone().unwrap_or_default();
+    let experimental_features = json!(ExperimentalFeaturesConfig {
+        negative_impls: experiments.contains("negative_impls"),
+    });
+
```
mkaput, line 144: no need to do that here, we should just pass-through what's contained within Scarb.toml

mkaput (changes requested):

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -198,6 +198,13 @@ pub struct TomlPackage {
     /// **UNSTABLE** This package does not depend on Cairo's `core`.
     pub no_core: Option<bool>,
     pub cairo_version: Option<MaybeWorkspaceField<VersionReq>>,
+    pub experiments: Option<HashSet<String>>,
```
mkaput, line 201:
```suggestion
    pub experiments: HashSet<String>,
```

I don't think we'll have to differentiate _none_ from _not specified_, so you can skip the `Option` as well, and use an empty `HashSet` to represent nothing.

FYI empty Vec/HashSet/HashMap etc. do not allocate and are basically free perf wise

utils/scarb-test-support/src/project_builder.rs
```diff
@@ -24,6 +24,7 @@ pub struct ProjectBuilder {
     src: HashMap<Utf8PathBuf, String>,
     deps: Vec<(String, Value)>,
     dev_deps: Vec<(String, Value)>,
+    experiments: Vec<String>,
```
mkaput, line 27: This is such an esoteric field, I'd rather skip adding extra logic for it. We do not want this helper to grow infinitely. You can always call `manifest_extra` manually in tests needing it.

website/docs/reference/manifest.md
```diff
@@ -209,6 +209,15 @@ Keys are human-readable link names, and values are URLs.
...
 
+### `experiments`
+
+This field is responsible for setting experimental flags to be used on the package for the compiler.
+
+```toml
+[package]
+experiments = ["negative_impls"]
+```
+
```
mkaput, line 220: Please create an issue to include a list of all supported experimental features in our documentation. I'd like this list to be auto-generated from code, so that won't be trivial so let's do this as separate chunk of work

scarb/src/core/manifest/mod.rs
```diff
@@ -49,6 +49,9 @@ pub struct Manifest {
     pub compiler_config: ManifestCompilerConfig,
     #[builder(default)]
     pub scripts: BTreeMap<SmolStr, ScriptDefinition>,
+    /// Allow experimental features.
+    #[builder(default)]
+    pub experiments: Option<HashSet<String>>,
```
mkaput, line 54:
```suggestion
    pub experiments: Option<HashSet<SmolStr>>,
```

My intuition says that `SmolStr` would be a better choice here, because it avoids allocations and has cheap clones.

scarb/src/compiler/db.rs
```diff
@@ -93,16 +93,23 @@ fn build_project_config(unit: &CompilationUnit) -> Result<ProjectConfig> {
...
+                .clone()
+                .unwrap_or_default();
             (
                 component.cairo_package_name(),
                 CrateSettings {
                     edition: component.package.manifest.edition,
-                    experimental_features: Default::default(),
+                    experimental_features: cairo_lang_filesystem::db::ExperimentalFeaturesConfig {
+                        negative_impls: experiments.contains("negative_impls"),
+                    },
```
mkaput, line 108: I have an idea how to ensure in compile-time that the list of experiments is exhaustive:

Write a macro ([`macro_rules!`](https://blog.logrocket.com/macros-in-rust-a-tutorial-with-examples/)), which usage would look like this:

```rust
experimental_features!(ExperimentalFeatues, [negative_impls,something_else,etc])
```

which would expand to something like this:

```rust
// Struct name is the first argument to the macro.
struct ExperimentalFeatures(HashSet<SmolStr>);

impl ExperimentalFeatures {
    fn new_checked(values: HashSet<SmolStr>, package: PackageId, ui: &Ui) -> Self {
        for feature in &values {
             match feature {
                 "negative_impls" => {},
                 "something_else" => {},
                 "etc" => {},
                 _ => ui.warn(&format!("{package}: unknown experimental feature: {feature}"))
             }
        }
        Self(values)
    }

    fn to_compiler_features(&self) -> cairo_lang_filesystem::db::ExperimentalFeaturesConfig {
        // Rust compiler will check exhaustiveness here
        cairo_lang_filesystem::db::ExperimentalFeaturesConfig {
            negative_impls: self.0.contains("negative_impls"),
        }
    }
}
```

---

this can be done in separate PR; please create an issue for this

mkaput (changes requested):

scarb-metadata/src/lib.rs
tomek0123456789: So, do we want all of the fields (in `TomlManifest`, `Manifest` and `PackageMetadata`) to be vectors?

scarb/src/compiler/db.rs
tomek0123456789: Issue #1040
This should be also done in `snforge-test-collector` [here](https://github.com/software-mansion/scarb/blob/8733f15164d767a4581a0be1835bd3bd0516244d/extensions/scarb-snforge-test-collector/src/metadata.rs#L92), right?

website/docs/reference/manifest.md
tomek0123456789: #1041

scarb-metadata/src/lib.rs
mkaput: discussed offline

utils/scarb-test-support/src/project_builder.rs
mkaput: hmm right. Perhaps it'd be better to have `manifest_package_extra` method?

scarb/src/compiler/db.rs
mkaput: yes, though I don't like this fact

utils/scarb-test-support/src/project_builder.rs
tomek0123456789: done, added that

scarb/src/core/manifest/toml_manifest.rs
tomek0123456789: changed that, all of those are now `Option`s

mkaput (commented):
Looks great so far!

scarb/src/core/manifest/toml_manifest.rs
tomek0123456789: Added info to the #1040 issue about the test and raising a warning

mkaput (approved):

tomek0123456789:
✓ Commit merged in pull request [#1012](https://github.com/software-mansion/scarb/pull/1012)

PR #1012, opened 2023-12-22
Add support for compiler experimental features in `snforge-test-collector`

tomek0123456789:


extensions/scarb-snforge-test-collector/src/metadata.rs
```diff
@@ -93,7 +93,14 @@ impl CompilationUnit<'_> {
                         } else {
                             Edition::default()
                         },
-                        experimental_features: Default::default(),
+                        experimental_features: self
+                            .metadata
+                            .get_package(&component.package)
+                            .unwrap_or_else(|| {
+                                panic!("Failed to find = {} package", component.package)
+                            })
```
maciektr, line 101: Since this pattern is used more than once - maybe helper? (e.g. `fetch_package`)

tomek0123456789: Changed `get_package` to return `&PackageMetadata` with the `unwrap_or_else` call since the only other usage of this method was the same, but with a slightly differently phrased error in `panic!`.

scarb-metadata/src/lib.rs
```diff
@@ -470,8 +470,11 @@ pub enum Cfg {
 
 impl Metadata {
     /// Returns reference to [`PackageMetadata`] corresponding to the [`PackageId`].
-    pub fn get_package(&self, id: &PackageId) -> Option<&PackageMetadata> {
-        self.packages.iter().find(|p| p.id == *id)
+    pub fn get_package(&self, id: &PackageId) -> &PackageMetadata {
+        self.packages
+            .iter()
+            .find(|p| p.id == *id)
+            .unwrap_or_else(|| panic!("Failed to find = {} package", id))
```
mkaput, line 477: this is an API breaking change, please revert this

mkaput (changes requested):

mkaput (approved):

PR #1033, opened 2024-01-04
Use dev-deps during compilation in `snforge-test-collector`

piotmag769:
Closes https://github.com/foundry-rs/starknet-foundry/issues/1460

The solution is to fetch dependencies from unit test compilation unit.

After talking with @mkaput we decided that to ensure this heuristic solution doesn't break we check if all other test compilation units have the same dependencies (excluding dependencies being sources for integration tests i. e., starting with `{package_path}/tests/`).

### DIFF
--- a/extensions/scarb-snforge-test-collector/src/metadata.rs
+++ b/extensions/scarb-snforge-test-collector/src/metadata.rs
@@ -1,9 +1,11 @@
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, ensure, Context, Result};
 use cairo_lang_filesystem::db::{CrateSettings, Edition};
 use cairo_lang_project::AllCratesConfig;
 use cairo_lang_utils::ordered_hash_map::OrderedHashMap;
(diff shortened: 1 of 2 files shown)


extensions/scarb-snforge-test-collector/src/metadata.rs
```diff
@@ -18,23 +20,45 @@ pub fn compilation_unit_for_package<'a>(
...
-            name @ "starknet-contract" => (0, name),
-            name @ "lib" => (1, name),
-            name => (2, name),
+        .find(|unit| {
+            unit.package == package_metadata.id
+                && unit.target.kind == "test"
+                && unit.target.params
+                    == json!({
+                      "test-type": "unit"
+                    })
```
maciektr, line 32: Pls try getting this type by key.

maciektr (commented):
Looks good, one nit comment

Arcticae (approved):

maciektr (approved):

PR #1046, opened 2024-01-11
Add `scarb check`

### LINKED ISSUE #137: Investigate add `scarb check`

tomek0123456789:
Closes #137


mkaput (commented):
Yeah, the overall implementation idea is correct here. If this doesn't bring any improvements in other Cairo codebases, then let's abandon this feature.

mkaput:
Note: In the future I'd like to move `RootDatabase` creation and diagnostics checking outside `Compiler` implementation responsibility, but as for now this looks good.

scarb/src/compiler/compilers/lib.rs
mkaput, line 1: Hmm, I have an idea for alternative implementation. What if, instead of reusing the `compile` op with `CompileMode` enum, we create a separate `ops::check`, that is building compiler database on its own, skipping the `Compiler` trait altogether?

mkaput (commented):
I had a look at these numbers once again, and I actually think there might be a viable path to go here. Have a look at my alternative implementation detail comment.

scarb/src/compiler/compilers/lib.rs
mkaput: Especially as in the scenario of testing multiple targets at once (which should be the default behaviour), I think we could reuse single database instance and just reconfigure it for each target in a sequential loop.

scarb/tests/build_starknet_contract_allowed_libfuncs.rs
```diff
@@ -103,13 +105,13 @@ fn check_false() {
...
+        .arg("check")
         .current_dir(&t)
         .assert()
         .success()
         .stdout_matches(indoc! {r#"
-        [..] Compiling hello v0.1.0 ([..])
-        [..]  Finished release target(s) in [..]
+        [..]  Checking hello v0.1.0 ([..])
+        [..]  Finished checking release target(s) in [..]
         "#});
```
mkaput, line 115: If `check` does not check this fact, that changing `build` to `check` here makes the test misbehaving, as we want to ensure that the warning does not pop up here.

mkaput (changes requested):

mkaput (commented):
This looks excellent code-wise! Wdyt about re-running benchmarks at this stage?

tomek0123456789:
[wip]

---------------------

benchmark on [keep-starknet-strange/alexandria](https://github.com/keep-starknet-strange/alexandria) (10 runs):
```
$ hyperfine --show-output --warmup 3  '../../scarb/target/release/scarb check -q' '../../scarb/target/release/scarb build -q'       

Benchmark 1: ../../scarb/target/release/scarb check -q
  Time (mean ± σ):      8.427 s ±  0.161 s    [User: 7.930 s, System: 0.378 s]
  Range (min … max):    8.176 s …  8.625 s    10 runs
 
Benchmark 2: ../../scarb/target/release/scarb build -q
  Time (mean ± σ):      8.707 s ±  0.184 s    [User: 8.181 s, System: 0.402 s]
  Range (min … max):    8.495 s …  9.043 s    10 runs
 
Summary
  ../../scarb/target/release/scarb check -q ran
    1.03 ± 0.03 times faster than ../../scarb/target/release/scarb build -q
```
with `--test` flag:
```
$ hyperfine --show-output --warmup 3  '../../scarb/target/release/scarb check -q --test' '../../scarb/target/release/scarb build -q --test'

Benchmark 1: ../../scarb/target/release/scarb check -q --test
  Time (mean ± σ):     22.622 s ±  0.357 s    [User: 21.250 s, System: 1.021 s]
  Range (min … max):   22.019 s … 23.273 s    10 runs
 
Benchmark 2: ../../scarb/target/release/scarb build -q --test
  Time (mean ± σ):     25.854 s ±  1.548 s    [User: 23.762 s, System: 1.220 s]
  Range (min … max):   24.623 s … 29.331 s    10 runs
 
Summary
  ../../scarb/target/release/scarb check -q --test ran
    1.14 ± 0.07 times faster than ../../scarb/target/release/scarb build -q --test
```
with `lib.sierra-text` target:
```
$ hyperfine --show-output --warmup 3  '../../scarb/target/release/scarb check -q' '../../scarb/target/release/scarb build -q'

Benchmark 1: ../../scarb/target/release/scarb check -q
  Time (mean ± σ):      8.097 s ±  0.107 s    [User: 7.605 s, System: 0.351 s]
  Range (min … max):    7.990 s …  8.294 s    10 runs
 
Benchmark 2: ../../scarb/target/release/scarb build -q
  Time (mean ± σ):      8.524 s ±  0.141 s    [User: 8.004 s, System: 0.380 s]
  Range (min … max):    8.306 s …  8.761 s    10 runs
 
Summary
  ../../scarb/target/release/scarb check -q ran
    1.05 ± 0.02 times faster than ../../scarb/target/release/scarb build -q
```

scarb/src/bin/scarb/args.rs
```diff
@@ -166,6 +166,8 @@ pub enum Command {
     /// Manipulate packages cache.
     #[clap(subcommand)]
     Cache(CacheSubcommand),
+    /// Compile until diagnostics checking.
```
maciektr, line 169: The `diagnostics checking` sounds kind of cryptic to me. Note, that this will be shown to users on `--help`. Maybe we could use something more similar to what Cargo does?
```
    check, c    Analyze the current package and report errors, but don't build object files
```

mkaput (approved):
This looks great! LGTM from my side, but let's wait for @maciektr review

maciektr (approved):
Looks great!

PR #1057, opened 2024-01-17
Fix `snforge-test-collector`'s default `available_gas` value

Arcticae:
This needs to be `None` because older versions of snforge would crash on `u32::max` (which seems to be a default value), with an error that the attribute is not supported.

### DIFF
--- a/extensions/scarb-snforge-test-collector/tests/test.rs
+++ b/extensions/scarb-snforge-test-collector/tests/test.rs
@@ -48,10 +48,7 @@ fn forge_test_locations() {
     assert_eq!(&json[1]["test_cases"][0]["name"], "tests::tests::test");
     assert_eq!(&json[1]["tests_location"], "Tests");
 
-    assert_eq!(
-        &json[0]["test_cases"][0]["available_gas"],
-        &Value::Null
-    );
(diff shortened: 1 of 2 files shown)


extensions/scarb-snforge-test-collector/tests/test.rs
```diff
@@ -48,10 +48,7 @@ fn forge_test_locations() {
     assert_eq!(&json[1]["test_cases"][0]["name"], "tests::tests::test");
     assert_eq!(&json[1]["tests_location"], "Tests");
 
-    assert_eq!(
-        &json[0]["test_cases"][0]["available_gas"],
-        &Value::Number(Number::from(u32::MAX))
-    );
+    assert_eq!(&json[0]["test_cases"][0]["available_gas"], &Value::Null);
```
war-in, line 51: Would be nice to have a test when `available_gas` is set to the highest possible value

Arcticae: It's out of scope for this PR IMO

maciektr (approved):

war-in (approved):

piotmag769:
🎉

PR #1060, opened 2024-01-17
Add `ProcMacroHost` definition

maciektr:
commit-id:18967d01

---

**Stack**:
- #1143
- #1148
- #1100
- #1110
- #1093
- #1091
- #1060 ⬅

### DIFF
--- a/scarb/src/compiler/plugin/proc_macro/host.rs
+++ b/scarb/src/compiler/plugin/proc_macro/host.rs
@@ -1,0 +1,69 @@
+use crate::compiler::plugin::proc_macro::ProcMacroInstance;
+use crate::core::Package;
+use anyhow::Result;
+use cairo_lang_defs::plugin::{MacroPlugin, MacroPluginMetadata, PluginResult};
+use cairo_lang_semantic::plugin::PluginSuite;
+use cairo_lang_syntax::node::ast::ModuleItem;
+use cairo_lang_syntax::node::db::SyntaxGroup;
+use std::sync::Arc;
+
+/// A Cairo compiler plugin controlling the procedural macro execution.
+///
+/// This plugin decides which macro plugins (if any) should be applied to the processed AST item.
+/// It then redirects the item to the appropriate macro plugin for code expansion.
+#[derive(Debug)]
+pub struct ProcMacroHostPlugin {
+    macros: Vec<Arc<ProcMacroInstance>>,
+}
+
+impl ProcMacroHostPlugin {
+    pub fn new(macros: Vec<Arc<ProcMacroInstance>>) -> Self {
+        Self { macros }
+    }
+}
+
+impl MacroPlugin for ProcMacroHostPlugin {
+    fn generate_code(
+        &self,
+        _db: &dyn SyntaxGroup,
+        _item_ast: ModuleItem,
+        _metadata: &MacroPluginMetadata<'_>,
+    ) -> PluginResult {
+        // Apply expansion to `item_ast` where needed.
+        // TODO(maciektr): Implement
+        PluginResult::default()
+    }
+
+    fn declared_attributes(&self) -> Vec<String> {
+        self.macros
+            .iter()
+            .flat_map(|m| m.declared_attributes())
+            .collect()
+    }
+}
+
+/// A Scarb wrapper around the `ProcMacroHost` compiler plugin.
+///
+/// This struct represent the compiler plugin in terms of Scarb data model.
+/// It also builds a plugin suite that enables the compiler plugin.
+#[derive(Default)]
+pub struct ProcMacroHost {
+    macros: Vec<Arc<ProcMacroInstance>>,
+}
+
+impl ProcMacroHost {
+    pub fn register(&mut self, package: Package) -> Result<()> {
+        // Create instance
+        // Register instance in hash map
+        let instance = ProcMacroInstance::try_new(package)?;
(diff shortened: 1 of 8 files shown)


plugins/cairo-lang-macro-interface/Cargo.toml
```diff
@@ -14,6 +14,7 @@ repository.workspace = true
 
 [dependencies]
 anyhow.workspace = true
+libc.workspace = true
```
mkaput, line 17: unused?

scarb/src/core/package/name.rs
```diff
@@ -30,6 +30,7 @@ impl PackageName {
     pub const CORE: Self = PackageName(SmolStr::new_inline(CORELIB_CRATE_NAME));
     pub const STARKNET: Self = PackageName(SmolStr::new_inline("starknet"));
     pub const TEST_PLUGIN: Self = PackageName(SmolStr::new_inline("test_plugin"));
+    pub const PROC_MACRO_HOST: Self = PackageName(SmolStr::new_inline("proc_macro_host"));
```
mkaput, line 33: unused?

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -0,0 +1,64 @@
...
+use typed_builder::TypedBuilder;
+
+/// A Cairo compiler plugin controlling the procedural macro execution.
+///
+/// This plugin decides which macro plugins (if any) should be applied to the processed AST item.
+/// It then redirects the item to the appropriate macro plugin for code expansion.
+#[derive(Debug, TypedBuilder)]
+pub struct ProcMacroHost {
+    macros: HashMap<SmolStr, Arc<ProcMacroInstance>>,
+}
```
mkaput, line 20: Are you going to add a lot of fields later on to this structure? Because otherwise I suggest doing the classical `new` constructor method. If you can make stuff simpler then why not? TypedBuilder also will probably slow down compilation time and code size.

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+/// A Scarb wrapper around the `ProcMacroHost` compiler plugin.
+///
+/// This struct represent the compiler plugin in terms of Scarb data model.
+/// It also builds a plugin suite that enables the compiler plugin.
+#[derive(Default)]
+pub struct ProcMacroHostPlugin {
+    macros: HashMap<SmolStr, Arc<ProcMacroInstance>>,
```
mkaput, line 45: why hashmap? why not just a vector of instances?

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -0,0 +1,64 @@
...
+        self.macros.keys().map(|name| name.to_string()).collect()
+    }
+}
+
+/// A Scarb wrapper around the `ProcMacroHost` compiler plugin.
+///
+/// This struct represent the compiler plugin in terms of Scarb data model.
+/// It also builds a plugin suite that enables the compiler plugin.
+#[derive(Default)]
+pub struct ProcMacroHostPlugin {
```
mkaput, line 44: Shouldn't these two types have names swapped? `ProcMacroHost` is the actual compiler plugin...

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -0,0 +1,64 @@
...
+        _item_ast: ModuleItem,
+        _metadata: &MacroPluginMetadata<'_>,
+    ) -> PluginResult {
+        // Apply expansion to `item_ast` where needed.
+        // TODO(maciektr): Implement
+        PluginResult::default()
+    }
+
+    fn declared_attributes(&self) -> Vec<String> {
+        self.macros.keys().map(|name| name.to_string()).collect()
```
mkaput, line 35: A single instance will be capable of providing multiple macros, so I think you'd better be with having `declared_attributees` method in the instance; instead of binding it to name

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -0,0 +1,64 @@
...
+        self.macros.insert(name, Arc::new(instance));
+        Ok(())
+    }
+
+    pub fn plugin_suite(&self) -> PluginSuite {
+        let macro_host = ProcMacroHost::builder().macros(self.macros.clone()).build();
+        let mut suite = PluginSuite::default();
+        suite.add_plugin_ex(Arc::new(macro_host));
+        suite
+    }
```
mkaput, line 63:
```suggestion
    pub fn into_plugin_suite(self) -> PluginSuite {
        let macro_host = ProcMacroHost::builder().macros(self.macros.clone()).build();
        let mut suite = PluginSuite::default();
        suite.add_plugin_ex(Arc::new(macro_host));
        suite
    }
```

? no need to think about cloning macros

mkaput (changes requested):

scarb/src/compiler/plugin/proc_macro/host.rs
maciektr: Yep, I just assumed we would change it later, as we have deprioritized the multiple macros per package feature :D  
Will add it now

mkaput (approved):

PR #1085, opened 2024-01-25
Expose per-compilation-unit cfg sets in scarb metadata

### LINKED ISSUE #1076: Expose per-crate cfg sets in scarb metadata
### Problem

Cairo 2.5.0 introduced per-crate cfg sets (https://github.com/starkware-libs/cairo/pull/4831). We need to include them in scarb-metadata output.

### Proposed Solution

Add a new field to `CompilationUnitComponentMetadata`:

```rust
#[serde(default)]
pub cfg: Vec<Cfg>
```

and populate it with cfg items to-be-attached to this unit.

The global per-unit cfg items set should get main component's cfg items, for backwards compatibility reason.

### Notes

1. Don't forget about extensions that are pulling information from.
2. Make `scarb-metadata` release after merging these changes.

szymmis:
Depends on #1084 
Resolves #1076

### DIFF
--- a/scarb-metadata/src/lib.rs
+++ b/scarb-metadata/src/lib.rs
@@ -350,6 +350,12 @@ pub struct CompilationUnitComponentMetadata {
     pub name: String,
     /// Path to the root Cairo source file.
     pub source_path: Utf8PathBuf,
+    /// CfgSet this component is compiled with
+    ///
+    /// This will differ from parent CompilationUnit's CfgSet
+    /// when compiling for test target
+    #[serde(default)]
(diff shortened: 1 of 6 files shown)


maciektr:
I feel like this ops/metadata and db helper functions are to smart. Why not store the cfg set on `CompilationUnitComponent` where applicable and keep all cfg resolution in the compilation unit generation step (similar to what has been done in `CompilationUnit`)?

szymmis:
Makes sense @maciektr, I'll look into it

PR #1088, opened 2024-01-25
Make `parse_stream()` and `scarb_command()` public

Draggu:
Make `parse_stream()` and `scarb_command()` public so raw command output can be captured for pretty printing

### DIFF
--- a/scarb-metadata/src/command/metadata_command.rs
+++ b/scarb-metadata/src/command/metadata_command.rs
@@ -126,7 +126,8 @@ impl MetadataCommand {
         self
     }
 
-    fn scarb_command(&self) -> Command {
+    /// Creates [`Command`] from this [`MetadataCommand`]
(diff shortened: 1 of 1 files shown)


mkaput:
Leaving review to the team. My question: why do you need this? What do you mean by this?

> so raw command output can be captured for pretty printing

My gut feelings are that this is not the path to achieve what you want. Don't you just want to execute `scarb fetch` and pipe its output instead?

Draggu:
> Leaving review to the team. My question: why do you need this? What do you mean by this?
> 
> > so raw command output can be captured for pretty printing
> 
> My gut feelings are that this is not the path to achieve what you want. Don't you just want to execute `scarb fetch` and pipe its output instead?

I need to add some additional logic to `exec()`. To avoid copy-pasting parsing logic, it would be awesome to reuse it from here.

PR #1091, opened 2024-01-29
Implement procedural macro host `generate_code` for attribute macro

### LINKED ISSUE #1127: Implement foundations of Procedural Macro Host
### Problem

like title says 

### Proposed Solution

_No response_

### Notes

_No response_

maciektr:
Resolves #1127

commit-id:e59370d2

---

**Stack**:
- #1159
- #1157
- #1143
- #1151
- #1150
- #1148
- #1100
- #1155
- #1110
- #1093
- #1091 ⬅

### DIFF
--- a/scarb/src/compiler/plugin/proc_macro/host.rs
+++ b/scarb/src/compiler/plugin/proc_macro/host.rs
@@ -1,10 +1,16 @@
-use crate::compiler::plugin::proc_macro::ProcMacroInstance;
-use crate::core::Package;
+use crate::compiler::plugin::proc_macro::{FromItemAst, ProcMacroInstance};
+use crate::core::{Package, PackageId};
 use anyhow::Result;
-use cairo_lang_defs::plugin::{MacroPlugin, MacroPluginMetadata, PluginResult};
+use cairo_lang_defs::plugin::{
+    MacroPlugin, MacroPluginMetadata, PluginGeneratedFile, PluginResult,
+};
+use cairo_lang_macro::{ProcMacroResult, TokenStream};
 use cairo_lang_semantic::plugin::PluginSuite;
-use cairo_lang_syntax::node::ast::ModuleItem;
+use cairo_lang_syntax::attribute::structured::AttributeListStructurize;
+use cairo_lang_syntax::node::ast;
 use cairo_lang_syntax::node::db::SyntaxGroup;
+use itertools::Itertools;
+use smol_str::SmolStr;
 use std::sync::Arc;
 
 /// A Cairo compiler plugin controlling the procedural macro execution.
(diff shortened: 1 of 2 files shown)


mkaput (approved):

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -16,22 +22,131 @@ pub struct ProcMacroHostPlugin {
...
+            .chain(self.handle_derive(db, item_ast.clone()));
+
+        let mut token_stream = TokenStream::from_item_ast(db, item_ast);
+        let mut modified = false;
+        for input in expansions {
+            let instance = self
+                .macros
+                .iter()
+                .find(|m| m.declared_attributes().contains(&input.id.to_string()))
+                .unwrap();
```
maciektr, line 120: Multiple macros per instance will be handled later on

piotmag769:
I know it's unrelated but it took me some time to realise why do we have `PackageId` in  `ProcMacroInstace` (that it is actually a package a macro should be applied to) - I think the comment to this struct or the name could be improved.

I also think the struct `ProcMacroHost` may be excessive, a single method for building a plugin suite added to `ProcMacroHostPlugin` would suffice imo. Cannot tell for sure though, maybe there is more code related to `ProcMacroHost` later on.

Those are just my personal opinions though, feel free to ignore them xD

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -16,22 +22,131 @@ pub struct ProcMacroHostPlugin {
...
+            }
+            ast::ModuleItem::FreeFunction(free_func_ast) => Some(free_func_ast.attributes(db)),
+            _ => None,
+        };
+
+        attrs
+            .map(|attrs| attrs.structurize(db))
+            .unwrap_or_default()
+            .iter()
+            .filter(|attr| self.declared_attributes().contains(&attr.id.to_string()))
```
piotmag769, line 80: Was looking at the code and I don't seem to get why `declared_attributes` for `ProcMacroInstance` return its package name - is it some simplification we want to do at early stages of implementation?

piotmag769 (commented):
I assume TODO's are done later in the PR stack, right?

I noticed that the `TokenStream` is a simple `String` rn, we want to change it to be an iterator over tokens later on, right?

After digging into the code - LGTM

piotmag769 (approved):

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -16,22 +22,131 @@ pub struct ProcMacroHostPlugin {
...
+            .map(|attr| ProcMacroInput {
+                id: attr.id.clone(),
+                kind: ProcMacroKind::Attribute,
```
piotmag769, line 83: Wondering if including the `ProcMacroInstance` associated with the proc macro input wouldn't be good - especially later on when we want to allow the same macro names in different packages - then the namespace resolution could be done here to not leak abstraction

scarb/src/compiler/plugin/proc_macro/host.rs
maciektr: Yes. The ProcMacroInstance represents a package (Scarb package with Rust source code), that implements the proc macro. For now, it is only possible to have one macro per in such package. Ultimately, it should be possible to define multiple macros in a single package.

scarb/src/compiler/plugin/proc_macro/host.rs
maciektr: @piotmag769  Yes, this stems from the one macro per package simplification. Ultimately, it should discover macros exposed by the shared library, and load all them with their names. Right now I've skipped that part, so I would not need to implement any so contents discovery yet.

PR #1098, opened 2024-01-29
Support for human readable message in `MetadataCommand`

Draggu:
Added support for human readable message in `MetadataCommand` by removing `--json` flag from `MetadataCommand` and adding `inherit_stdout()` method.


scarb-metadata/src/command/internal_command.rs
```diff
@@ -117,12 +116,6 @@ impl InternalScarbCommandBuilder {
         self
     }
 
-    /// Set output format to JSON.
-    pub fn json(&mut self) -> &mut Self {
-        self.json = true;
-        self
-    }
-
```
maciektr, line 125: Can we please add `json` flag to `MetadataCommand` and keep it here? This would allow more flexibility in the interface (which would be able to output either json or textual representation.

scarb-metadata/src/version_pin.rs
```diff
@@ -49,9 +49,23 @@ impl Serialize for VersionPin {
...
+                let err_message = err.to_string();
+
+                if err_message.contains("missing field") {
+                    err
+                } else {
+                    Error::custom(format!(
+                        "{}\n expected metadata version to be of type u64",
+                        err_message
+                    ))
+                }
```
maciektr, line 66: Why does it need to change?

scarb-metadata/src/command/metadata_command.rs
```diff
@@ -142,6 +150,9 @@ impl MetadataCommand {
         let mut cmd = self.scarb_command();
         let output = cmd.output()?;
         if !output.status.success() {
+            if self.inherit_stdout {
+                stdout().write_all(&output.stdout)?;
+            }
```
maciektr, line 155: Shouldn't we be able to pass stdout messages when the command is successful as well?

scarb-metadata/src/command/metadata_command.rs
```diff
@@ -153,33 +164,69 @@ impl MetadataCommand {
...
-        // field first. If one is spotted before the metadata message, deserialization of it should
-        // fail. Instead of immediately returning the error, we save it, and continue to look up
-        // for the metadata later in output. If it is found -- great, we will return it and forget
-        // about the error. Otherwise, we assume that the found message was meant actually to be
-        // the metadata object that failed to deserialize, and thus we report deserialization error
-        // instead of `NotFound`.
-        match serde_json::from_slice(&line) {
-            Ok(metadata) => return Ok(metadata),
-            Err(serde_err) => err = Some(serde_err.into()),
+    let data = std::str::from_utf8(data).unwrap();
```
maciektr, line 168: Can we avoid unwrap?

maciektr (commented):
Great work! 💪 
Please find some comments attached below.

.tool-versions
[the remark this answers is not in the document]
Draggu: Sure, my bad

scarb-metadata/src/version_pin.rs
Draggu: So it can be captured in `parse_stream` for returned error consistency

scarb-metadata/src/command/metadata_command.rs
Draggu: If command is successful we return it's output parsed as `Metadata` so I don't see the point in printing it.

scarb-metadata/src/command/metadata_command.rs
Draggu: With a little bit of allocation, yes. Fixed.

scarb-metadata/src/command/internal_command.rs
Draggu: After conversation with @mkaput we decided to remove it, but yes it can be kept.

scarb-metadata/src/command/metadata_command.rs
maciektr: Even if the command is successful, the metadata itself is just one of messages sent to stdout. The other may include resolver warning, or info like the one below. 
```
    Updating git repository https://github.com/keep-starknet-strange/alexandria
{
  "version": 1,
``` 
So imho it would be cool to forward them as well, for full functionality of stdout inheritance (except for the metadata one ofc).

scarb-metadata/src/command/internal_command.rs
maciektr: Since you have already added support for it in the parser, I think we should keep it. CC @mkaput

scarb-metadata/src/version_pin.rs
maciektr: Yes, I see that you capture it later, but why can't we simply return `MetadataCommandError::Json` regardless of the underlying deserialization error source?

Draggu: I implemented it that way for compatibility (see tests). I agree it would be simpler and should not matter. If you are sure it won't break anything i can change it.

scarb-metadata/src/command/metadata_command.rs
```diff
@@ -15,10 +15,7 @@ use crate::{Metadata, VersionPin};
 pub enum MetadataCommandError {
     /// `scarb metadata` command did not produce any metadata
     #[error("`scarb metadata` command did not produce any metadata")]
-    NotFound {
-        /// Captured standard output if any.
-        stdout: String,
-    },
+    NotFound,
```
maciektr, line 18: Can we keep that? Shouldn't be too hard I think (the last resort would be to return it from `exec`).

maciektr (commented):
Great! Last one thing to consider before we can merge it.

maciektr (approved):
LGTM! 💯

mkaput (approved):
Looks ok. I see that this code does buffer entire stdout while it shouldn't, but that's something Scarb team can built on top of this code.

PR #1099, opened 2024-01-30
`scarb check` docs

tomek0123456789:


website/docs/guides/working-on-an-existing-package.md
```diff
@@ -31,3 +31,20 @@ $ scarb build
 ```
 
 This will fetch all the dependencies and then build them, along with the package.
+
+You might want to simply check the package and all of its dependencies for errors without performing the last step of generating Sierra files.
```
maciektr, line 35:
```suggestion
You might want to simply check the package and all of its dependencies for errors without performing the final step of output files generation.
```
Not always sierra

maciektr (approved):
One small comment

PR #1109, opened 2024-02-01
Procedural macros design document

maciektr:


design/01-proc-macro.md
```diff
@@ -0,0 +1,290 @@
...
+    - Can remove the `ModuleItem`.
+    - Can contain Cairo code (as a virtual source file) to inject as a submodule.
+    - Emits compiler diagnostics.
+    - Can emit additional arbitrary data structure along generated file.
+
+Macro plugins are executed on each node in the Cairo AST.
+
+Custom plugins can be fed to the compiler via [`RootDatabaseBuilder::with_plugin_suite` API](https://github.com/starkware-libs/cairo/blob/8290898d602f87fb5bc3a18f28a42430cf6d3fdc/crates/cairo-lang-compiler/src/db.rs#L96). It takes a [`PluginSuite` object](https://github.com/starkware-libs/cairo/blob/8290898d602f87fb5bc3a18f28a42430cf6d3fdc/crates/cairo-lang-semantic/src/plugin.rs#L18-L25), which defines a set of objects implementing the [`MacroPlugin` trait](https://github.com/starkware-libs/cairo/blob/8290898d602f87fb5bc3a18f28a42430cf6d3fdc/crates/cairo-lang-defs/src/plugin.rs#L80).
+
+Scarb currently simply assumes a fixed, predefined plugin sets for `lib`, `starknet-contract` and `test` targets. This interface is not stable and is subject to changes in future compiler versions.
```
mkaput, line 38: That's not true anymore, because now we have builtin packages that provide plugins

design/01-proc-macro.md
```diff
@@ -0,0 +1,290 @@
...
+1. https://blog.jetbrains.com/rust/2022/03/18/procedural-macros-under-the-hood-part-i/
+2. https://blog.jetbrains.com/rust/2022/07/07/procedural-macros-under-the-hood-part-ii/
+
+### System programming primitives
+
+This post assumes reader knowledge on foreign function interface (FFI) mechanisms in modern operating systems.
+
+### Plugins API
+
+The Cairo compiler has a system of semantic plugins, that allow generating Cairo code with Rust code. The entry point of compiler plugins API is the [`MacroPlugin`] trait, which, via another mechanisms, boils down to allowing users to provide a callback, which:
```
mkaput, line 24: nit, several links like this are dead

design/01-proc-macro.md
```diff
@@ -0,0 +1,290 @@
...
+This document proposes an introduction of new compiler plugins type, called procedural macros, as described below.
+
+### Procedural macro user perspective
+
+To use a procedural macro, a Cairo programmer will have to:
+- Declare a dependency on a package, that implements the procedural macro, by adding it to the `dependencies` field in the Scarb manifest file.
+- Use the procedural macro in Cairo code, by calling it, or adding an attribute or derive macro to a Cairo item.
+
+The overall complexity of the actual procedural macro execution should be hidden from the user abstraction level.
+In particular, this two steps can be performed without any knowledge of Rust, or even the fact that the procedural macro is implemented in Rust.
```
mkaput, line 64: New macro downloads will require having Rust compiler installed though. This is worth noting here imho

design/01-proc-macro.md
```diff
@@ -0,0 +1,290 @@
...
+
+[cairo-plugin]
+```
+
+```rust
+/// The entry point of procedural macro implementation.
+pub fn generate_code(token_stream: TokenStream) -> ProcMacroResult {
+    // no-op
+    ProcMacroResult::Leave
+}
```
mkaput, line 158:
```suggestion
use cairo-lang-macro::{inline_macro, ProcMacroResult, TokenStream};

/// The entry point of procedural macro implementation.
#[inline_macro]
pub fn some_macro(token_stream: TokenStream) -> ProcMacroResult {
    // no-op
    ProcMacroResult::Leave
}
```

1. Let's emphasize usage of the API crate.
2. This outline does not say how macro name and kind is determined.
3. Without `extern "C"` Rust will mangle the `generate_code` name, so it wasn't a good FFI entrypoint. Wrapping this into custom procmacro will allow us to hide this implementation detail.

design/01-proc-macro.md
```diff
@@ -0,0 +1,290 @@
...
+## Reference-level explanation
+
+### Procedural macros will be compiled into shared libraries
+
+- Procedural macros will be implemented in Rust.
+- Procedural macros will be compiled into shared libraries, that Scarb can load at the runtime through dynamic loading mechanism of the host operating system.
+- Procedural macros will be compiled to `.dylib` files on OSX, `.so` files on Linux and `.dll` on Windows.
+- Scarb will use [`libloading` crate](https://docs.rs/libloading/latest/libloading/), which exposes a cross-platform interface for interactions with platform dynamic library loading primitives, to load shared libraries into memory.
+
+### Procedural macros will be special Scarb packages containing Rust code
```
mkaput, line 170: Two questions to resolve here:
1. Should the auto-generated Cargo.toml be included in package archives? Or should Scarb always re-generate Cargo.toml before compiling. Both variants have pros and cons I believe.
2. How will we control what version of the `cairo-lang-macro` crate will be used? How users will control compatibility with Scarb releases?

design/01-proc-macro.md
```diff
@@ -0,0 +1,290 @@
...
+version = "0.1.0"
+
+[lib]
+crate-type = "dylib"
+
+[dependencies]
+cairo-lang-macro = "1.0.0-alpha.7"
+```
+
+### Procedural macro packages can be used as regular dependencies
```
mkaput, line 223: These two sections are duplicated. Do you have a reasoning behind doing this? 🤔

design/01-proc-macro.md
```diff
@@ -0,0 +1,290 @@
...
+
+In order to meet the specified requirements, instead of syntax tree nodes, we can rather pass plain Cairo tokens which will be parsed by the procedural macro itself.
+
+A token can be defined as a string with one of three kinds assigned to it during parsing. The token kinds are: identifier, punctuation symbol and literal. Kinds of different tokens can be represented as an enum called TokenTree. An iterator over TokenTrees is called a TokenStream. A TokenStream is a representation of a Cairo code.
+
+Implementation of the communication protocol will require some adjustments in the Cairo language parser, so that it can be used for TokenStreams. Namely, it should be possible to use TokenStreams as input instead of Cairo files, skipping the lexical analysis required for Cairo files (as TokenStream can already be analyzed on the compiler side).
+
+## **Drawbacks**
+
+- Since shared libraries are loaded into the process memory space, some errors in the execution might result in hard to debug segmentation faults.
```
mkaput, line 283: I've been thinking of extracing Scarb's compiler machinery into a separate binary, that would be run as separate process. That level of isolation would at least prevent memory corruption from killing entire Scarb execution, so `scarb build` will at least be able to recover with good error message.

design/01-proc-macro.md
```diff
@@ -0,0 +1,290 @@
...
+- Procedural macros are also in fact just Rust code run on end user’s hardware. This makes them more susceptible to bad actors actions, than a standard Cairo source code. Thus, Procedural macros should always be treated with special attention, and only ones from trusted sources should be used.
```
mkaput, line 284:
```suggestion
- Procedural macros are also in fact just Rust code run on end user’s hardware. This makes them more susceptible to bad actors actions, than a standard Cairo source code. Thus, procedural macros should always be treated with special attention, and only ones from trusted sources should be used.
```

mkaput (changes requested):

design/01-proc-macro.md
maciektr: I will note this in "future possibilities" section if you are ok with that.

design/01-proc-macro.md
maciektr: Just in case someone skipped guide-level proposal and went straight to the reference-level one. I will reword the former to avoid straight duplication.

design/01-proc-macro.md
maciektr: 1. Personally, I think it should be regenerated on each project resolution. This will avoid problem of compatibility of the file content between Scarb versions. 
2. One idea I have is to treat the cairo-lang-macro versioning similar to what we do with `starknet` package today. That is, use a version equal to Scarb version.

design/01-proc-macro.md
mkaput: That would be great!

design/01-proc-macro.md
mkaput: You can just make references instead of copying paragraphs

design/01-proc-macro.md
mkaput: Ad 1. 👍🏻 
Ad 2. So you mean that users will use `cairo-version` to determine what `cairo-lang-macro` version are they targeting? I wonder if this wouldn't be too restrictive for us, and it wouldn't be better to decouple versioning schemes for these; like Rust's `syn`/`quote` crates do

design/01-proc-macro.md
maciektr: Actually, wouldn't it be easier to use package name as macro name? 🤔

mkaput: no, because I imagine it would be popular to have single package exposing a collection of related macros

maciektr: I understand this would be nice to have, but it would also open other problems - like we could no longer rely on scarb resolve to ensure no two macros with same name are declared.

mkaput: I think an ability to easily bundle multiple macros together, with shared state and code etc. is much bigger benefit than simplifying duplicate names check (which actually should boil down to just looping over all loaded plugins before feeding them into the DB as far as I imagine)

design/01-proc-macro.md
maciektr: Replaced with a reference.

MaksymilianDemitraszek (approved):
Don't see any major issues

design/01-proc-macro.md
```diff
@@ -0,0 +1,296 @@
...
+
+**Example:**
+
+The procedural macro source code:
+
+```rust
+pub fn generate_code(token_stream: TokenStream) -> ProcMacroResult {
+	// Implementation skipped for brevity.
+}
+```
```
mkaput, line 270: Outdated. I strongly suggest making references to the guide instead of copy-pasting exactly because of this reason 😅

mkaput (changes requested):

design/01-proc-macro.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Procedural macro author perspective
+
+To implement a procedural macro, a programmer will have to:
+- Create a new package, with a `Scarb.toml` manifest file and a `src/` directory besides.
+- The manifest file must define a `cairo-plugin` target type.
+- Write a Rust library, inside the `src/` directory, that implements the procedural macro API.
+- Note, that the package does not contain any `Cargo.toml` file, as it will be generated by Scarb.
```
piotmag769, line 128: Will a user be able to add needed rust packages as dependencies to the macro package?

piotmag769 (approved):
LGTM, really good job 🎉

tomek0123456789 (approved):
Didn't understand anything beforehand, now I understand something, thanks and good job!

design/01-proc-macro.md
tomek0123456789: Yes, I think you're looking for [this](https://github.com/software-mansion/scarb/pull/1109/files#diff-fa728b0211f1a1dbe27555484d8998cb7f4dd6970c00f90171419cfb5a585643R212)

maciektr: Yes, it will be possible.

design/01-proc-macro.md
maciektr: Updated with information about separate versioning of interface crates

glihm:
Looks great, amazing unlock for customization and abstraction in projects!

1. I'm not sure having read something about a possible limitation of one macro by package. Will it be like in rust, we can expose several macros in the same package?
2. In the rust code, the `TokenStream` a macro receives is exclusively the part of the tree that is concerned by the macro I guess:
```rust
#[derive(ToValue)]
struct Input {
    value: felt252,
}
```
```rust
// Let's say something similar to that.
#[derive_macro(ToValue)]
pub fn to_value(token_stream: TokenStream) -> ProcMacroResult {
    // only access the struct Input.
    ProcMacroResult::Leave
}
```
Is it considerable to give more context to the macro? I'm imagining some scenario where for instance the full path of the module in which the macro is executed, the crate id or even the file name could be useful. Or is this something not available at the time the macro is being called?

Awesome job on that guys, keen to contribute on that and see it evolve! :rocket:

design/01-proc-macro.md
```diff
@@ -0,0 +1,299 @@
...
+- The procedural macro will be triggered by one of three Cairo expressions
+    - Macro call, e.g. `xyz!`
+    - Macro attribute, e.g. `#[xyz]`
+    - Macro derive, e.g. `#[derive(Xyz)]`
+- The API for writing procedural macros for Cairo will be published for programmers, versioned separately from Scarb. This will provide a stability for the procedural macro implementation.
+- In total, the implementation will consist of three Rust crates, two of which will be visible by the macro author. These crates will be automatically added to the generated `Cargo.toml` file.
+  - First one, called `cairo-lang-macro-interface`, will contain the API definitions of the `TokenStream` and `ProcMacroResult` types used as input and output for macro implementation.
+  - The second one, called `cairo-lang-macro-attributes`, will contain implementation of Rust macros used for wrapping procedural macro entrypoint function. This will hide details of FFI communication from the procedural macro author.
```
mkaput, line 258: I think `-attributes` crate should be a dependency of the `-interface` one, and `-interface` should simply do `pub use _attributes::*`. That's what most crates do, for example `serde` vs `serde_derive`; or `clap`

Also, that way, the `-interface` part can be skipped imho

mkaput (changes requested):

design/01-proc-macro.md
maciektr: Yep, reworded, thank you!

maciektr:
Thanks for feedback @glihm !

> Looks great, amazing unlock for customization and abstraction in projects!

😍 

> 1. I'm not sure having read something about a possible limitation of one macro by package. Will it be like in rust, we can expose several macros in the same package?

Yes, it will be possible! Possibly not from the day one, but not long after it. 

> Is it considerable to give more context to the macro? I'm imagining some scenario where for instance the full path of the module in which the macro is executed, the crate id or even the file name could be useful. Or is this something not available at the time the macro is being called?

As of the design, the `ProcMacroHost` that would be responsible for execution of the procedural macros,  is in fact just a Cairo plugin. Consequently, it has access to the Cairo compiler database, and some additional metadata. So it would be technically possible to get the full module path, crate id etc. at the time of procedural macro execution. 

On the other hand, the current design proposes running procedural macro's only on simple and explicit request from the user made in the Cairo code, for instance by adding the `#[derive(MyMacro)]` attribute to their struct. So my concern is, that additional, custom logic based on some module metadata hard-coded  into the procedural macro would obfuscate the compilation process quite a bit 😟. Although, if the community decides otherwise - supporting additional metadata in the future should not be complicated.

mkaput (approved):

maciektr:
Hi everyone!
We will be merging this design document **this week**, so make sure to submit all your comments!

milancermak:
I've read through the doc. The whole apparatus seems very well though through as far as I can tell. Sadly, I haven't written a Rust macro in my life, but I have a couple of (potentially dumb, please bear with me) questions based on what I'd like to do with macros.

1. Will this be possible?

```rust
#[derive(Packing(v0))]
struct Values {
    a: u64,
    b: u8,
    c: u16,
    d: u128
}
```
The idea is that this macro would expand into a full implementation of `StorePacking` for the `Values` struct. I'm particularly interested in the `v0` part, because in the future, there can be a `v1` and `v2` that offer a smarter but backwards incompatible packing algorithm under the hood.

2. I'd like to have autogenerated getters (and maybe setters). Imagine having

```rust
struct Storage {
    #[getter]
    foo: u64,
    #[getter]
    bar: LegacyMap<ContractAddress, u128>
    #[getter]
    #[setter(guard=only_admin)]
    baz: ContractAddress
}
```

that would expand into sth like

```rust
fn get_foo(self: @ContractState) -> u64 {
    self.foo.read()
}

fn get_bar(self: @ContractState, value: ContractAddress) -> u128 {
    self.bar.read(value)
}

fn get_baz(self: @ContractState) -> ContractAddress {
    self.baz.read()
}

fn set_baz(ref self: ContractState, new_value: ContractAddress) {
    assert_only_admin();
    self.baz.write(new_value);
}
```

Will it be possible to have macros "stacked" like the `#[getter]` + `#[setter]` combo?

Also note that while this example looks cool, it complicates writing the `#[starknet::interface]` for the contract because the autogenerated fns would have to either be declared inside the `#[starknet::interface]` trait manually (which just complicates dev-ex) or be placed inside the contract but then available under a different interface. Maybe there's a more elegant solution to it, I haven't thought about it deeply.

3. This is probably too much on the crazy scale, but would it be possible to "hook into" the already present stuff like `#[starknet::interface]` and run some more macro expansion when that happens?

4. Let's say I want to mark a fn as non reentrant:

```rust
#[starknet::contract]
mod C {
    /// ....
    
    #[abi(embed_v0)]
    impl Foo of super::IFoo<ContractState> {
        /// ...
        
        #[non_reentrant]
        fn do_stuff_with_callback(ref self: ContractState) {
            /// ...
        }
    }
}
```

Now in this case, say the macro has a dependency on OZ's cairo-contracts, so it pulls in the [reentrancy guard](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/security/reentrancyguard.cairo) component. Can the `non_reentrant` macro place the `component!` code in the `mod C` scope (that is, two levels "above" it)? Will the code compile correctly, since `component!` is another macro, albeit a "native" one? What if I want to have multiple `#[non_reentrant]` fns in my impl? Will I be able to write the macro's Rust code in a way that it can detect there's already a `component!` in scope from a previous expansion? I hope this all makes sense... 😇 

5. Will it be possible to write a macro that looks sth like `tag! "foo"`, that is, without enclosing parentheses? Similar to the [`html!`](https://blog.jetbrains.com/rust/2022/03/18/procedural-macros-under-the-hood-part-i/) macro mentioned in the linked posts.

6. As mentioned, I have 0 experience writing Rust macros, but I imagine it would be great to have some kind of tool to help debug macro expansion. Are you thinking of offering sth like that, to make writing macros easier?

maciektr:
Hi @milancermak !

This are all really great questions! Thanks for your input. 

I don’t think I will exhaust all of these topics with my response, but I will try to give my thoughts right away. Please note, that the design doc in review is the first attempt at bringing said functionality to Cairo, so we intentionally want to keep things simple. Some topics mentioned here should ultimately become separate issues and be discussed as such. 

By question number:

1. Why not just `PackingV0` then? How would this be different? 🤔
2. I would imagine this to be implemented as a derive macro on struct, rather than attributes on struct fields. This way the macro implementation can modify struct implementation directly. As far as I know, that would also be the way to implement this in Rust (see `getset` or `derive_getters` libs as an example). As for “stacking” - yes, it will be possible to call multiple attribute macros on a single item. 
3. If I understand correctly what you mean, I don’t think so. We believe that the execution of macro implementation should be independent from other macros and plugins. They can be executed sequentially though. 
4. “Can the (...) macro place the `component!` code in `mod C` scope (…) two levels "above" it?”
No, the current design implies that the attribute can only change implementation of the item it is defined on (in this case the `do_stuff_with_callback` ). This makes the consequences of the proc macro execution more natural for Cairo code reader, by reducing the number of possible side effects. 
”Will the code compile correctly, since `component!` is another macro” 
Code generated by macros can call another macros. Unfortunately, as I said in previous point, attribute defined for a function cannot change the “parent” module definition.  
5. Good point! Ultimately, I would love to support impls as macro params (i.e. `macro! { smth }`) in Cairo, like it is supported in Rust. Quite frankly, I am not yet sure if we would need to make some changes to Cairo parser to implement this, or not. Sopossibly not from day one.   
6. We would like to support the expansion in the Cairo language server after it is available in Scarb, but this will require additional work on the LS side.

maciektr:
I would like to extend my thanks to everyone engaged in review and comments of this document. 😍 😍 😍  
I will merge this now, but please feel welcomed to continue the discussion in issues, feature requests, other PRs and our support channels.

PR #1110, opened 2024-02-01
Procedural macro compilation

### LINKED ISSUE #1129: Implement procedural macro package compilation mechanism
### Problem

May require manual definition of Cargo.toml manifests 

### Proposed Solution

_No response_

### Notes

_No response_

maciektr:
Resolves #1129

commit-id:aebc6424

---

**Stack**:
- #1161
- #1159
- #1157
- #1143
- #1151
- #1150
- #1148
- #1100
- #1155
- #1110 ⬅

### DIFF
--- a/scarb/src/compiler/compilation_unit.rs
+++ b/scarb/src/compiler/compilation_unit.rs
@@ -91,6 +91,10 @@ impl CompilationUnit {
         ws.target_dir().child(self.profile.as_str())
     }
 
+    pub fn is_cairo_plugin(&self) -> bool {
+        self.target().is_cairo_plugin()
+    }
+
     pub fn is_sole_for_package(&self) -> bool {
         self.main_component()
             .package
--- a/scarb/src/compiler/plugin/proc_macro/compilation.rs
+++ b/scarb/src/compiler/plugin/proc_macro/compilation.rs
@@ -1,8 +1,13 @@
 use crate::compiler::plugin::proc_macro::PROC_MACRO_BUILD_PROFILE;
-use crate::core::{Config, Package};
+use crate::compiler::CompilationUnit;
+use crate::core::{Config, Package, Workspace};
 use crate::flock::Filesystem;
+use crate::process::exec_piping;
+use anyhow::Result;
 use camino::Utf8PathBuf;
 use libloading::library_filename;
+use std::process::Command;
+use tracing::trace_span;
 
 /// This trait is used to define the shared library path for a package.
 pub trait SharedLibraryProvider {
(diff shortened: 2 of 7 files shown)


scarb/src/compiler/plugin/proc_macro/compilation.rs
```diff
@@ -26,3 +32,52 @@ impl SharedLibraryProvider for Package {
...
+struct CargoCommand {
+    current_dir: Utf8PathBuf,
+    target_dir: Utf8PathBuf,
+}
+
+impl From<CargoCommand> for Command {
+    fn from(args: CargoCommand) -> Self {
+        let mut cmd = Command::new("cargo");
+        cmd.stdout(Stdio::inherit());
+        cmd.stderr(Stdio::inherit());
```
mkaput, line 76: I'm not sure. I think it'd be better if we use the stderr redirection mechanism that we're using for `git` calls. Stdout is fine.

scarb/src/compiler/compilation_unit.rs
```diff
@@ -91,6 +91,10 @@ impl CompilationUnit {
         ws.target_dir().child(self.profile.as_str())
     }
 
+    pub fn is_cairo_plugin(&self) -> bool {
```
mkaput, line 94: Have you considered using an `enum` for representing compilation units? I'm not saying you have to use them, just wondering.

scarb/src/compiler/plugin/proc_macro/compilation.rs
```diff
@@ -26,3 +32,52 @@ impl SharedLibraryProvider for Package {
...
+                cmd
+            ));
+        }
+    }
+
+    Ok(())
+}
+
+#[derive(TypedBuilder)]
+#[builder(build_method(into = Command))]
```
mkaput, line 66: Do you really need the builder pattern here? It's one-off code

scarb/src/compiler/plugin/proc_macro/compilation.rs
```diff
@@ -26,3 +32,52 @@ impl SharedLibraryProvider for Package {
...
+        let _ = trace_span!("compile_proc_macro").enter();
+        let status = cmd
+            .status()
+            .with_context(|| format!("Failed to execute {:?}", cmd))?;
+        if !status.success() {
+            return Err(anyhow::anyhow!(
+                "Failed to compile procedural macro plugin: {:?}",
+                cmd
+            ));
+        }
```
mkaput, line 59: 1. errors start with lowercase
2. why not context chaining?
3. consider using `ensure!`

mkaput (changes requested):

scarb/src/compiler/plugin/proc_macro/compilation.rs
maciektr: Note, that in Cargo all human addressed communication is on stderr - e.g. `Compiling scarb v2.5.3 (/[..]/scarb/scarb)` etc.

scarb/src/compiler/compilation_unit.rs
maciektr: Done in #1155

scarb/src/compiler/plugin/proc_macro/compilation.rs
mkaput: Ok, so shouldn't we redirect this to stdout then to match what Scarb does?

scarb/src/compiler/plugin/proc_macro/compilation.rs
```diff
@@ -33,3 +38,46 @@ impl SharedLibraryProvider for Package {
...
+impl From<CargoCommand> for Command {
+    fn from(args: CargoCommand) -> Self {
+        let mut cmd = Command::new("cargo");
+        cmd.stdout(Stdio::inherit());
+        cmd.stderr(Stdio::inherit());
+        cmd.current_dir(args.current_dir);
+        cmd.args(["build", "--release"]);
+        cmd.arg("--target-dir");
+        cmd.arg(args.target_dir);
+        cmd
```
mkaput, line 81: consider using https://github.com/software-mansion/scarb/blob/fe47e6ab16acf742ae1a028e84ef79f573330632/scarb/src/process.rs#L90-L148 for smoother UX. You'd have to add a configuration option to deal with Cargo's stderr, but it should be fine imo

mkaput (changes requested):
See my answers

scarb/src/compiler/plugin/proc_macro/compilation.rs
maciektr: Hmm, to be honest I don't really see a point? Imho just passing both streams gives the most flexibility and the cleanest output 🤔 Would you mind to elaborate?

scarb/src/process.rs
```diff
@@ -87,9 +87,21 @@ mod imp {
...
+    }
+    exec_piping(cmd, config, print_fn)
+}
+
+/// Runs the process, waiting for completion, and mapping non-success exit codes to an error.
+#[tracing::instrument(level = "trace", skip_all)]
+pub fn exec_piping(
+    cmd: &mut Command,
+    config: &Config,
+    print_fn: fn(String, &Config),
```
mkaput, line 103: why not?

```suggestion
    print_fn: impl Fn(&str),
```

1. This would allow you to pass closures, so no `Config` plumbing
2. Passing `&str` would reduce allocation

mkaput (changes requested):

mkaput (approved):

PR #1113, opened 2024-02-02
Add `scarb cairo-run` tests

### LINKED ISSUE #1089: Write tests for `scarb cairo-run` program arguments
We are lacking tests for this argument:

https://github.com/software-mansion/scarb/blob/e7fc484c79b713bebd6fda5fa0bc5a977e988232/extensions/scarb-cairo-run/src/main.rs#L40-L46

tomek0123456789:
Closes #1089 

Please let me know if I've missed any tests cases, thanks!

### DIFF
--- a/extensions/scarb-cairo-run/tests/arguments.rs
+++ b/extensions/scarb-cairo-run/tests/arguments.rs
@@ -50,12 +50,15 @@ fn invalid_number_of_args() {
     Scarb::quick_snapbox()
         .arg("cairo-run")
         .arg("--")
-        .arg(r#"[2, 1, 3, 7]"#)
+        .arg(r#"[0, 1, 2, 3]"#)
         .current_dir(&t)
         .assert()
         .failure()
-        .stderr_matches(indoc! {r#"
-            Error: failed to run the function
+        .stdout_matches(indoc! {r#"
+               Compiling hello v0.1.0 ([..]/Scarb.toml)
+                Finished release target(s) in [..]
+                 Running hello
+            error: failed to run the function
 
             Caused by:
                 Function expects arguments of size 3 and received 4 instead.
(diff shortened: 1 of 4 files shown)


extensions/scarb-cairo-run/tests/arguments.rs
```diff
@@ -0,0 +1,106 @@
...
+        .build(t);
+}
+
+#[test]
+fn valid_number_of_args() {
+    let t = TempDir::new().unwrap();
+    setup_fib_three_felt_args(&t);
+
+    Command::new(cargo_bin("scarb"))
+        .env("SCARB_TARGET_DIR", t.path())
```
maciektr, line 31: I don't think you need it here. It has been added to tests that run on projects from `examples` directory, as in that case the `target` dir is shared by all the tests (e.g. `examples/hello_world/target` - this path does not change at all!). Here you setup a temp dir for a test project each time, so I think you should be good to go. 
While at it - would you mind adding a comment in the other tests file explaining why has this been done?

extensions/scarb-cairo-run/tests/arguments.rs
```diff
@@ -0,0 +1,210 @@
...
+
+#[test]
+fn invalid_number_of_args() {
+    let t = TempDir::new().unwrap();
+    setup_fib_three_felt_args(&t);
+
+    Scarb::quick_snapbox()
+        .arg("cairo-run")
+        .arg("--")
+        .arg(r#"[2, 1, 3, 7]"#)
```
maciektr, line 53: Please don't

tomek0123456789:
Not sure if all errors should be printed to `stdout`, as it might be a breaking change?

maciektr (approved):
Great work!

PR #1119, opened 2024-02-06
Check backward compatibility of metadata command deserialization

maciektr:
commit-id:2ea89d24

### DIFF
--- a/scarb-metadata/tests/command.rs
+++ b/scarb-metadata/tests/command.rs
@@ -1,14 +1,22 @@
 use assert_fs::TempDir;
 use snapbox::cmd::{cargo_bin, Command};
+use std::env;
+use std::path::PathBuf;
 
 use scarb_metadata::MetadataCommand;
+
+fn scarb_bin() -> PathBuf {
+    env::var_os("SCARB_TEST_BIN")
+        .map(PathBuf::from)
+        .unwrap_or_else(|| cargo_bin("scarb"))
+}
 
 #[test]
 fn empty_project() {
     let t = TempDir::new().unwrap();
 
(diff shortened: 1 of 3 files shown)


scarb-metadata/tests/command.rs
```diff
@@ -1,14 +1,22 @@
 use assert_fs::TempDir;
 use snapbox::cmd::{cargo_bin, Command};
+use std::env;
+use std::path::PathBuf;
 
 use scarb_metadata::MetadataCommand;
 
+fn scarb_bin() -> PathBuf {
+    env::var_os("SCARB_BIN_PATH")
```
mkaput, line 9:
```suggestion
    env::var_os("SCARB")
```

let's use our standard var name?

scarb-metadata/tests/command.rs
```diff
@@ -1,14 +1,22 @@
...
+        .map(PathBuf::from)
+        .unwrap_or(cargo_bin("scarb"))
```
mkaput, line 11:
```suggestion
        .unwrap_or_else(|| cargo_bin("scarb"))
```

scarb-metadata/tests/command.rs
maciektr: I have used a different var on purpose, not to confuse with other configuration when running tests locally on user env.

mkaput: Hmm, this was not obvious from me from `SCARB_BIN_PATH` name alone 😕 Your reasoning seems right, so perhaps rename this to something obviously telling that this influences tests?

mkaput (approved):

PR #1120, opened 2024-02-06
Prepare scarb-metadata release `1.11.1`

maciektr:


scarb-metadata/CHANGELOG.md
```diff
@@ -3,7 +3,9 @@
 All notable changes to this project will be documented in this file.
 
 ## Unreleased
-- Fix backward compatibility of `experimental_features` field deserialization.
+
+## 1.11.1 (2024-02-06)
+Fix backward compatibility of `experimental_features` field deserialization.
```
mkaput, line 8:
```suggestion
- Fix backward compatibility of `experimental_features` field deserialization.
```

mkaput:
I think we should also yank `1.11.0` on crates.io

maciektr:
> I think we should also yank `1.11.0` on crates.io

Yep, agree :/

PR #1135, opened 2024-02-07
Fix `edition` for main crates in test compilation in test collector

piotmag769:


extensions/scarb-snforge-test-collector/src/compilation/test_collector.rs
```diff
@@ -180,13 +188,19 @@ fn insert_lib_entrypoint_content_into_db(
...
-            crate_root.to_path_buf(),
-        ))),
+        Some(CrateConfiguration {
+            root: Directory::Real(crate_root.to_path_buf()),
+            settings: CrateSettings {
+                edition: main_crate_edition,
+                cfg_set: None,
+                experimental_features: Default::default(),
+            },
+        }),
```
piotmag769, line 203: This is the actual fix

extensions/scarb-snforge-test-collector/src/metadata.rs
```diff
@@ -150,11 +167,29 @@ impl CompilationUnit<'_> {
...
-        self.unit_metadata.target.source_root().to_path_buf()
+    pub fn main_package_source_root(&self) -> Utf8PathBuf {
+        self.main_package_metadata.source_root().to_path_buf()
     }
 
-    pub fn source_file_path(&self) -> &Utf8Path {
-        &self.unit_metadata.target.source_path
+    pub fn main_package_source_file_path(&self) -> &Utf8Path {
+        &self.main_package_metadata.source_path
+    }
```
piotmag769, line 176: Changed it bcs we actually want the source root and source file path of the package we are compiling the tests for, not of the target. They are the same for now bcs we take the CU with main package being the package we want to compile tests for, but the logic it is less prone to break in the future this way.
cc @maciektr to check if I'm right

maciektr: Ok, if you don't fully integrate with the Scarb data model it kind of makes sense.

extensions/scarb-snforge-test-collector/src/compilation/test_collector.rs
```diff
@@ -180,13 +188,19 @@ fn insert_lib_entrypoint_content_into_db(
...
         main_crate_id,
-        Some(CrateConfiguration::default_for_root(Directory::Real(
-            crate_root.to_path_buf(),
-        ))),
+        Some(CrateConfiguration {
+            root: Directory::Real(crate_root.to_path_buf()),
+            settings: CrateSettings {
+                edition: main_crate_edition,
+                cfg_set: None,
+                experimental_features: Default::default(),
```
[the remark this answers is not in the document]
piotmag769: Done

Arcticae (approved):

PR #1143, opened 2024-02-13
Add aux_data to proc macro result

maciektr:
commit-id:89f844e2

---

**Stack**:
- #1165
- #1161
- #1159
- #1157
- #1143 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #1155, opened 2024-02-19
Change compilation unit to enum

maciektr:
commit-id:bf749c51

---

**Stack**:
- #1165
- #1161
- #1159
- #1157
- #1143
- #1155 ⬅

### DIFF
--- a/scarb/src/compiler/compilation_unit.rs
+++ b/scarb/src/compiler/compilation_unit.rs
@@ -11,10 +11,16 @@ use crate::flock::Filesystem;
 use crate::internal::stable_hash::StableHasher;
 
 /// An object that has enough information so that Scarb knows how to build it.
+pub enum CompilationUnit {
+    Cairo(CairoCompilationUnit),
+    ProcMacro(ProcMacroCompilationUnit),
+}
+
+/// An object that has enough information so that Scarb knows how to build Cairo code with it.
 #[derive(Clone, Debug)]
 #[non_exhaustive]
-pub struct CompilationUnit {
-    /// The Scarb [`Package`] to be build.
+pub struct CairoCompilationUnit {
+    /// The Scarb [`Package`] to be built.
     pub main_package_id: PackageId,
 
     /// Collection of all [`Package`]s needed to provide as _crate roots_ to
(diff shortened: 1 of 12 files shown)


scarb/src/ops/metadata.rs
```diff
@@ -270,6 +262,50 @@ fn collect_compilation_unit_metadata(
         .unwrap()
 }
 
+fn collect_proc_macro_compilation_unit_metadata(
+    compilation_unit: &ProcMacroCompilationUnit,
+) -> m::CompilationUnitMetadata {
+    let components = collect_compilation_unit_components(compilation_unit.components.iter());
+    m::CompilationUnitMetadataBuilder::default()
+        .id(compilation_unit.id())
+        .package(wrap_package_id(compilation_unit.main_package_id))
```
tomek0123456789, line 271: If the method `main_package_id` stays, shouldn't it be used here? A bit of an overhead, but we should stay consistent with it

scarb/src/ops/resolve.rs
```diff
@@ -153,22 +156,25 @@ pub fn generate_compilation_units(
...
+            _ => None,
+        })
         .flat_map(|unit| unit.cairo_plugins.clone())
         .filter(|plugin| !plugin.builtin)
         .map(|plugin| plugin.package.clone())
+        .chain(ws.members().filter(|member| member.is_cairo_plugin()))
         .unique_by(|plugin| plugin.id)
         .collect_vec();
 
+    // cairo_plugins.extend(ws.members().filter(|member| member.is_cairo_plugin()));
```
tomek0123456789, line 176: 👀

scarb/src/ops/resolve.rs
```diff
@@ -439,10 +445,11 @@ fn check_cairo_version_compatibility(packages: &[Package], ws: &Workspace<'_>) -
 
 fn generate_cairo_plugin_compilation_units(
     member: &Package,
-    ws: &Workspace<'_>,
+    _ws: &Workspace<'_>,
```
tomek0123456789, line 448: Did this parameter stay in order to be consistent between `generate_cairo_compilation_units` and this function?

tomek0123456789 (commented):
image

scarb/src/compiler/compilation_unit.rs
```diff
@@ -128,12 +137,42 @@ impl CompilationUnit {
             write!(&mut string, " ").unwrap();
         }
 
-        write!(&mut string, "{}", self.main_package_id).unwrap();
+        write!(&mut string, "{}", self.main_package_id()).unwrap();
```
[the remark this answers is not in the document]
maciektr: Note, that this is actually a default implementation for a function in trait. It's not implemented for a struct (but rather a trait), so it cannot access any struct fields, only functions exposed by the trait.

scarb/src/ops/resolve.rs
maciektr: Indeed, can be removed.

scarb/src/ops/metadata.rs
maciektr: Note, that the whole existence of the `main_package_id()` is rather a technicality - so it can be used in default trait impls. All the fields in both CU structs are still public, so can be used outside it's implementation without getters.

scarb/src/compiler/compilation_unit.rs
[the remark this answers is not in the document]
tomek0123456789: Oh, didn't notice that it is inside the trait, my bad

tomek0123456789 (approved):
lgtm!

scarb/src/compiler/compilation_unit.rs
```diff
@@ -145,6 +184,45 @@ impl CompilationUnit {
...
+        }
+    }
+
+    pub fn target_dir(&self, ws: &Workspace<'_>) -> Filesystem {
+        ws.target_dir().child(self.profile.as_str())
+    }
+
+    pub fn is_cairo_plugin(&self) -> bool {
+        self.target().is_cairo_plugin()
+    }
```
mkaput, line 223: Can you please add a doc comment stating why does this method exist for cairo compilation unit, given that we have a different enum branch to represent plugins? As this looks contradictory from the first glimpse.

mkaput (approved):

scarb/src/compiler/compilation_unit.rs
maciektr: Good point, removed it.

PR #1159, opened 2024-02-19
Implement procedural macro diagnostics

maciektr:
commit-id:f06e9653

---

**Stack**:
- #1166
- #1165
- #1161
- #1159 ⬅

### DIFF
--- a/plugins/cairo-lang-macro-stable/src/ffi.rs
+++ b/plugins/cairo-lang-macro-stable/src/ffi.rs
@@ -1,0 +1,37 @@
+/// This struct encapsulates a stable ABI representation of information required to construct a slice.
+///
+/// Please be aware that the memory management of values passing the FFI-barrier is tricky.
+/// The memory must be freed on the same side of the barrier, where the allocation was made.
+#[repr(C)]
+#[derive(Debug)]
+#[doc(hidden)]
+pub struct StableSlice<T> {
+    ptr: *mut T,
+    len: usize,
+}
+
+impl<T> StableSlice<T> {
+    /// Create a new `StableSlice` from a Vector.
+    /// Note that the vector will not be deallocated automatically.
+    /// Please make sure to use `into_owned` afterward, to free the memory.
+    pub fn new(mut x: Vec<T>) -> Self {
+        x.shrink_to_fit();
+        assert_eq!(x.len(), x.capacity());
+        let ptr = x.as_mut_ptr();
+        let len = x.len();
+        std::mem::forget(x);
+        Self { ptr, len }
+    }
+
+    /// Convert to owned vector.
+    pub fn into_owned(self) -> Vec<T> {
+        unsafe { Vec::from_raw_parts(self.ptr, self.len, self.len) }
+    }
+
+    /// Returns raw pointer and length.
+    /// Can be used to construct a slice.
+    /// No ownership is transferred.
+    pub fn raw_parts(&self) -> (*mut T, usize) {
+        (self.ptr, self.len)
+    }
+}
--- a/plugins/cairo-lang-macro-stable/src/lib.rs
+++ b/plugins/cairo-lang-macro-stable/src/lib.rs
@@ -1,35 +1,65 @@
+use crate::ffi::StableSlice;
 use std::ffi::{CStr, CString};
+use std::num::NonZeroU8;
 use std::os::raw::c_char;
+
+pub mod ffi;
 
 /// Token stream.
 ///
 /// This struct implements FFI-safe stable ABI.
 #[repr(C)]
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct StableTokenStream(*mut c_char);
 
 #[repr(C)]
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub enum StableAuxData {
     None,
     Some(*mut c_char),
 }
 
+/// Diagnostic returned by the procedural macro.
+///
+/// This struct implements FFI-safe stable ABI.
+#[repr(C)]
+#[derive(Debug)]
+pub struct StableDiagnostic {
+    pub message: *mut c_char,
+    pub severity: StableSeverity,
+}
+
+/// The severity of a diagnostic.
+///
+/// This struct implements FFI-safe stable ABI.
+pub type StableSeverity = NonZeroU8;
+
 /// Procedural macro result.
 ///
 /// This struct implements FFI-safe stable ABI.
 #[repr(C)]
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub enum StableProcMacroResult {
     /// Plugin has not taken any action.
-    Leave,
+    Leave {
+        diagnostics: StableSlice<StableDiagnostic>,
+    },
     /// Plugin generated [`StableTokenStream`] replacement.
     Replace {
+        diagnostics: StableSlice<StableDiagnostic>,
         token_stream: StableTokenStream,
         aux_data: StableAuxData,
     },
     /// Plugin ordered item removal.
-    Remove,
+    Remove {
+        diagnostics: StableSlice<StableDiagnostic>,
+    },
+}
+
+#[repr(C)]
+pub struct StableResultWrapper {
+    pub input: StableTokenStream,
+    pub output: StableProcMacroResult,
 }
 
 impl StableTokenStream {
--- a/plugins/cairo-lang-macro/src/lib.rs
+++ b/plugins/cairo-lang-macro/src/lib.rs
@@ -1,20 +1,26 @@
+pub use cairo_lang_macro_attributes::*;
+use cairo_lang_macro_stable::ffi::StableSlice;
+use cairo_lang_macro_stable::{
+    StableAuxData, StableDiagnostic, StableProcMacroResult, StableSeverity, StableTokenStream,
+};
 use std::ffi::{c_char, CStr, CString};
 use std::fmt::Display;
-
-pub use cairo_lang_macro_attributes::*;
-use cairo_lang_macro_stable::{StableAuxData, StableProcMacroResult, StableTokenStream};
+use std::num::NonZeroU8;
+use std::slice;
+use std::vec::IntoIter;
 
 #[derive(Debug)]
 pub enum ProcMacroResult {
     /// Plugin has not taken any action.
-    Leave,
+    Leave { diagnostics: Vec<Diagnostic> },
     /// Plugin generated [`TokenStream`] replacement.
     Replace {
         token_stream: TokenStream,
         aux_data: Option<AuxData>,
+        diagnostics: Vec<Diagnostic>,
     },
     /// Plugin ordered item removal.
-    Remove,
+    Remove { diagnostics: Vec<Diagnostic> },
 }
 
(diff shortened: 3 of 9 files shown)


plugins/cairo-lang-macro/src/lib.rs
```diff
@@ -143,6 +213,91 @@ impl AuxData {
...
+    ///
+    /// # Safety
+    #[doc(hidden)]
+    pub unsafe fn allocate(diagnostics: Vec<Self>) -> (*mut StableDiagnostic, usize) {
+        let stable_diagnostics = diagnostics
+            .into_iter()
+            .map(|diagnostic| diagnostic.into_stable())
+            .collect::<Vec<_>>();
+        let n = stable_diagnostics.len();
+        let ptr = malloc(std::mem::size_of::<StableDiagnostic>() * n) as *mut StableDiagnostic;
```
mkaput, line 250: https://doc.rust-lang.org/stable/std/alloc/struct.System.html

> However, it is not valid to mix use of the backing system allocator with System, as this implementation may include extra work, such as to serve alignment requests greater than the alignment provided directly by the backing system allocator.

So I don't think it's valid to use `malloc/free` for the solution. Nor it is elegant IMHO.

Also, as far as I understand, you should be safe to use `Box<T, std::alloc::System>` in FFI boundary because [`Box` has a well-defined memory layout](https://doc.rust-lang.org/beta/std/boxed/index.html#memory-layout), and `std::alloc::System` should be always the same entity in our setup on all platforms.

And, because memory layout of `str` is the same as the layout of `[u8]`, I think you should be safe to use the following types, instead of pointer shenanigans that happen here:
* `Box<str, System>`
* `Box<[(Severity, Box<str, System>)], System>` (tuple might be replaced by `repr(C)` struct, I believe)

Because these types live within `System` allocator, their `Drop` implementations should deal with memory ownership for you, though you might need to use `std::mem::forget` and `Box::from_raw` or smth.

Reference: https://doc.rust-lang.org/reference/type-layout.html

wdyt?

mkaput (changes requested):

plugins/cairo-lang-macro/src/lib.rs
maciektr: Thanks for linked materials! Seems like `Box<>` is the way to go indeed.

plugins/cairo-lang-macro/Cargo.toml
```diff
@@ -14,6 +14,7 @@ repository.workspace = true
 
 [dependencies]
 anyhow.workspace = true
+libc.workspace = true
```
mkaput, line 17: Unneeded

mkaput (changes requested):

plugins/cairo-lang-macro/src/lib.rs
```diff
@@ -53,22 +57,70 @@ impl Display for AuxData {
...
     /// Convert to FFI-safe representation.
     ///
     /// # Safety
     #[doc(hidden)]
     pub fn into_stable(self) -> StableProcMacroResult {
         match self {
-            ProcMacroResult::Leave => StableProcMacroResult::Leave,
-            ProcMacroResult::Remove => StableProcMacroResult::Remove,
+            ProcMacroResult::Leave { diagnostics } => {
+                let (ptr, n) = unsafe { Diagnostic::allocate(diagnostics) };
```
mkaput, line 98: could you please elaborate why do you need these allocate/deallocate functions, when you can pass Box etc. via FFI boundary, which have Drop implemented etc?

scarb/tests/build_cairo_plugin.rs
```diff
@@ -252,3 +257,97 @@ fn compile_cairo_plugin_with_other_target() {
...
+        .arg("build")
+        // Disable output from Cargo.
+        .env("CARGO_TERM_QUIET", "true")
+        .current_dir(&project)
+        .assert()
+        .success()
+        .stdout_matches(indoc! {r#"
+            [..] Compiling some v1.0.0 ([..]Scarb.toml)
+            [..] Compiling hello v1.0.0 ([..]Scarb.toml)
+            warn: Plugin diagnostic: Some warning from macro.
```
mkaput, line 299: > `Plugin diagnostic:`

Where does this come from? Is it neccessary? I don't feel we need it?

plugins/cairo-lang-macro-stable/src/lib.rs
```diff
@@ -15,21 +15,49 @@ pub enum StableAuxData {
     Some(*mut c_char),
 }
 
+/// Diagnostic returned by the procedural macro.
+///
+/// This struct implements FFI-safe stable ABI.
+#[repr(C)]
+#[derive(Debug)]
+pub struct StableDiagnostic {
+    pub message: *mut c_char,
```
mkaput, line 24:
```suggestion
    pub message: Box<str, System>,
```

why not etc.

mkaput (changes requested):

plugins/cairo-lang-macro/src/lib.rs
maciektr: Sure, apparently while Box is safe to pass, a slice is not. There even is a clippy lint.
```
warning: `extern` fn uses type `[cairo_lang_macro_stable::StableDiagnostic]`, which is not FFI-safe
  --> scarb/src/compiler/plugin/proc_macro/ffi.rs:75:19
   |
75 | type ExpandCode = extern "C" fn(StableTokenStream) -> StableProcMacroResult;
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not FFI-safe
   |
   = help: consider using a raw pointer instead
   = note: slices have no C equivalent
   = note: `#[warn(improper_ctypes_definitions)]` on by default
```
I've also tried `Box<[Box<str>]>` with the same results 😢 
So next best thing for me is to use `into_raw` etc. 
What do you think? 🤔

scarb/tests/build_cairo_plugin.rs
maciektr: I don't think we need it at all, but it comes from Cairo compiler. 
All diagnostics from plugins are registered as `SemanticDiagnostics`, which format with such prefix. 

https://github.com/starkware-libs/cairo/blob/a771425155ec5ea740eea5e7ee00ea5870b498a3/crates/cairo-lang-semantic/src/db.rs#L1126

https://github.com/starkware-libs/cairo/blob/a771425155ec5ea740eea5e7ee00ea5870b498a3/crates/cairo-lang-semantic/src/diagnostic.rs#L511

plugins/cairo-lang-macro/src/lib.rs
mkaput: Maybe try the following?
1. Wrap the pointers into a struct to isolate this unsafety. Like this: https://docs.rs/ffi-std/latest/src/ffi_std/slice.rs.html (though don't pull this as a library)
2. If needed, use [`ManuallyDrop`](https://doc.rust-lang.org/stable/std/mem/struct.ManuallyDrop.html) to pass these things via FFI and still control deallocation via `Drop`.

scarb/tests/build_cairo_plugin.rs
mkaput: meh

plugins/cairo-lang-macro/src/lib.rs
maciektr: Great, I'll try right away!

plugins/cairo-lang-macro-stable/src/ffi.rs
```diff
@@ -0,0 +1,44 @@
...
+    }
+}
+
+impl<T> StableSlice<T> {
+    /// Create a new `StableSlice` from a Vector.
+    /// Note that the vector will not be deallocated automatically.
+    /// Please make sure to use `into_owned` afterward, to free the memory.
+    pub fn new(mut x: Vec<T>) -> Self {
+        x.shrink_to_fit();
+        assert_eq!(x.len(), x.capacity());
```
MaksymilianDemitraszek, line 26: Is this actually a guaranteed property of `shrink_to_fit`? Docs do not say that, If it is not we probably have to store capacity as well in the `StableSlice`? 🤔 Not sure though

plugins/cairo-lang-macro-stable/src/ffi.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub fn into_owned(self) -> Vec<T> {
+        unsafe { Vec::from_raw_parts(self.ptr, self.len, self.len) }
+    }
+
+    /// Returns raw pointer and length.
+    /// Can be used to construct a slice.
+    /// No ownership is transferred.
+    pub fn into_raw_parts(self) -> (*mut T, usize) {
+        (self.ptr, self.len)
+    }
```
MaksymilianDemitraszek, line 43: I am not sure what is the purpose of this function. Wouldn't be possible to have `into_vec` instead? In this scenario seems cleaner to have function `new` and its inverse. Probably would require to ensure T can be converted from stable as well, but seems like desired property.

MaksymilianDemitraszek (commented):
Some comments, I am not fully in the loop yet so keep in mind those might be dumb.

plugins/cairo-lang-macro-stable/src/ffi.rs
maciektr: I use it to construct a slice from raw pointer, it could return a slice already, but lifetimes management gets tricky then and I just kind of did this as a workaround 😅

MaksymilianDemitraszek: I wouldn't bother, this is not user facing API

plugins/cairo-lang-macro-stable/src/ffi.rs
```diff
@@ -0,0 +1,45 @@
+/// This struct encapsulates a stable ABI representation of struct.
+///
+/// Please be aware that the memory management of values passing the FFI-barrier is tricky.
+/// The memory must be freed on the same side of the barrier, where the allocation was made.
+#[repr(C)]
+#[derive(Debug)]
+#[doc(hidden)]
+pub struct StableSlice<T> {
```
mkaput, line 8: If this is taking ownership, then the implementations of Copy and Clone are incorrect because you must clone items, and copy by definition should not exist at all.

If this is taking referefence, then you must protect items with lifetimes

plugins/cairo-lang-macro-stable/src/lib.rs
```diff
@@ -15,21 +18,52 @@ pub enum StableAuxData {
...
+/// The severity of a diagnostic.
+///
+/// This struct implements FFI-safe stable ABI.
+#[repr(C)]
+#[derive(Debug)]
+pub enum StableSeverity {
+    Error,
+    Warning,
+}
+
```
mkaput, line 40: I think this should be just

```rust
type StableSeverity = NonZero<u8>;
```

By using enums you missed two erroneous cases:
1. Scarb introduced a new severity level, but the plugin is using old API
2. Plugin is using new API with new severity level, but it's being loaded into old Scarb.

mkaput (changes requested):

PR #1166, opened 2024-03-01
Implement aux_data callback

maciektr:


maciektr:
Replacing with https://github.com/software-mansion/scarb/pull/1186 (sorry)

PR #1161, opened 2024-02-22
Add remove/replace proc macro e2e tests

maciektr:
commit-id:ecc87fd5

---

**Stack**:
- #1166
- #1165
- #1161 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #1174, opened 2024-03-05
Fix memory managment around the FFI-barrier

maciektr:
commit-id:e557eb16

---

**Stack**:
- #1166
- #1165
- #1161
- #1159
- #1174 ⬅


mkaput (changes requested):
As discussed offline, let's postpone fixing these issues and just leak memory as for now

mkaput (approved):

PR #1184, opened 2024-03-13
Remove macro_commons

maciektr:
**Stack**:
- #1166
- #1184 ⬅

### DIFF
--- a/plugins/cairo-lang-macro/src/lib.rs
+++ b/plugins/cairo-lang-macro/src/lib.rs
@@ -9,6 +9,14 @@ use std::num::NonZeroU8;
 use std::slice;
 use std::vec::IntoIter;
 
+/// Free the memory allocated for the [`StableProcMacroResult`].
+///
+/// # Safety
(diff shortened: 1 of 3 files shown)


plugins/cairo-lang-macro/src/lib.rs
```diff
@@ -9,6 +9,14 @@ use std::num::NonZeroU8;
 use std::slice;
 use std::vec::IntoIter;
 
+/// Free the memory allocated for the [`StableProcMacroResult`].
+///
+/// # Safety
+#[no_mangle]
+pub unsafe extern "C" fn free_result(result: StableProcMacroResult) {
+    ProcMacroResult::from_owned_stable(result);
+}
```
mkaput, line 18: I believe this should be `#[doc(hidden)]` as this is implementation detail which the user should not bother about.

Also please document here what we are actually doing with this `#[no_mangle]` thingy

mkaput (changes requested):

PR #1186, opened 2024-03-13
Implement aux_data callback

maciektr:
commit-id:a9ba1335

---

**Stack**:
- #1200
- #1195
- #1194
- #1192
- #1186 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #1187, opened 2024-03-14
Move Scarb StableHash to separate crate

maciektr:
**Stack**:
- #1194
- #1192
- #1186
- #1188
- #1187 ⬅


utils/scarb-stable-hash/src/lib.rs
```diff
@@ -7,6 +7,7 @@ use xxhash_rust::xxh3::Xxh3;
 ///
 /// The hasher should be fast and have a low chance of collisions (but is not sufficient for
 /// cryptographic purposes).
+#[derive(Default)]
```
mkaput, line 10: if we derive Default, then let's `StableHasher::new` use it

it amazes me that clippy didn't catch this

mkaput (approved):

PR #1188, opened 2024-03-14
Store AuxData as Vec<u8> instead of serde/string

maciektr:
commit-id:eacd6dd6

---

**Stack**:
- #1195
- #1194
- #1192
- #1186
- #1188 ⬅

### DIFF
--- a/scarb/tests/build_cairo_plugin.rs
+++ b/scarb/tests/build_cairo_plugin.rs
@@ -91,7 +91,7 @@ fn simple_project_with_code(t: &impl PathChild, code: impl ToString) {
 
 fn simple_project(t: &impl PathChild) {
     let code = indoc! {r#"
-        use cairo_lang_macro::{ProcMacroResult, TokenStream, attribute_macro};
+        use cairo_lang_macro::{ProcMacroResult, TokenStream, attribute_macro, AuxData};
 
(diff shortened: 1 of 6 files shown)


scarb/tests/build_cairo_plugin.rs
```diff
@@ -84,14 +84,16 @@ fn simple_project_with_code(t: &impl PathChild, code: impl ToString) {
         [dependencies]
         cairo-lang-macro = {{ path = {macro_lib_path}}}
         cairo-lang-macro-stable = {{ path = {macro_stable_lib_path}}}
+        serde = {{ version = "*", features = ["derive"] }}
+        serde_json = "*"
```
mkaput, line 88: unused?

scarb/tests/build_cairo_plugin.rs
mkaput, line 1: There is no test that is affected by this change. Is this by design? 🤔

mkaput (changes requested):

scarb/tests/build_cairo_plugin.rs
maciektr: Test case for aux data is added with #1186, as it would really make sense to emit any with no way to collect them back.

mkaput (approved):

PR #1190, opened 2024-03-14
Add Cairo -> Sierra mappings to test artifacts

piotmag769:
Not sure if I will need it in the profiler now, but I prob will in the future. It doesn't break anything since we won't consume this field in `snforge`


maciektr (approved):

piotmag769:
Will be rewriting this

PR #1192, opened 2024-03-15
Store file path and file id in token stream metadata

maciektr:
commit-id:98fd5a13

---

**Stack**:
- #1200
- #1195
- #1194
- #1192 ⬅

### DIFF
--- a/plugins/cairo-lang-macro-stable/src/lib.rs
+++ b/plugins/cairo-lang-macro-stable/src/lib.rs
@@ -1,23 +1,45 @@
 use crate::ffi::StableSlice;
-use std::ffi::{CStr, CString};
+use std::ffi::CStr;
 use std::num::NonZeroU8;
 use std::os::raw::c_char;
+use std::ptr::NonNull;
 
 pub mod ffi;
+
+/// An option.
+///
+/// This struct implements FFI-safe stable ABI.
+#[repr(C)]
+#[derive(Debug)]
+pub enum StableOption<T> {
+    None,
+    Some(T),
+}
 
 /// Token stream.
 ///
 /// This struct implements FFI-safe stable ABI.
 #[repr(C)]
 #[derive(Debug)]
-pub struct StableTokenStream(*mut c_char);
+pub struct StableTokenStream {
+    pub value: *mut c_char,
+    pub metadata: StableTokenStreamMetadata,
+}
 
+/// Token stream metadata.
+///
+/// This struct implements FFI-safe stable ABI.
 #[repr(C)]
 #[derive(Debug)]
-pub enum StableAuxData {
-    None,
-    Some(StableSlice<u8>),
+pub struct StableTokenStreamMetadata {
+    pub original_file_path: Option<NonNull<c_char>>,
+    pub file_id: Option<NonNull<c_char>>,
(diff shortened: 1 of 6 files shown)


plugins/cairo-lang-macro-stable/src/lib.rs
```diff
@@ -1,24 +1,44 @@
...
+/// An option.
+///
+/// This struct implements FFI-safe stable ABI.
+#[repr(C)]
+#[derive(Debug)]
+pub enum StableOption<T> {
+    None,
+    Some(T),
+}
+
```
mkaput, line 17: You're using this to store pointers. Can't you put `NonNull` there and use standard Option instead?

https://doc.rust-lang.org/stable/std/option/index.html#representation

mkaput (changes requested):

scarb/tests/build_cairo_plugin.rs
```diff
@@ -530,7 +530,50 @@ fn can_return_aux_data_from_plugin() {
...
+    let temp = TempDir::new().unwrap();
+    let t = temp.child("some");
+    simple_project_with_code(
+        &t,
+        indoc! {r##"
+        use cairo_lang_macro::{ProcMacroResult, TokenStream, attribute_macro};
+
+        #[attribute_macro]
+        pub fn some_macro(token_stream: TokenStream) -> ProcMacroResult {
+            println!("{:?}", token_stream.metadata());
```
mkaput, line 547:
```suggestion
            println!("{:#?}", token_stream.metadata());
```

and you'll get much easier to diff output

plugins/cairo-lang-macro/src/lib.rs
```diff
@@ -66,18 +69,59 @@ pub enum ProcMacroResult {
...
 #[derive(Debug, Default, Clone)]
-pub struct TokenStream(String);
+pub struct TokenStream {
+    value: String,
+    metadata: TokenStreamMetadata,
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct TokenStreamMetadata {
+    original_file_path: Option<String>,
```
mkaput, line 79: Do we need file path if we have fileid ?

We wanted to include paths in spans anyway, and in this context both fields provide the same value

plugins/cairo-lang-macro/src/lib.rs
```diff
@@ -66,18 +69,59 @@ pub enum ProcMacroResult {
...
+            s.with_file_path(file_path)
+        } else {
+            s
+        }
+    }
+
+    #[doc(hidden)]
+    pub fn with_file_path(mut self, file_path: impl ToString) -> Self {
+        self.original_file_path = Some(file_path.to_string());
+        self.file_id = Some(short_hash(file_path.to_string()));
```
mkaput, line 123: shouldn't this hash be computed inside scarb?

mkaput (changes requested):

plugins/cairo-lang-macro/src/lib.rs
maciektr: Discussed offline

mkaput (approved):

PR #1203, opened 2024-03-19
fix: derive Clone and Debug for CompilationUnit enum

glihm:
Currently, all the variant of the enum derive `Clone` and `Debug`. Could be helpful to clone regardless of the variant if it's ok for you. :+1:


maciektr (approved):

maciektr:
Sure, thank you!

PR #1213, opened 2024-03-21
Rust update - fix build and clippy erorrs

maciektr:
commit-id:77034927

---

**Stack**:
- #1220
- #1219
- #1218
- #1217
- #1216
- #1215
- #1214
- #1221
- #1222
- #1213 ⬅

### DIFF
--- a/scarb/src/core/registry/client/local.rs
+++ b/scarb/src/core/registry/client/local.rs
@@ -213,6 +213,7 @@ fn edit_records(records_path: &Path, func: impl FnOnce(&mut IndexRecords)) -> Re
         .read(true)
         .write(true)
         .create(true)
+        .truncate(false)
         .open(records_path)
         .context("failed to open file")?;
 
--- a/scarb/src/flock.rs
+++ b/scarb/src/flock.rs
@@ -148,8 +148,9 @@ pub struct AdvisoryLock<'f> {
     config: &'f Config,
 }
 
-#[derive(Debug)]
-pub struct AdvisoryLockGuard(Arc<FileLockGuard>);
+pub struct AdvisoryLockGuard {
+    _inner: Arc<FileLockGuard>,
+}
 
 impl<'f> AdvisoryLock<'f> {
     /// Acquires this advisory lock in an async manner.
@@ -172,7 +173,9 @@ impl<'f> AdvisoryLock<'f> {
                 arc
             }
         };
-        Ok(AdvisoryLockGuard(file_lock_arc))
+        Ok(AdvisoryLockGuard {
+            _inner: file_lock_arc,
+        })
     }
 }
 
--- a/scarb/src/ops/lockfile.rs
+++ b/scarb/src/ops/lockfile.rs
@@ -12,6 +12,7 @@ pub fn read_lockfile(ws: &Workspace<'_>) -> Result<Lockfile> {
         .read(true)
         .write(true)
         .create(true)
+        .truncate(false)
         .open(ws.lockfile_path())
         .context("failed to open lockfile")?;
 
--- a/utils/scarb-build-metadata/build.rs
+++ b/utils/scarb-build-metadata/build.rs
@@ -52,7 +52,10 @@ fn cairo_version() {
         .root
         .expect("Expected metadata resolve root to be present.");
     assert!(
-        root.repr.starts_with("scarb "),
+        // The first condition for Rust >= 1.77
+        // (After the PackageId spec stabilization)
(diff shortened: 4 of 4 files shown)


scarb/src/flock.rs
```diff
@@ -149,6 +149,7 @@ pub struct AdvisoryLock<'f> {
 }
 
 #[derive(Debug)]
+#[allow(dead_code)]
```
mkaput, line 152: Why?

scarb/src/core/registry/client/local.rs
```diff
@@ -213,6 +213,7 @@ fn edit_records(records_path: &Path, func: impl FnOnce(&mut IndexRecords)) -> Re
         .read(true)
         .write(true)
         .create(true)
+        .truncate(false)
```
mkaput, line 216: Why?

scarb/src/ops/lockfile.rs
```diff
@@ -12,6 +12,7 @@ pub fn read_lockfile(ws: &Workspace<'_>) -> Result<Lockfile> {
         .read(true)
         .write(true)
         .create(true)
+        .truncate(false)
```
mkaput, line 15: Why?

utils/scarb-build-metadata/build.rs
```diff
@@ -52,7 +52,7 @@ fn cairo_version() {
         .root
         .expect("Expected metadata resolve root to be present.");
     assert!(
-        root.repr.starts_with("scarb "),
+        root.repr.contains("scarb#"),
```
mkaput, line 55: The change done this way effectively sets our MRSV to 1.77, so we should set appropriate field in our Cargo.tomls. 

@fracek are you able to use Rust 1.77 in Nix pkgs ?

mkaput (changes requested):

utils/scarb-build-metadata/build.rs
maciektr: Maybe we could just check for either `starts_with("scarb ")` or `contains("scarb#")`

scarb/src/core/registry/client/local.rs
maciektr:
```
warning: file opened with `create`, but `truncate` behavior not defined
   --> scarb/src/core/registry/client/local.rs:215:10
    |
215 |         .create(true)
    |          ^^^^^^^^^^^^- help: add: `.truncate(true)`
    |
    = help: if you intend to overwrite an existing file entirely, call `.truncate(true)`
    = help: if you instead know that you may want to keep some parts of the old file, call `.truncate(false)`
    = help: alternatively, use `.append(true)` to append to the file instead of overwriting it
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_open_options
    = note: `#[warn(clippy::suspicious_open_options)]` on by default
```

scarb/src/ops/lockfile.rs
maciektr: vide supra

scarb/src/flock.rs
maciektr:
```
warning: field `0` is never read
   --> scarb/src/flock.rs:152:30
    |
152 | pub struct AdvisoryLockGuard(Arc<FileLockGuard>);
    |            ----------------- ^^^^^^^^^^^^^^^^^^
    |            |
    |            field in this struct
    |
    = note: `AdvisoryLockGuard` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
    = note: `#[warn(dead_code)]` on by default
help: consider changing the field to be of unit type to suppress this warning while preserving the field numbering, or remove the field
    |
152 | pub struct AdvisoryLockGuard(());
    |                              ~~

```

utils/scarb-build-metadata/build.rs
maciektr: Done.

scarb/src/flock.rs
mkaput: let's drop `Debug` impl then perhaps?

mkaput (changes requested):

scarb/src/flock.rs
maciektr: Sure! If that's the only reason for this :D

mkaput:
```suggestion
```

mkaput (changes requested):

mkaput (approved):

PR #1224, opened 2024-03-22
Consider a case when the test fn return types are actually an empty vec

Arcticae:
Closes https://github.com/foundry-rs/starknet-foundry/issues/1934


maciektr (approved):
Please rebase

piotmag769 (approved):

PR #1244, opened 2024-04-04
Resolve full path markers

maciektr:
commit-id:80676f2f

---

**Stack**:
- #1249
- #1244 ⬅

### DIFF
--- a/scarb/src/compiler/plugin/proc_macro/host.rs
+++ b/scarb/src/compiler/plugin/proc_macro/host.rs
@@ -1,26 +1,36 @@
 use crate::compiler::plugin::proc_macro::{Expansion, FromSyntaxNode, ProcMacroInstance};
 use crate::core::{Config, Package, PackageId};
 use anyhow::{ensure, Result};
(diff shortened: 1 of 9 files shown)


scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -1,34 +1,45 @@
...
 use itertools::Itertools;
 use scarb_stable_hash::short_hash;
 use std::any::Any;
-use std::sync::Arc;
+use std::collections::HashMap;
+use std::sync::{Arc, RwLock};
 use std::vec::IntoIter;
 use tracing::{debug, trace_span};
 
+const FULL_PATH_MARKER_KEY: &str = "full_path_marker";
```
mkaput, line 33: What do you think about namespacing macro-specific attrs, like `starknet` does?

```suggestion
const FULL_PATH_MARKER_KEY: &str = "macro::full_path_marker";
```

mkaput (approved):

PR #1245, opened 2024-04-05
Read ui verbosity from env var

maciektr:


scarb-metadata/src/command/metadata_command.rs
```diff
@@ -61,7 +61,9 @@ impl MetadataCommand {
     /// Creates a default `scarb metadata` command, which will look for `scarb` in `$PATH` and
     /// for `Scarb.toml` in the current directory or its ancestors.
     pub fn new() -> Self {
-        Self::default()
+        let mut c = Self::default();
+        c.env_remove("SCARB_UI_VERBOSITY");
+        c
```
mkaput, line 66: Hmm, this looks like a hack. Wouldn't it be more correct if `scarb metadata` outputted metadata, no matter what verbosity is set?

extensions/scarb-cairo-run/src/main.rs
```diff
@@ -49,7 +50,10 @@ struct Args {
 }
 
 fn main() -> Result<()> {
-    let ui = Ui::new(Verbosity::default(), OutputFormat::Text);
+    let verbosity = env::var_os("SCARB_UI_VERBOSITY")
+        .and_then(|level| scarb_ui::Verbosity::from_str(&level.to_string_lossy()).ok())
+        .unwrap_or_default();
+    let ui = Ui::new(verbosity, OutputFormat::Text);
```
mkaput, line 56: IMHO, `scarb_ui` should provide ready-to-use clap arguments and the UI should be able to be constructed from them in a one-liner. Too much logic becomes smeared throughout different components.

extensions/scarb-cairo-run/tests/arguments.rs
```diff
@@ -335,3 +335,26 @@ fn cannot_set_gas_limit_for_package_with_disabled_gas_calculation() {
             error: gas calculation disabled for package `hello`, cannot define custom gas limit
         "#});
 }
+
+#[test]
+fn can_control_verbosity() {
```
mkaput, line 340: I like all tests added in this PR. Let's leave them even if we rewrite the underlying logic.

scarb/src/bin/scarb/args.rs
```diff
@@ -120,14 +121,24 @@ impl ScarbArgs {
 
     /// Get [`ui::Verbosity`] out of these arguments.
     pub fn ui_verbosity(&self) -> scarb_ui::Verbosity {
-        let filter = self.verbose.log_level_filter().as_trace();
-        if filter >= LevelFilter::WARN {
-            scarb_ui::Verbosity::Verbose
-        } else if filter > LevelFilter::OFF {
-            scarb_ui::Verbosity::Normal
+        let verbosity = if !self.verbose.is_present() {
```
mkaput, line 124: I think we reached a point then `clap_verbosity_flag` just doesn't work for us. If we handled this env within clap argument, we wouldn't need a lot of the logic here.

Perhaps we should just copy-paste `clap_verbosity_flag` into `scarb_ui` and edit it to operate on UI Verbosity and tracing LogLevels at once? And support control via environment variables.

mkaput (changes requested):
I think

mkaput (approved):

PR #1257, opened 2024-04-11
Document features

fmkra:


website/docs/reference/features.md
```diff
@@ -0,0 +1,60 @@
+# Features
+
+Features in Scarb provide a way to conditionally compile specific parts of the code during the build process.
+
+## `[features]` section
+
+In the `Scarb.toml` file, features are defined within the `[features]` section. Each feature lists other features that it enables.
```
maciektr, line 7:
```suggestion
A package defines a set of named features in the `[features]` section of `Scarb.toml` file. Each defined feature can list other features that should be enabled with it.
```

website/docs/reference/features.md
```diff
@@ -0,0 +1,60 @@
...
+For example, a package supporting various hash functions might define features like this:
+
+```toml
+[features]
+poseidon = []
+pedersen = []
+keccak = []
+```
+
+With these features set, conditional compilation expressions (`cfg` expressions) can selectively include code to support requested features during compile time. For instance:
```
maciektr, line 18:
```suggestion
With these features set, conditional compilation (`cfg`) attributes can be used to selectively include code to support requested features during compile time. For instance:
```

website/docs/reference/features.md
```diff
@@ -0,0 +1,60 @@
...
+
+```rust
+// Conditionally include a package
```
maciektr, line 21:
```suggestion
// Conditionally include a module
```

website/docs/reference/features.md
```diff
@@ -0,0 +1,60 @@
...
+
+```
+scarb build --features poseidon,pedersen
+```
+
+Enabling all features can be done with the `--all-features` flag.
+
+## `default` features
+
+By default, all features are disabled unless explicitly enabled with the `--features` flag. However, this behavior can be changed by specifying a default feature in the `[features]` section, like so:
```
maciektr, line 42:
```suggestion
By default, all features are disabled unless explicitly enabled. However, this behaviour can be changed by specifying a default feature in the `[features]` section, like so:
```

website/docs/reference/features.md
```diff
@@ -0,0 +1,60 @@
...
+
+```toml
+[features]
+default = ["poseidon", "pedersen"]
+poseidon = []
+pedersen = []
+keccak = []
+```
+
+During compilation, the compiler will enable the default feature, which in turn activates all listed features.
```
maciektr, line 52:
```suggestion
When the package is built, the default feature is enabled which in turn enables the listed features.
```

maciektr (commented):
🥰

website/docs/reference/features.md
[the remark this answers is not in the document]
fmkra: Yeah, maybe it's a good idea, because it isn't that big of a feature :) and wouldn't be used that much to have it's own "main" page. I'll move it there.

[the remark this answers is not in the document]
fmkra: By a sub section, you mean a separate page but grouped with Conditional Compilation (like Starknet pages in Core extensions) or just putting it on one page underneath the text of Conditional Compilation?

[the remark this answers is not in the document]
maciektr: Hi!
So sorry for late response!
The first one, or even on the same page.

fmkra:
I addressed everything, so I think we are ready to merge

PR #1268, opened 2024-04-15
Support diagnostic error codes

### LINKED ISSUE #1178: Print Cairo error codes
### Problem

In 2.7.0 Cairo will attach unique codes to various errors it emits (the `error_code` method in `DiagnosticEntry` trait). The `FormattedDiagnosticEntry` struct also stores this error code and its `Display` impl has a capability to print them. In Scarb though, we have our printing logic for errors (namely, to add term colors to `error`/`warning` prefixes) which we need to update.

### Proposed Solution

_No response_

### Notes

_No response_

maciektr:
fix #1178

### DIFF
--- a/utils/scarb-ui/src/components/typed.rs
+++ b/utils/scarb-ui/src/components/typed.rs
@@ -12,6 +12,9 @@ pub struct TypedMessage<'a> {
     r#type: &'a str,
     message: &'a str,
 
+    #[serde(skip_serializing_if = "Option::is_none")]
+    code: Option<&'a str>,
+
     #[serde(skip)]
     type_style: Option<&'a str>,
     #[serde(skip)]
@@ -26,6 +29,7 @@ impl<'a> TypedMessage<'a> {
             message,
             type_style: Some(type_style),
             skip_type_for_text: false,
+            code: None,
         }
     }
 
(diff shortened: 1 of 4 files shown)


utils/scarb-ui/src/lib.rs
```diff
@@ -119,6 +119,24 @@ impl Ui {
...
+
+    /// Print an error to the user.
+    pub fn error_with_code(&self, code: impl AsRef<str>, message: impl AsRef<str>) {
+        self.print(TypedMessage::styled(
+            format!("error[{}]", code.as_ref()).as_str(),
+            "red",
+            message.as_ref(),
+        ))
+    }
+
```
mkaput, line 139: This will print `error[CODE]` as message type in JSON output which is definitely not something expected

mkaput (changes requested):

utils/scarb-ui/src/lib.rs
```diff
@@ -119,6 +119,22 @@ impl Ui {
...
+        )
+    }
+
+    /// Print an error to the user.
+    pub fn error_with_code(&self, code: impl AsRef<str>, message: impl AsRef<str>) {
+        self.print(
+            TypedMessage::styled("error", "red", message.as_ref())
+                .with_type_suffix_for_text(format!("[{}]", code.as_ref()).as_str()),
+        )
+    }
```
mkaput, line 136: Hmm, what do you think about just having `error_code` field in the JSON output that exists if the code is known?

mkaput (changes requested):

utils/scarb-ui/src/components/typed.rs
```diff
@@ -43,8 +51,28 @@ impl<'a> TypedMessage<'a> {
...
         }
     }
+
+    /// Add additional key-value pair to the message.
+    ///
+    /// This will be exposed in the structured output.
+    pub fn with_extra(mut self, key: &'a str, value: &'a str) -> Self {
+        self.extra.insert(key, value);
+        self
+    }
```
mkaput, line 75: I'm afraid this API might be too powerful. 😕 It enables pushing information just for the human display, or just for the JSON-one, which we strive to avoid (not at all cost, though) Have you considered just storing the `error_code: &str` in `TypedMessage` directly?

mkaput (approved):

PR #1282, opened 2024-04-23
Fix year in footer

tomek0123456789:
😎


mkaput:
https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

PR #1283, opened 2024-04-23
Add option to choose function to run by name

### LINKED ISSUE #1063: feat: enable running arbitrary functions with the Scarb Runner
### Problem

The Scarb Runner is the wrapper around cairo that allows you to run Cairo functions and get the returned output. The program is hardcoded to run the `::main` function of the compiled sierra; however, allowing custom arguments specifying which function to run would be nice to develop Cairo tooling.

Here's how I plan to use it:
- I want to create a testing framework that allows you to unit tests your cairo functions with parametrized arguments, capture the output of the function run, and compare this output to a battle-tested implementaion of that behavior (diff testing)
- For that, I would like to use Scarb as an abstraction of the internals of the compiler and simply be able to invoke the runner directly in a rust program.
- The scarb-cairo-run extension only supports running the `::main` function. Maybe there could be an optional flag to specify the name of the function to run?
- If this approach is not correct, I'm down to discuss alternatives. The main things to keep in mind is that
  - we should be able to run the same functions multiple times, with different arguments, without re-building or re-loading files
  - we should be able to capture the output and return it to the caller
  
 

### Proposed Solution

_No response_

### Notes

_No response_

maciektr:
fix #1063

### DIFF
--- a/extensions/scarb-cairo-run/src/main.rs
+++ b/extensions/scarb-cairo-run/src/main.rs
@@ -27,6 +27,10 @@ struct Args {
     #[command(flatten)]
     packages_filter: PackagesFilter,
 
+    /// Specify name of the function to run.
+    #[arg(long)]
+    function: Option<String>,
+
     /// Maximum amount of gas available to the program.
     #[arg(long)]
     available_gas: Option<usize>,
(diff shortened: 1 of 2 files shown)


mkaput:
This fixes #1063, isn't it?

mkaput (approved):

mkaput (approved):

Issue #676, opened 2023-09-14
Support casm serialised output 


maciektr:
### Problem

As of now,  `lib` target compilation (as opposed to starknet contract targets) is capable of producing CASM artefacts in a textual, human readable format only (equivalent to `sierra-compile` compiler cli). In contrast, starknet contract compilation uses serialisable json format instead. It is possible to define a similar serialisable casm format for non-contract projects as well. This alternative should be supported as an artefact of `lib` target compilation. 

### Proposed solution

- Modify `lib` target properties to allow casm artefact format specification
    - `lib` target properties are defined here: https://github.com/software-mansion/scarb/blob/ba2c45574cf149a2d7422cb7c89aef5a203ec970/scarb/src/compiler/compilers/lib.rs#L18-L22
    - Analogously  to `sierra` artefact, it should allow `casm` and `casm_text` boolean inputs.
    - `casm` should create json serialized artefacts, while `casm-text` should create textual ones.
    - By default both should be disabled.
    - This properties will be available to users as `casm` and `casm-text` keys on `[lib]` object in `Scarb.toml` manifest files.
- Modify `lib` target compiler to produce new artefacts.
    - This compiler abstraction is implemented in this file: https://github.com/software-mansion/scarb/blob/ba2c45574cf149a2d7422cb7c89aef5a203ec970/scarb/src/compiler/compilers/lib.rs#L39
    - Should read properties from previous point and produce artefacts accordingly to user specification.
    - Artefact for serialised output should be written with `.casm.json` extension into the target directory with target name as filename.
    - Artefact for textual output should be written with `.casm` extension into the target directory, with the same name.
- Ensure new artefact type is produced
    - It should be enough to modify this test case https://github.com/software-mansion/scarb/blob/ba2c45574cf149a2d7422cb7c89aef5a203ec970/scarb/tests/build_targets.rs#L72

### Notes
See https://github.com/software-mansion/scarb/pull/637/files for similar change introducing new artefact format.

mkaput:
> * Analogously  to `sierra` artefact, it should allow `casm` and `casm_text` boolean inputs.
> * `casm` should create json serialized artefacts, while `casm-text` should create textual ones.
> * By default both should be disabled.
> * This properties will be available to users as `casm` and `casm-text` keys on `[lib]` object in `Scarb.toml` manifest files.

[...]

PR #1287, opened 2024-04-28
[DRAFT] add casm json target

tohrnii:
Addressing #676

### DIFF
--- a/scarb/src/compiler/compilers/lib.rs
+++ b/scarb/src/compiler/compilers/lib.rs
@@ -1,6 +1,6 @@
 use anyhow::{Context, Result};
-use cairo_lang_compiler::db::RootDatabase;
 use cairo_lang_casm::hints::Hint;
+use cairo_lang_compiler::db::RootDatabase;
 use cairo_lang_sierra::program::VersionedProgram;
 use cairo_lang_sierra_to_casm::compiler::SierraToCasmConfig;
 use cairo_lang_sierra_to_casm::metadata::{calc_metadata, calc_metadata_ap_change_only};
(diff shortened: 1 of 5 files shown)


scarb/src/compiler/compilers/lib.rs
```diff
@@ -127,3 +149,16 @@ impl Compiler for LibCompiler {
         Ok(())
     }
 }
+
+/// Represents a contract in the Starknet network.
+#[derive(Clone, Default, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct SerializedCasm {
```
maciektr, line 155: Why don't we just define some serialization for the struct in Cairo repo (in that repo)? https://github.com/starkware-libs/cairo/blob/233c0fdabe8c5010d10b5d16e02fecf6b656ef6b/crates/cairo-lang-sierra-to-casm/src/compiler.rs#L95

maciektr (commented):
Hi!

scarb/src/compiler/compilers/lib.rs
tohrnii: You're right it should ideally go there but I wasn't sure if the cairo team wanted to make this a part of that repo based on some discussion in related issues. I'll open a PR there.

maciektr:
Now that I think about it, please consider if your usage isn't already covered by a tool called `universal-sierra-compiler` https://github.com/software-mansion/universal-sierra-compiler/tree/master, which is already capable of compiling Sierra json to Casm json format (and has not been available when the issue has been created).

tohrnii:
@maciektr Thank you. That tool should already cover the use case. I'll close this PR.

tohrnii:
@maciektr After discussion with @rodrigo-pino, there are a couple things missing from the generated casm json, namely entrypoint and builtins. It might be better for scarb to produce the json file directly instead of relying on yet another tool since it already generates the sierra json file. What do you think?

Arcticae:
@tohrnii It can be added to the `universal-sierra-compiler` if you need it. Happy to accept any PRs there :)

maciektr:
Hi @tohrnii!

Very sorry for the delay in response, we had a national holidays here in Poland, so I've been out of office for a bit. 

> After discussion with @rodrigo-pino, there are a couple things missing from the generated casm json by the universal-sierra-compiler, namely entrypoint and builtins. 

I believe this should be added to USC then. I second @Arcticae here. 

> It might be better for scarb to produce the json file directly instead of relying on yet another tool since it already generates the sierra json file. 

We do see Scarb as primarily Cairo -> Sierra compiler, and the CASM is just an addition here. We recommend using the USC wherever you rely on CASM programatically and the builtin Scarb feature for by-hand debuging.  We move our tooling to use USC as well. 

Would it be ok for you?

tohrnii:
Thank you @maciektr @Arcticae for the confirmation. Sounds good, I'll open a PR there. Closing this one then.

PR #1295, opened 2024-05-06
Add warning on no-executables & no-debug lib target build

maciektr:
commit-id:38745386

---

**Stack**:
- #1296
- #1295 ⬅

### DIFF
--- a/scarb/src/compiler/compilers/lib.rs
+++ b/scarb/src/compiler/compilers/lib.rs
@@ -1,8 +1,11 @@
 use anyhow::{Context, Result};
 use cairo_lang_compiler::db::RootDatabase;
+use cairo_lang_compiler::CompilerConfig;
+use cairo_lang_defs::db::DefsGroup;
 use cairo_lang_sierra::program::VersionedProgram;
 use cairo_lang_sierra_to_casm::compiler::SierraToCasmConfig;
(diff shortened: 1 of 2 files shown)


scarb/src/compiler/compilers/lib.rs
```diff
@@ -127,3 +131,25 @@ impl Compiler for LibCompiler {
...
+    // This does not apply to debug build (expressed by `replace_ids` flag),
+    // which is a goal by itself.
+    // See starkware-libs/cairo#5440 for more context.
+    let executable_plugin = db
+        .macro_plugins()
+        .iter()
+        .any(|plugin| !plugin.executable_attributes().is_empty());
+    if !executable_plugin && !compiler_config.replace_ids {
+        ws.config().ui().warn(
+            "no executable plugin found, and `replace_ids` is disabled, \
```
mkaput, line 151: As a Cairo code developer who has no knowledge of compiler/scarb internals (i.e. I only know I can write some code and test it with `snforge`), I would have no idea what this message is about 😕 

The term `executable plugin` is vague to the user. `replace_ids` also would probably be because by itself it is not stating that this is about compiler configuration flag.

This warning is also missing information how to fix it. It'd be best to include several notes that will suggest code snippets to put in Scarb.toml, or tell what commands user might want to actually run.

mkaput (changes requested):

mkaput (approved):
😍

PR #1304, opened 2024-05-13
[scarb-doc] Fix `trait_item_names`

drknzz:
- use cairo at [290f51f554c978180ca9d91177423305959cbed5](https://github.com/starkware-libs/cairo/commit/290f51f554c978180ca9d91177423305959cbed5)
- trait_item_names -> trait_required_item_names

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,28 +36,28 @@ assert_fs = "1"
 async-trait = "0.1"
 axum = { version = "0.6", features = ["http2"] }
 cairo-felt = "0.9.1"
-cairo-lang-casm = { git = "https://github.com/starkware-libs/cairo", rev = "b60791d93a367d570535757cb221b4d114a31183" }
-cairo-lang-compiler = { git = "https://github.com/starkware-libs/cairo", rev = "b60791d93a367d570535757cb221b4d114a31183" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -119,8 +119,8 @@ toml = "0.8"
 toml_edit = { version = "0.22", features = ["serde"] }
 tower-http = { version = "0.4", features = ["fs"] }
 tracing = "0.1"
-tracing-log = "0.2"
 tracing-core = "0.1"
+tracing-log = "0.2"
```
maciektr, line 123: why?

drknzz: It just got moved one line below D:

drknzz: I've used this
```
cargo xtask set-cairo-version --rev 290f51f554c978180ca9d91177423305959cbed5
```

drknzz: It sorts the deps from what I see
https://github.com/software-mansion/scarb/blob/110826a7abf9b61d6dcee7da38a1538fb1298470/xtask/src/set_cairo_version.rs#L79

maciektr (approved):

Cargo.toml
maciektr: Cool, we must have missed it then!

PR #1308, opened 2024-05-13
Add forge template to `scarb new` and `scarb init` under separate flag

### LINKED ISSUE #1208: Add forge template to `scarb new` under separate flag.
### Problem

As in title

### Proposed Solution

_No response_

### Notes

_No response_

Draggu:
Closes  #1208

- Added `--snforge` flag that enables forge template for
   - `new` command
   - `init` command
   
  On forge side https://github.com/foundry-rs/starknet-foundry/pull/2113

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -241,6 +241,10 @@ pub struct InitArgs {
     /// Do not initialize a new Git repository.
     #[arg(long)]
     pub no_vcs: bool,
+
+    /// Initialize snforge setup.
+    #[arg(long)]
+    pub snforge: bool,
 }
 
 /// Arguments accepted by the `metadata` command.
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -242,7 +242,7 @@ pub struct InitArgs {
     #[arg(long)]
     pub no_vcs: bool,
 
-    /// Initialize snforge setup.
+    /// Use a Starknet Foundry Forge template.
     #[arg(long)]
     pub snforge: bool,
 }
--- a/scarb/src/ops/new.rs
+++ b/scarb/src/ops/new.rs
@@ -211,7 +211,7 @@ fn init_snforge(name: PackageName, target_dir: Utf8PathBuf, config: &Config) -> 
     let mut process = Command::new("snforge")
         .arg("init")
         .arg(name.as_str())
-        .current_dir(target_dir.canonicalize_utf8()?)
+        .current_dir(fsx::canonicalize_utf8(&target_dir)?)
         .envs(get_env_vars(config, Some(target_dir))?)
         .stderr(Stdio::inherit())
         .stdout(Stdio::inherit())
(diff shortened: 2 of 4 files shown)


scarb/src/bin/scarb/args.rs
```diff
@@ -241,6 +241,10 @@ pub struct InitArgs {
     /// Do not initialize a new Git repository.
     #[arg(long)]
     pub no_vcs: bool,
+
+    /// Initialize snforge setup.
```
mkaput, line 245:
```suggestion
    /// Use a Starknet Foundry Forge template.
```

image

scarb/src/bin/scarb/args.rs
```diff
@@ -241,6 +241,10 @@ pub struct InitArgs {
     /// Do not initialize a new Git repository.
     #[arg(long)]
     pub no_vcs: bool,
```
maciektr, line 243: This args would be ignored, right?

scarb/src/ops/new.rs
```diff
@@ -20,6 +21,7 @@ pub struct InitOptions {
     pub path: Utf8PathBuf,
     pub name: Option<PackageName>,
     pub vcs: VersionControl,
```
maciektr, line 23: `name` and `vcs` args would be ignored?

scarb/src/ops/new.rs
```diff
@@ -193,6 +199,19 @@ fn mk(
         "#})
     }
 
+    if snforge {
+        init_snforge(name)?;
```
maciektr, line 203: I am not sure if we want to both create the package here with Scarb and then with Snforge. Will Snforge overwrite what has been created by Scarb? It does call Scarb `new` once again under the hood, is it ok?

scarb/src/ops/new.rs
```diff
@@ -193,6 +199,19 @@ fn mk(
...
+    }
+
+    Ok(())
+}
+
+fn init_snforge(name: PackageName) -> Result<()> {
+    Command::new("snforge")
```
maciektr, line 210: Shouldn't this inherit stdout/stderr? 
Shouldn't this get all env vars usually set by Scarb for subcommands (like `SCARB` so snforge does not call any other Scarb binary) - see https://github.com/software-mansion/scarb/blob/61dd36af1d28d60499832f06b3a5c9e83ba7647a/scarb/src/subcommands.rs#L11

scarb/src/ops/new.rs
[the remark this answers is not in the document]
Draggu: https://github.com/foundry-rs/starknet-foundry/pull/2113

scarb/src/ops/new.rs
Draggu: Changed forge behaviour in https://github.com/foundry-rs/starknet-foundry/pull/2113

scarb/src/ops/new.rs
Draggu: `name` is used and `vcs` is ignored by forge

mkaput (approved):

scarb/src/ops/new.rs
maciektr: If we allow to use both `--no-vcs` and `--snforge` simultaneously - we should support it.

mkaput: I believe snforge is NOT touching vcs stuff; and this code is handling VCS by itself

maciektr: Ok, cool, makes sense!

maciektr (approved):
I am not sure about the logic on snforge side (expressed in comment in the forge repo). 

This side look good to go!

PR #1311, opened 2024-05-21
Add `group_id` to target props

maciektr:
commit-id:7964f21f

---

**Stack**:
- #1313
- #1312
- #1311 ⬅

### DIFF
--- a/scarb/src/core/manifest/toml_manifest.rs
+++ b/scarb/src/core/manifest/toml_manifest.rs
@@ -592,6 +592,7 @@ impl TomlManifest {
             self.lib.as_ref(),
             &package_name,
             root,
+            None,
         )?);
 
         targets.extend(Self::collect_target(
@@ -599,6 +600,7 @@ impl TomlManifest {
             self.cairo_plugin.as_ref(),
             &package_name,
             root,
+            None,
         )?);
 
         for (kind, ext_toml) in self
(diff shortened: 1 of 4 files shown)


scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -280,6 +280,7 @@ pub struct DetailedTomlDependency {
 pub struct TomlTarget<P> {
     pub name: Option<SmolStr>,
     pub source_path: Option<Utf8PathBuf>,
+    pub group_id: Option<SmolStr>,
```
mkaput, line 283: Hmm I cannot imagine how we should deal with users manually specifying `group_id` in their tomls. Do we have to expose this property here and not just make it a private thing?

maciektr: Hmm, I don't see why this would be a problem? If users asks us to compile a set of targets simultaneously and this is possible, then why not? 🤔

mkaput: 1. I'm just afraid of making this "optimization" an actual API
2. I do not recall a check that valid `lib.cairo` is actually constructible for the group. Or does it exist somewhere?

maciektr: 2. It should be validated in `CompilationUnitComponent::try_new`

maciektr: 1. Ok,  I will disallow this then

mkaput (approved):

PR #1315, opened 2024-05-21
[scarb-doc] Inner attributes

drknzz:


extensions/scarb-doc/tests/test.rs
```diff
@@ -91,24 +106,46 @@ fn test_main() {
         .success();
     let stdout = std::str::from_utf8(&output.get_output().stdout).unwrap();
 
+    println!("{:?}", &stdout);
```
Arcticae, line 109:
```suggestion
```

extensions/scarb-doc/src/types.rs
```diff
@@ -135,22 +138,39 @@ impl Module {
...
+            text_without_trivia: node.lookup(db).as_syntax_node().get_text_without_trivia(db),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct LookupItemData {
+    pub doc: Option<String>,
+    pub definition: String,
+}
```
Arcticae, line 167: Why is a separate struct needed here?

extensions/scarb-doc/src/types.rs
```diff
@@ -214,15 +252,51 @@ pub struct Struct {
...
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Member {
+    pub id: MemberId,
+    pub node: ast::MemberPtr,
+
+    pub item_data: ItemData,
+    // TODO: no LookupItemId enum variant for struct / enum items
```
Arcticae, line 289: I kind of worry that we'll lose the track of those

extensions/scarb-doc/src/types.rs
```diff
@@ -286,15 +408,118 @@ pub struct Trait {
...
+}
+
+#[derive(Clone, Debug)]
+pub struct TraitConstant {
+    pub id: TraitConstantId,
+    pub node: ast::TraitItemConstantPtr,
+
+    pub item_data: ItemData,
+    pub lookup_item_data: LookupItemData,
+}
```
Arcticae, line 462: Why this can't be a regular constant? 🤔

Arcticae: I guess this applies to all of the nested things inside of the traits

extensions/scarb-doc/src/types.rs
```diff
@@ -304,15 +529,83 @@ pub struct Impl {
...
+
+#[derive(Clone, Debug)]
+pub struct ImplFunction {
+    pub id: ImplFunctionId,
+    pub node: ast::FunctionWithBodyPtr,
+
+    pub item_data: ItemData,
+    pub lookup_item_data: LookupItemData,
+}
```
Arcticae, line 596: This also seems like something that could be generalised by a function type

extensions/scarb-doc/src/types.rs
piotmag769: Fixed

extensions/scarb-doc/src/types.rs
piotmag769: Cairo stuff I guess

extensions/scarb-doc/src/types.rs
piotmag769: Depends if we want to keep it isomorphic to structs from compiler or not

extensions/scarb-doc/src/types.rs
piotmag769: Will remove, good point

extensions/scarb-doc/src/types.rs
Arcticae: What do you think?

piotmag769: I would keep it this way for now, if we don't find any use case for that we can merge it

piotmag769:
@Arcticae I will open a separate PR with new database for `scarb-doc`, merge it first into main and then rebase this one on top of main. It is due to Cairo already being bumped on main

extensions/scarb-doc/src/types.rs
```diff
@@ -184,12 +224,19 @@ pub struct Use {
     pub item_data: ItemData,
 }
 
+// TODO: do we even want to document this?
```
piotmag769, line 227: @Arcticae

Cargo.toml
```diff
@@ -136,7 +136,7 @@ zip = { version = "0.6", default-features = false, features = ["deflate"] }
 zstd = "0.13"
 
 [profile.release]
-lto = true
```
mkaput, line 139: why

piotmag769: Daily me forgetting about this xD

extensions/scarb-doc/src/types.rs
Arcticae: Not really, wouldn't document it

Arcticae (approved):
LGTM

PR #1323, opened 2024-05-27
Add derives for `Severity` and `Diagnostics`

Draggu:
Add derives for `Diagnostic`, `Severity` and `Diagnostics`, also `Deref` for `Diagnostics`. In current shape these are very hard to work with, which is caused by lack of basic functionalities (like `Eq`), also `Diagnostics` can't be merged or extended with ready `Diagnostic` (only by matching on it's `Severity` and then choosing correct method to reconstruct in again)

### DIFF
--- a/plugins/cairo-lang-macro/src/types/mod.rs
+++ b/plugins/cairo-lang-macro/src/types/mod.rs
@@ -5,7 +5,7 @@ mod conversion;
 mod expansions;
 
 pub use expansions::*;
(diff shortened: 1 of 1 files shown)


maciektr (approved):

mkaput (changes requested):
What do you need this for?

mkaput (approved):
Lifting my merge ban as my voice has been heard

PR #1337, opened 2024-06-04
Print Cargo logs with SpinnerBar::println

### LINKED ISSUE #1336: Output of cargo build for macros is going wild together with the command indicator
### Problem

See the screenshot in _Steps_.

The indicator is being constantly re-displayed in separate lines, instead of being interactively pinned to the bottom of the terminal.

### Steps

Tested here: https://github.com/dojoengine/dojo/tree/main/examples/spawn-and-move

Set `asdf local scarb nightly-2024-06-01`

image

### Possible Solution(s)

Sounds like `cargo build` output is not routed through https://docs.rs/indicatif/latest/indicatif/struct.ProgressBar.html#method.println?

### Notes

_No response_

### Version

```text
`nightly-2024-06-01`
```

maciektr:
fix #1336

### DIFF
--- a/utils/scarb-ui/src/components/spinner.rs
+++ b/utils/scarb-ui/src/components/spinner.rs
@@ -1,6 +1,6 @@
 use std::time::Duration;
 
-use indicatif::{ProgressBar, ProgressStyle};
+use indicatif::{ProgressBar, ProgressDrawTarget, ProgressStyle};
 
 use crate::Widget;
 
@@ -37,7 +37,7 @@ impl Widget for Spinner {
     type Handle = SpinnerHandle;
 
     fn text(self) -> Self::Handle {
-        let pb = ProgressBar::new_spinner()
+        let pb = ProgressBar::with_draw_target(None, ProgressDrawTarget::stdout())
             .with_style(Spinner::default_style())
             .with_message(self.message);
         pb.enable_steady_tick(Duration::from_millis(120));
--- a/utils/scarb-ui/src/components/spinner.rs
+++ b/utils/scarb-ui/src/components/spinner.rs
@@ -1,8 +1,8 @@
 use std::time::Duration;
 
-use indicatif::{ProgressBar, ProgressDrawTarget, ProgressStyle};
+use indicatif::{ProgressBar, ProgressDrawTarget, ProgressStyle, WeakProgressBar};
 
-use crate::Widget;
+use crate::{Widget, WidgetHandle};
 
 /// Spinner widget informing about an ongoing process.
 pub struct Spinner {
@@ -30,6 +30,12 @@ pub struct SpinnerHandle {
 impl Drop for SpinnerHandle {
     fn drop(&mut self) {
         self.pb.finish_and_clear()
+    }
+}
+
+impl WidgetHandle for SpinnerHandle {
+    fn weak_progress_bar(&self) -> Option<WeakProgressBar> {
+        Some(self.pb.downgrade())
     }
 }
 
--- a/utils/scarb-ui/src/lib.rs
+++ b/utils/scarb-ui/src/lib.rs
@@ -22,10 +22,13 @@
 #![warn(rust_2018_idioms)]
 
 use clap::ValueEnum;
+use indicatif::WeakProgressBar;
 pub use indicatif::{
     BinaryBytes, DecimalBytes, FormattedDuration, HumanBytes, HumanCount, HumanDuration,
     HumanFloatCount,
 };
+use std::fmt::Debug;
+use std::sync::{Arc, RwLock};
 
 pub use message::*;
 pub use verbosity::*;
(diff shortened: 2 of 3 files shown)


mkaput (commented):
Shouldn't this be, like, more intelligent?

If there is an active spinner handle:
1. Do not allow creating other spinners
2. Redirect all `Ui::print` & co. calls to this spinner's `println`

utils/scarb-ui/src/lib.rs
```diff
@@ -149,7 +179,28 @@ impl Ui {
...
+                            // The driver is inactive. Print has failed.
+                            message.print_text();
+                            // Clear widget from state.
+                            drop(read_lock);
+                            self.state
+                                .write()
+                                .expect("failed to acquire write lock")
+                                .active_print_overrider = None;
+                        }
+                    }
```
mkaput, line 199: this is weird

make the `Weak` should be simply made part of public API? because you are simply creating an abstraction over abstraction that could be wrong in the first place

utils/scarb-ui/src/message.rs
```diff
@@ -16,7 +16,7 @@ pub trait Message {
     /// Return textual representation of this message.
     ///
     /// Default implementation returns empty string, making [`Ui`] skip printing this message.
-    fn text(self) -> String
+    fn text(&self) -> String
```
mkaput, line 19: this kinda contradicts the idea of messages being one-off. I know I forgot to document this principle, sorry for that

mkaput (changes requested):

scarb/tests/build_cairo_plugin.rs
mkaput, line 1: why did you make changes in this test?

utils/scarb-ui/src/components/spinner.rs
```diff
@@ -33,13 +34,27 @@ impl Drop for SpinnerHandle {
     }
 }
 
+impl PrintOverrider for ProgressBar {
```
mkaput, line 37: it's a tricky game to abuse submodule relationships, to workaround orphan rule, to write a trait impl that is a public API

was something off with the private newtype that provided this impl?

utils/scarb-ui/src/widget.rs
```diff
@@ -1,9 +1,25 @@
...
+
+/// A handle for a widget that allows for interaction with it.
+pub trait WidgetHandle {
+    /// Get the print driver for the widget.
+    fn print_driver(&self) -> Option<Weak<dyn PrintOverrider>> {
+        None
+    }
+}
+
+/// A trait for providing the print functionality of a widget.
```
mkaput, line 21: this comment does not describe what this is for

utils/scarb-ui/src/lib.rs
```diff
@@ -53,10 +55,28 @@ pub enum OutputFormat {
...
+            .field("output_format", &self.output_format)
+            .finish()
+    }
+}
+
+/// An encapsulation of the UI state.
+///
+/// This can be used by `Ui` to store stateful information.
+#[derive(Default)]
+pub struct State {
```
mkaput, line 78: why is this pub?

utils/scarb-ui/src/lib.rs
```diff
@@ -65,6 +85,9 @@ impl Ui {
         Self {
             verbosity,
             output_format,
+            state: Arc::new(RwLock::new(State {
+                active_print_overrider: None,
+            })),
```
mkaput, line 90:
```suggestion
            state: Default::default(),
```

utils/scarb-ui/src/lib.rs
```diff
@@ -149,7 +179,24 @@ impl Ui {
...
+                        // Clear widget from state.
+                        drop(read_lock);
+                        self.state
+                            .write()
+                            .expect("failed to acquire write lock")
+                            .active_print_overrider = None;
+                    }
+                } else {
+                    message.print_text()
+                }
```
mkaput, line 198: I sense several issues here:
1. This seems to hold the read lock for too long.
2. This write seems unnecessary? If widget dies, then the weak will always fail to upgrade so this looks to be effectively no-op

FYI Take a look at this: https://gcc.godbolt.org/z/rfj7MEG3s. It looks like, in hot path and aggressive optimizations, `Weak::upgrade` is almost the same cost-wise (here is even better -- fewer instructions) as matching over `Option`.

Heck, you can even replace `Option<Weak<...>>` with just `Weak` and use `Weak::new()` as `None` replacement.

mkaput (changes requested):

scarb/tests/build_cairo_plugin.rs
maciektr: indicatif::ProgressBar writes to stderr by default

mkaput: you mean, that now cargo output is pushed to stderr? then this needs fixing ?

PR #1344, opened 2024-06-06
Scarb expand initial impl

maciektr:
commit-id:8273b556

---

**Stack**:
- #1348
- #1347
- #1344 ⬅

### DIFF
--- a/scarb/src/ops/expand.rs
+++ b/scarb/src/ops/expand.rs
@@ -1,0 +1,214 @@
+use crate::compiler::db::{build_scarb_root_database, ScarbDatabase};
+use crate::compiler::helpers::{build_compiler_config, write_string};
+use crate::compiler::{CairoCompilationUnit, CompilationUnit, CompilationUnitAttributes};
+use crate::core::{Package, TargetKind, Workspace};
+use crate::ops;
+use crate::ops::FeaturesOpts;
+use anyhow::{anyhow, bail, Context, Result};
+use cairo_lang_compiler::db::RootDatabase;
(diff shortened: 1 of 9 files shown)


scarb/src/ops/expand.rs
```diff
@@ -0,0 +1,214 @@
...
+        .diagnostics_reporter
+        .ensure(&db)
+        .map_err(|err| err.into())
+        .map_err(|err| {
+            if !suppress_error(&err) {
+                ws.config().ui().anyhow(&err);
+            }
+
+            anyhow!("could not check `{package_name}` due to previous error")
+        })?;
```
mkaput, line 62: LS will need to expand macros on potentially erroring projects

scarb/src/ops/expand.rs
```diff
@@ -0,0 +1,214 @@
...
+    do_expand(&db, &compilation_unit, opts, ws)?;
+
+    Ok(())
+}
+
+/// Memorize opened modules for adding appropriate bracketing to the code.
+struct ModuleStack(Vec<String>);
+
+impl ModuleStack {
+    pub fn new() -> Self {
```
mkaput, line 73: If type is private, then you do not need to make its methods `pub`

scarb/src/ops/expand.rs
mkaput, line 1: any tests?

THenry14 (approved):
lgtm

scarb/src/ops/expand.rs
maciektr: I will change it in subsequent PR.

PR #1347, opened 2024-06-11
Add target selector

maciektr:
commit-id:2ec662b7

---

**Stack**:
- #1348
- #1347 ⬅

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -228,6 +228,14 @@ pub struct ExpandArgs {
     #[command(flatten)]
     pub features: FeaturesSpec,
 
+    /// Specify the target to expand by target kind.
+    #[arg(long)]
+    pub target_kind: Option<String>,
+
+    /// Specify the target to expand by target name.
+    #[arg(long)]
+    pub target_name: Option<String>,
+
     /// Do not attempt formatting.
     #[arg(long, default_value_t = false)]
     pub ugly: bool,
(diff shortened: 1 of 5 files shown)


mkaput (approved):

scarb/src/bin/scarb/args.rs
```diff
@@ -228,6 +228,14 @@ pub struct ExpandArgs {
...
     pub features: FeaturesSpec,
 
+    /// Specify the target to expand by target kind.
+    #[arg(long)]
+    pub target_kind: Option<String>,
+
+    /// Specify the target to expand by target name.
+    #[arg(long)]
+    pub target_name: Option<String>,
+
```
THenry14, line 238: noob question: is it possible someone sets a target_name and a target_kind that are incompatible with each other?

THenry14 (approved):
lgtm, left just one question but since it may come from my lack of understanding of the project (yet!) this is rather a nitpick

scarb/src/bin/scarb/args.rs
THenry14: Ah ok, so setting kind will result in extending all applicable targets of this kind, right? whereas setting the name will result in extending only this specific target? 

Makes sense, thanks for clarification :+1:

maciektr: > setting kind will result in extending all applicable targets of this kind

Not really, right now this would be the first one that matches.

maciektr: discussed offline

PR #1352, opened 2024-06-11
:arrow_up: Upgrading Cairo to 2.6.4

florian-bellotti:


maciektr:
We are in the process of releasing Scarb from https://github.com/software-mansion/scarb/releases/tag/v2.6.5

PR #1353, opened 2024-06-11
Merge tools definition recursively

maciektr:


scarb/src/internal/serdex.rs
```diff
@@ -1,9 +1,73 @@
...
+            "merge" => Ok(Self::Merge),
+            _ => Err(anyhow!(
+                "invalid merge strategy: {}, must be one of `merge`, `override`",
+                value
+            )),
+        }
+    }
+}
+
+pub fn toml_merge_apply_strategy<'de, T, S>(target: &T, source: &S) -> Result<T>
```
mkaput, line 30: doc, you can use excerpts from the design doc

scarb/src/internal/serdex.rs
```diff
@@ -1,9 +1,73 @@
...
+#[derive(Debug, Default)]
+pub enum MergeStrategy {
+    #[default]
+    Override,
+    Merge,
+}
+
+impl TryFrom<&str> for MergeStrategy {
+    type Error = anyhow::Error;
+    fn try_from(value: &str) -> std::result::Result<Self, Self::Error> {
```
mkaput, line 18:
```suggestion
    fn try_from(value: &str) -> Result<Self, Self::Error> {
```

you can do this with `anyhow::Result` as well

mkaput (changes requested):
please do not forget writing docs on scarb website

scarb/src/internal/serdex.rs
maciektr: Sure, I will add it as soon as it will be accepted :)

mkaput: I meant, document this function :)

THenry14, line 30: great idea! 🔥 
I wonder though if this function could be refactored a bit, I'm generally not a fan of doing multiple nested loops. Maybe in order to improve readability and maintainability you could do something like (please treat it as a pseudocode as I have not tested this):

```
    let mut target_value = Value::try_from(target)?;
    let source_value = Value::try_from(source)?;
    
    if let (Some(target_table), Some(source_table)) = (target_value.as_table_mut(), source_value.as_table()) {
        for (key, value) in source_table {
            match target_table.get_mut(key) {
                Some(target_subvalue) if target_subvalue.is_table() && value.is_table() => {
                    let target_subtable = target_subvalue.as_table_mut().unwrap();
                    let value_subtable = value.as_table().unwrap();

                    let strategy = value_subtable.get(MERGE_STRATEGY_KEY)
                        .and_then(|v| v.as_str())
                        .map_or(Ok(MergeStrategy::default()), MergeStrategy::try_from)?;

                    match strategy {
                        MergeStrategy::Override => {
                            *target_subvalue = value_subtable.clone();
                        }
                        MergeStrategy::Merge => {
                            *target_subvalue = toml_merge_apply_strategy(target_subtable, value_subtable)?;
                        }
                    }
                }
                _ => {
                    target_table.insert(key.clone(), value.clone());
                }
            }
        }
    }

    Ok(toml::Value::try_into(target_value)?)
```
So we could get rid of a few ifs there, and flatten the code a bit. Let me know what you think, though you may treat this comment as a nitpick, as I'm not sure if this fits in your (scarb) code-style yet 😅

maciektr: Surprisingly, making stuff simpler instead of more complicated does fit the Scarb code-style 🤣

THenry14: Alt Text

maciektr: Yep, I meant it too :D

scarb/src/internal/serdex.rs
```diff
@@ -1,11 +1,92 @@
...
+/// If `source` and `target` are tables and a specific key exists in both, the conflict will be
+/// resolved as follows:
+/// 1) If value under the conflicting key in either `source` or `target` is not a table, the value
+/// in `target` will be overridden with the value from `source`.
+/// 2) If values under the conflicting key in both `source` and `target` are tables,
+/// the conflict will be resolved with one of two strategies defined by `source`:
+///    a) If `source` have a key `merge-strategy` with value `override`, the value in `target` will
+///     be overridden with the value from `source`.
+///    b) If `source` have a key `merge-strategy` with value `merge`, the value in `target` will be
+///     merged with the value from `source` recursively with this function.
```
THenry14, line 41: nit: maybe it would be a good idea to be more clear on the difference between merge/override?
eg it is not clear to me that when I use override, I will need to pass values for all the keys from source (otherwise they are lost)

THenry14 (approved):
lgtm

mkaput (approved):

PR #1364, opened 2024-06-17
Warn if cairo-test plugin is missing

maciektr:
commit-id:9ff328fc

### DIFF
--- a/extensions/scarb-cairo-test/src/main.rs
+++ b/extensions/scarb-cairo-test/src/main.rs
@@ -7,6 +7,7 @@ use cairo_lang_test_plugin::{TestCompilation, TestCompilationMetadata};
 use cairo_lang_test_runner::{CompiledTestRunner, RunProfilerConfig, TestRunConfig};
 use camino::Utf8PathBuf;
 use clap::Parser;
+use indoc::formatdoc;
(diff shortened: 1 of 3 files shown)


extensions/scarb-cairo-test/src/main.rs
```diff
@@ -43,6 +44,7 @@ fn main() -> Result<()> {
     let metadata = MetadataCommand::new().inherit_stderr().exec()?;
 
     check_scarb_version(&metadata);
+    check_cairo_test_plugin(&metadata);
```
THenry14, line 47: when will this be shown?
I suppose it should only when someone tries to execute a test and does not have any plugin configured? or what is the deal?

maciektr: Only if you run `cairo-test` (this is the main file of the cairo-test binary distributed with Scarb) and you miss the plugin.

THenry14: aaaah, sorry misread the file name. thanks for pointing this out :+1:

THenry14 (approved):

piotmag769 (commented):
Gj

PR #1371, opened 2024-06-18
Add code mappings to plugin generated code

maciektr:


mkaput (approved):
nothing looking suspicious, though a test would be very welcome

PR #1372, opened 2024-06-19
Add cairo-test plugin to Scarb.toml manifest on new

maciektr:


THenry14 (commented):
What is the scenario when this dep is added? Does that mean that cairo-test stays as the default test plugin in scarb? Sorry for potentially dumb questions, jsut trying to understand the impact :)

maciektr:
@THenry14 
If you create a new package and run `scarb test` inside, it will try to run tests with `cairo-test`. As of now it will also fail to do so, as there is no `cairo_test` dependency in a newly created package.

THenry14 (approved):

PR #1390, opened 2024-06-21
Test asserts plugin

maciektr:


scarb/scarblib/cairo_test_asserts/Scarb.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "cairo_test_macros"
```
Draggu, line 2: Maybe we can name it `assert_macros`? This can be used not only in test, also there is nothing more than `assert_*!` here

Draggu (approved):

scarb/scarblib/cairo_test_asserts/Scarb.toml
Arcticae: +1

Arcticae (approved):

THenry14 (approved):

PR #1396, opened 2024-06-26
Doc cairo-run plugin

maciektr:
commit-id:f681b0b9

### DIFF
--- a/website/docs/extensions/cairo-run.md
+++ b/website/docs/extensions/cairo-run.md
@@ -1,6 +1,10 @@
+<script setup>
+import { data as rel } from "../../github.data";
+</script>
+
 # Using cairo-run in Scarb projects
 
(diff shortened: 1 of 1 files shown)


website/docs/extensions/cairo-run.md
```diff
@@ -9,6 +13,32 @@ If not provided, a gas usage is not limited.
 Gas usage can be disallowed by setting the value to `0`.
 The second one prints the full memory after the execution.
 
+## Choosing a function to run
+
+In general, a function to run can be specified in two ways:
+
+- by the name of the function
```
THenry14, line 20: nit: I wonder if we should include how to specify this name in this place, to have a nice top-level tldr

THenry14 (approved):

PR #1400, opened 2024-06-27
Cairo-test target type argument cli

maciektr:


extensions/scarb-cairo-test/src/main.rs
```diff
@@ -48,10 +70,30 @@ fn main() -> Result<()> {
...
+                })
+                .map(|t| t.name.clone())
+                .collect::<Vec<_>>()
+        })
+        .collect::<Vec<_>>();
 
     ScarbCommand::new()
         .arg("build")
         .arg("--test")
+        .env("SCARB_TARGET_NAMES", target_names.clone().join(","))
```
THenry14, line 96: nitpick: Is this going to be documented in different PR? or should it be done in this one as well?

THenry14 (approved):
💪

extensions/scarb-cairo-test/src/main.rs
maciektr: What kind of docs are you thinking?

THenry14: I assume https://docs.swmansion.com/scarb/docs/writing-extensions/subcommands.html#environment-variables ? but I can't see packages filter in there as well, so not sure what is your policy on what's added there and what is ot

maciektr: This is a list of variables that Scarb will setup while calling other processes. And the flag I added is rather an input to Scarb. It will show up in `scarb build --help` like this: 
```
  -F, --features <FEATURES>          Comma separated list of features to activate [env: SCARB_FEATURES=]
  ```

THenry14: Ok, so there is no place in the docs for scarb gloval env variables? (or I cannot find it). I'm not saying there should be, I just want to understand better what is the approach here :)

maciektr: What do you mean by `global`? :D 
I wouldn't say this one is global - it's only read when parsing `scarb build` cli args.

THenry14: sorry, should have been more precise - by global I mean not only for other processes, but for scarb as a whole. 
So say we talk about this specific env that is for scarb build only - what are the ways to discover this variable? My understanding is that only if I do scarb build --help, right? I wonder if those things should be better documented, eg by inclusion in scarb docs somewhere.

THenry14: as I said though, please treat this as a nitpick - this is non blocking for this PR, and probably a topic for separate issue

maciektr: > scarb build only - what are the ways to discover this variable 

All arguments (including env) that are parsed with clap are discoverable through autogenerated clap help (e.g. `scarb build --help`). 

>  I wonder if those things should be better documented

We don't document passing it as `scarb build --target-names` either, so why would we only document envs? 🤔 I am not sure if it's worth keeping the docs up to date manually later.

Issue #1393, opened 2024-06-25
Cairo-run: Support u128 in felt252 arguments deserialization 


maciektr:
### Problem

Running `scarb cairo-run '[10, 11, 1129815197211541481934112806673325772687763881719835256646064516195041515616]' ` results in `out of range` error, while `felt252` should support `u128`. We only support up to `u64`, as that's what `serde_json` allows for (see https://github.com/serde-rs/json/issues/846).

PR #1408, opened 2024-07-03
Add inlining strategy to compiler config

maciektr:
commit-id:65927daa

---

**Stack**:
- #1411
- #1410
- #1408 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #1410, opened 2024-07-03
Cairo run fix argument deserialization for numbers larger than u64

maciektr:
**Stack**:
- #1410 ⬅
- #1408


maciektr:
fix #1393

DelevoXDG (approved):

extensions/scarb-cairo-run/src/deserialization.rs
```diff
@@ -91,8 +91,8 @@ impl Args {
         for arg in iterator {
             match arg {
                 Value::Number(n) => {
-                    let n = n.as_u64().ok_or(ArgsError::NumberOutOfRange)?;
-                    args.push(Arg::Value(Felt252::from(n)));
+                    let n = Felt252::from_str(n.to_string().as_str())?;
```
THenry14, line 94: nitpick: I winder if we could get rid of intermediate string creation (from_str -> to_string -> as_str)
maybe we could consider using format! ? I imageine something like this could work `Felt252::from_str(&format!("{}", n))` . I guess this could be more idiomatic, but am not strong on this (hence nitpick)

THenry14 (approved):

extensions/scarb-cairo-run/src/deserialization.rs
maciektr: I don't really see how would it be different, as format! does allocate a String anyway?

THenry14: shoot, you're right, I forgot. yeah, then this makes this comment even more of a nitpick, as this would probably be purely cosmetic change. completely up to you, both are fine for me :+1:

PR #1421, opened 2024-07-05
Add `publish` field to Scarb.toml to mark package as not publishable to the registry

### LINKED ISSUE #636: Add `publish` field to Scarb.toml to mark package as not publishable to the registry
Following, package should yield errors when doing `scarb package` and `scarb publish`.

```toml
[package]
name = "foobar"
version = "1.0.0"
publish = false
```

This field should be `true` by default.

DelevoXDG:
Closes #636

### DIFF
--- a/scarb/src/core/manifest/toml_manifest.rs
+++ b/scarb/src/core/manifest/toml_manifest.rs
@@ -186,6 +186,7 @@ pub struct TomlPackage {
     pub name: PackageName,
     pub version: MaybeWorkspaceField<Version>,
     pub edition: Option<MaybeWorkspaceField<Edition>>,
+    pub publish: Option<bool>,
     pub authors: Option<MaybeWorkspaceField<Vec<String>>>,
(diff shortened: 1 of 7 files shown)


scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -456,6 +457,8 @@ impl TomlManifest {
 
         let targets = self.collect_targets(package.name.to_smol_str(), root)?;
 
+        let publish = package.publish.unwrap_or(true);
```
THenry14, line 460: was this discussed somewhere? I'm trying to understand how desirable would that be for users, maybe publishable packages should be publishable only by explicitly stating so?

DelevoXDG: I considered this as well, but it was described this way in the issue and cargo also has this as true by default

THenry14 (approved):

PR #1423, opened 2024-07-08
JSON output for `scarb-doc`

### LINKED ISSUE #1430: Allow to dump information from packages' to JSON
Some tools could use the intermediate struct (`Crate` struct) with info about top level language elements from packages to generate documentation on their own. The feature should be marked as unstable

piotmag769:
Closes #1430


extensions/scarb-doc/src/main.rs
```diff
@@ -12,6 +12,13 @@ use scarb_doc::generate_language_elements_tree_for_package;
 struct Args {
     #[command(flatten)]
     packages_filter: PackagesFilter,
+
+    /// Print information collected from packages to stdout in JSON format instead of generating a
+    /// documentation.
+    /// This feature may be useful if you want to generate documentation files by yourself.
+    /// The precise output structure is not guaranteed to be stable.
+    #[arg(long)]
+    unstable_json_output: bool,
```
mkaput, line 21: I suggest using an enum here under a flag `--output-format` so that you'll have room to add `html` output etc.

mkaput, line 21: missing e2e test

mkaput (changes requested):

extensions/scarb-doc/src/main.rs
piotmag769: Actually good idea, I guess I will just document this enum variant as unstable. I will dump it to a file to be consistent with other formats (md, html)

extensions/scarb-doc/src/main.rs
```diff
@@ -1,30 +1,89 @@
...
     let args = Args::parse();
 
-    let metadata = MetadataCommand::new().inherit_stderr().exec()?;
+    let metadata = MetadataCommand::new()
+        .inherit_stderr()
+        .exec()
+        .context("Metadata command failed")?;
     let metadata_for_packages = args.packages_filter.match_many(&metadata)?;
 
+    let mut json_output = serde_json::Map::new();
```
mkaput, line 43: have you considered using BTreeMap here instead? or just a vector of key-value pairs?

extensions/scarb-doc/src/main.rs
```diff
@@ -1,30 +1,89 @@
...
+            serde_json::to_value(crate_).expect("Failed to serialize information about a crate"),
+        );
+    }
+
+    let output_dir = metadata
+        .target_dir
+        .unwrap_or_else(|| metadata.workspace.root.join("target"))
+        .join("doc");
+
+    fs::create_dir_all(&output_dir).context("Failed to create output directory for scarb doc")?;
```
mkaput, line 63:
```suggestion
    fs::create_dir_all(&output_dir).context("failed to create output directory for scarb doc")?;
```

Scarb used to use lowercase error messages

mkaput (changes requested):

extensions/scarb-doc/src/main.rs
piotmag769: It is a BTreeMap underneath

mkaput: so why are you not using that one? you'd get better typing

extensions/scarb-doc/src/main.rs
```diff
@@ -1,30 +1,90 @@
...
     let args = Args::parse();
 
-    let metadata = MetadataCommand::new().inherit_stderr().exec()?;
+    let metadata = MetadataCommand::new()
+        .inherit_stderr()
+        .exec()
+        .context("metadata command failed")?;
     let metadata_for_packages = args.packages_filter.match_many(&metadata)?;
 
+    let mut package_information_map = BTreeMap::new();
```
mkaput, line 44: consider adding a `version: 1` field for ensuring future changes being feasible

extensions/scarb-doc/tests/data/json_output_test_data.json
```diff
@@ -0,0 +1,278 @@
...
+            "doc": "Pair type alias for a tuple of two u32 values\n",
+            "full_path": "hello_world::Pair",
+            "name": "Pair",
+            "signature": "type Pair = (u32, u32);"
+          }
+        }
+      ]
+    }
+  }
+}
```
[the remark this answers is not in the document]
piotmag769: Cannot do that, the content is compared

extensions/scarb-doc/src/main.rs
piotmag769: Good idea, was going with starknet flow aka "i will just break it" xD

extensions/scarb-doc/tests/data/json_output_test_data.json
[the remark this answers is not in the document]
mkaput: just print extra endline when emitting this :)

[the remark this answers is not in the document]
mkaput: it makes `cat`-ing such files actually work 🙃

extensions/scarb-doc/src/versioned_json_output.rs
```diff
@@ -0,0 +1,22 @@
...
+use serde::Serialize;
+use std::collections::BTreeMap;
+
+type PackageName = String;
+
+const FORMAT_VERSION: u8 = 1;
+
+#[derive(Serialize)]
+pub struct VersionedJsonOutput {
+    format_version: u8,
```
mkaput, line 11: FYI we use'd to use `version` there, but do as you like

mkaput (approved):

PR #1434, opened 2024-07-09
WIP: Allow for custom registries urls

THenry14:
WIP: please do not review yet

The PR aims to allow scarb to use a custom registry, by defining an env variable

### DIFF
--- a/scarb/src/core/registry/mod.rs
+++ b/scarb/src/core/registry/mod.rs
@@ -1,5 +1,7 @@
 use anyhow::Result;
 use async_trait::async_trait;
+use once_cell::sync::Lazy;
+use std::env;
 
 use crate::core::{ManifestDependency, Package, PackageId, Summary};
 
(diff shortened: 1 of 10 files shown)


scarb/src/core/registry/mod.rs
```diff
@@ -11,7 +13,15 @@ pub mod patch_map;
 pub mod patcher;
 pub mod source_map;
 
-pub const DEFAULT_REGISTRY_INDEX: &str = "https://there-is-no-default-registry-yet.com";
+static DEFAULT_REGISTRY_INDEX: Lazy<&'static str> = Lazy::new(|| {
+    let registry_url = env::var("SCARB_REGISTRY_URL")
+        .unwrap_or_else(|_| "https://there-is-no-default-registry-yet.com".into());
+    Box::leak(registry_url.into_boxed_str())
```
THenry14, line 19: I would be happy to hear any thoughts about this, as I'm not sure about its implications. Box::leak is used to convert the String to a &'static str by leaking the heap-allocated memory. This should be safe in this context because Lazy ensures the value is only initialized once and lives for the entire program duration, but still this may not be THE optimal way to achieve this.

mkaput: You can technically keep `Lazy<String>` there and it'd be more collect.

But I think this is not the best approach for implementing this. I'd rather put this into `Config`, so that this could eventually be stored in some kind of a file. After all, someone might want to use a custom proxy for example and they'd like to write the URL once to some file and leave it.

THenry14: Sure, could make more sense, thanks for pointers. 
btw is this something that, in your opinion, could be settable from the manifest file? Just wondering how I should approach this

mkaput: Definitely not from the manifest because this is something user-specific, not project

Cargo has this: https://doc.rust-lang.org/cargo/reference/config.html

THenry14: Ok cool! so based on this, and after discussing things with @maciektr , let's plan the work accordingly:

- I will first redo this PR, to read the url from a flag and hold it in the `Config` struct
- I will then implement reading this from said file, at some later date, in a different PR under this issue https://github.com/software-mansion/scarb/issues/1441

Sounds good?

mkaput: sounds good

PR #1447, opened 2024-07-15
Allow packaging with publish=false

maciektr:
- **Make packaging with publish=false possible**
- **Set manifest.publish default to true**


DelevoXDG (approved):
LGTM

PR #1461, opened 2024-07-17
Add book.toml generation

piotmag769:
commit-id:b5f6227b

---

**Stack**:
- #1464
- #1463
- #1462
- #1461 ⬅

### DIFF
--- a/extensions/scarb-doc/src/docs_generation/markdown/book_toml.rs
+++ b/extensions/scarb-doc/src/docs_generation/markdown/book_toml.rs
@@ -1,0 +1,29 @@
+use indoc::formatdoc;
+
+use crate::AdditionalMetadata;
+
+#[allow(dead_code)]
+pub fn generate_book_toml_content(package_metadata: &AdditionalMetadata) -> String {
+    formatdoc! {
+        r#"
+            [book]
+            authors = {:?}
+            language = "en"
+            multilingual = false
+            src = "src"
+            title = "{} - Cairo"
+
+            [output.html]
+            no-section-label = true
+
+            [output.html.playground]
+            runnable = false
+
+            [output.html.fold]
+            enable = true
+            level = 0
+        "#,
+        package_metadata.authors.clone().unwrap_or_else(|| vec!["<unknown>".to_string()]),
+        package_metadata.name
+    }
+}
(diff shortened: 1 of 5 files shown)


Cargo.lock
```diff
@@ -4738,6 +4738,7 @@ dependencies = [
  "cairo-lang-starknet",
  "cairo-lang-syntax",
  "cairo-lang-utils",
+ "camino",
```
mkaput, line 4741: something wrong happened to the lock while rebasing

extensions/scarb-doc/src/docs_generation/markdown/book_toml.rs
```diff
@@ -0,0 +1,29 @@
+use indoc::formatdoc;
+
+use crate::AdditionalMetadata;
+
+#[allow(dead_code)]
+pub(super) fn generate_book_toml_content(package_metadata: &AdditionalMetadata) -> String {
```
mkaput, line 6:
```suggestion
pub fn generate_book_toml_content(package_metadata: &AdditionalMetadata) -> String {
```

why not just do not re-export this item in super module?

mkaput (changes requested):

extensions/scarb-doc/src/docs_generation/markdown/book_toml.rs
mkaput: then this `(super)` part has no effect so you do not need it 😃 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dd44beed411270287a8fccc5b5826383

```rust
mod foo {
    mod bar {
        pub fn b() -> u64 {
            42
        }
    }

    pub fn f() -> u64 {
        bar::b()
    }
}

fn main() {
    println!("{}", foo::f());
    println!("{}", foo::bar::b());
}
```

```
   Compiling playground v0.0.1 (/playground)
error[E0603]: module `bar` is private
  --> src/main.rs:15:25
   |
15 |     println!("{}", foo::bar::b());
   |                         ^^^  - function `b` is not publicly re-exported
   |                         |
   |                         private module
   |
note: the module `bar` is defined here
  --> src/main.rs:2:5
   |
2  |     mod bar {
   |     ^^^^^^^

For more information about this error, try `rustc --explain E0603`.
error: could not compile `playground` (bin "playground") due to 1 previous error
```

mkaput (approved):

Arcticae (approved):

PR #1462, opened 2024-07-17
Add traits for generating markdown docs

piotmag769:
commit-id:9f0f272d

---

**Stack**:
- #1464
- #1463
- #1462 ⬅

### DIFF
--- a/extensions/scarb-doc/src/docs_generation.rs
+++ b/extensions/scarb-doc/src/docs_generation.rs
@@ -1,1 +1,156 @@
-mod markdown;
+#![allow(dead_code)]
+use crate::types::{
+    Constant, Crate, Enum, ExternFunction, ExternType, FreeFunction, Impl, ImplAlias, ImplConstant,
+    ImplFunction, ImplType, Member, Module, Struct, Trait, TraitConstant, TraitFunction, TraitType,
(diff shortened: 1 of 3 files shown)


mkaput (approved):

extensions/scarb-doc/src/docs_generation.rs
```diff
@@ -1 +1,156 @@
...
+fn collect_all_top_level_items(crate_: &Crate) -> TopLevelItems {
+    let mut top_level_items = TopLevelItems::default();
+
+    top_level_items.modules.push(&crate_.root_module);
+
+    collect_all_top_level_items_internal(&mut top_level_items, &crate_.root_module);
+    top_level_items
+}
+
+fn collect_all_top_level_items_internal<'a, 'b>(
```
Arcticae, line 34: Wouldn't writing this iteratively instead of recursion make this much simpler?

Arcticae (approved):

extensions/scarb-doc/src/docs_generation.rs
piotmag769: I don't see how - modules contain submodules, so it is more readable as recursive function imo. It just requires proficiency in lifetimes ;p

PR #1465, opened 2024-07-18
Enable registry usage by default

THenry14:


scarb/src/core/registry/mod.rs
```diff
@@ -11,7 +11,7 @@ pub mod patch_map;
 pub mod patcher;
 pub mod source_map;
 
-pub const DEFAULT_REGISTRY_INDEX: &str = "https://there-is-no-default-registry-yet.com";
+pub const DEFAULT_REGISTRY_INDEX: &str = "https://scarbs.xyz";
```
DelevoXDG, line 14: Wouldn’t it be more reasonable to insert the full index URL here instead:
```suggestion
pub const DEFAULT_REGISTRY_INDEX: &str = "https://scarbs.xyz/api/v1/index/";
```
and leave WELL_KNOWN_PATH as is?

It seems that  to `DEFAULT_REGISTRY_INDEX` could vary and be changed to something like 
- `https://scarbs.xyz/api/v2/index`
- `https://index.scarbs.xyz/`

While `WELL_KNOWN_PATH` as the naming suggest is something that is supposed to always be the same,  regardless of how the index path changes.

WDYT?

mkaput: even better, can't you just alias this path in the registry, so that `https://scarbs.xyz/config.json` actually resolves?

THenry14: Not sure about this - correct me if I'm wrong, but that would mean that in every lockfile (maybe other places as well) as a source we will list `registry+https://scarbs.xyz/api/v1/index/"` instead of `"registry+https://scarbs.xyz/"`.

I suppose it may make future changes (like aliasing path to the registry) a breaking change unnecessarily. 

If you think this is not a big issue though, let me know and I will make this change. 

sidenote: I will fix tests in a minute 😞

mkaput: I definitely would prefer `"registry+https://scarbs.xyz/"`. The `/api/...` path is registry implementation detail and is not bringing any value in lockfiles

DelevoXDG: > But it's also bad to assume all registries host their configs at /api/v1/index/config.json.

That was the idea, but I suppose we can leave this as is to avoid incorrectly populating lockfiles. I've created an issue for that to address this on the registry side: https://github.com/software-mansion/scarbs.xyz/issues/262

mkaput (approved):

maciektr (approved):

PR #1468, opened 2024-07-19
Foldable sidebar

piotmag769:


extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -215,3 +212,30 @@ fn generate_markdown_from_item_data(doc_item: &dyn DocItem, header_level: usize)
...
+
+macro_rules! impl_top_level_markdown_items_summary {
+    ($t:ty, $filename:expr) => {
+        impl TopLevelMarkdownItemsSummary for Vec<&$t> {
+            fn summary_filename(&self) -> &str {
+                $filename
+            }
+        }
+    };
+}
```
piotmag769, line 229: Same idea as above, I will need the file names in `SUMMARY.md`, so I want to keep them hardcoded in one place

piotmag769 (commented):
The general idea is to make a sidebar look like this, so the lists of items are foldable. To achieve this I need to generate additional files that are responsible for the foldable sections

image

extensions/scarb-doc/src/docs_generation.rs
```diff
@@ -153,3 +153,35 @@ impl_doc_item!(TraitType);
 impl_doc_item!(TraitFunction);
 impl_doc_item!(TypeAlias);
 impl_doc_item!(Variant);
+
+trait ItemsName<'a> {
+    const ITEMS_NAME: &'a str;
+}
```
piotmag769, line 159: The purpose of this trait is to avoid hardcoding the names multiple times (just hardcode them here)

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -215,3 +204,27 @@ fn generate_markdown_from_item_data(doc_item: &dyn DocItem, header_level: usize)
 
     markdown
 }
+
+pub trait TopLevelMarkdownItemsSummary<'a> {
+    const ITEMS_SUMMARY_FILENAME: &'a str;
+}
```
piotmag769, line 210: The purpose of this trait is to avoid hardcoding the names multiple times (just hardcode them here)

extensions/scarb-doc/src/docs_generation.rs
```diff
@@ -153,3 +153,35 @@ impl_doc_item!(TraitType);
...
+
+macro_rules! impl_items_name {
+    ($t:ty, $name:expr) => {
+        impl<'a> ItemsName<'a> for $t {
+            const ITEMS_NAME: &'a str = $name;
+        }
+    };
+}
```
mkaput, line 167: 1. wdyt?

```suggestion
trait ItemsName {
    const HEADER: &'static str;
}

macro_rules! impl_items_name {
    ($t:ty, $name:expr) => {
        impl ItemsName for $t {
            const HEADER: &'_ str = $name;
        }
    };
}
```

2. Why can't this be part of `DocItem` trait?

mkaput (changes requested):

mkaput (approved):

PR #1472, opened 2024-07-23
Make scarb new and init command interactive

### LINKED ISSUE #1207: Make `scarb new` command interactive
### Problem

As in the title 

### Proposed Solution

_No response_

### Notes

_No response_

ksew1:
Closes #1207 
Closes [#1982](https://github.com/foundry-rs/starknet-foundry/issues/1982)

Make scarb new and init command interactive


scarb/src/bin/scarb/args.rs
```diff
@@ -268,6 +268,12 @@ pub struct ScriptsRunnerArgs {
     pub args: Vec<OsString>,
 }
 
+#[derive(ValueEnum, Clone, Debug)]
+pub enum TestRunner {
+    StarknetFoundry,
+    CairoNativeRunner,
```
maciektr, line 274:
```suggestion
    CairoTest,
```

I think that `cairo-test` is usually used to refer to that

scarb/src/bin/scarb/commands/new.rs
```diff
@@ -18,7 +18,7 @@ pub fn run(args: NewArgs, config: &Config) -> Result<()> {
             } else {
                 VersionControl::Git
             },
-            snforge: args.init.snforge,
+            snforge: matches!(args.init.test_runner, Some(TestRunner::StarknetFoundry)),
```
maciektr, line 21: I feel like we could run the prompt here, instead of modyfing the command at `bin/main.rs`.

scarb/tests/new_and_init.rs
maciektr, line 1: Please also fix `snforge_init.rs` test. It's ignored, but runs in scheduled tests / release checks / nightlies

scarb/src/bin/scarb/commands/init.rs
```diff
@@ -24,7 +25,14 @@ pub fn run(args: InitArgs, config: &Config) -> Result<()> {
             } else {
                 VersionControl::Git
             },
-            snforge: args.snforge,
+            snforge: matches!(
+                if let Some(test_runner) = args.test_runner {
```
maciektr, line 29: Maybe https://doc.rust-lang.org/core/option/enum.Option.html#method.unwrap_or_else? 🤔

scarb/src/bin/scarb/interactive.rs
```diff
@@ -0,0 +1,23 @@
...
+    } else {
+        vec![
+            "Cairo Native runner (default)",
+            "Starknet Foundry (recommended, requires snforge installed)",
+        ]
+    };
+
+    let answer = Select::new("Which test runner do you want to set up?", options).prompt()?;
+
+    if answer.starts_with("Cairo Native") {
```
maciektr, line 18: Probably no longer so

maciektr:
Would you be able to add any tests for that?

scarb/src/bin/scarb/commands/init.rs
ksew1: You can't use `?` in closures and `unwrap_or_else` uses them

maciektr: Hmm, maybe something like this
```
Ok(args.test_runner).transpose().unwrap_or_else(ask_for_test_runner)?
```

ksew1: Looks clean 😄, I will add this tomorrow

scarb-metadata/tests/scarb_command.rs
```diff
@@ -40,7 +40,7 @@ fn sample_project() {
 
 fn init_project(t: &TempDir) {
     Command::new(scarb_bin())
-        .args(["init", "--name", "hello"])
+        .args(["init", "--name", "hello", "--test-runner", "cairo-test"])
```
maciektr, line 43: We have a funny test case here: https://github.com/software-mansion/scarb/blob/a142df1dd521b8a78cd265877d1cd753a7aa437e/.github/workflows/ci.yml#L67-L78
Which runs tests from this package with older version of Scarb, to ensure it's still compatible in regards to metadata format. 
This older version of scarb will not accept `test-runner` as an argument.

scarb-metadata/tests/command.rs
```diff
@@ -69,7 +69,7 @@ fn manifest_path() {
 
 fn init_project(t: &TempDir) {
     Command::new(scarb_bin())
-        .args(["init", "--name", "hello"])
+        .args(["init", "--name", "hello", "--test-runner", "cairo-test"])
```
maciektr, line 72: We have a funny test case here: https://github.com/software-mansion/scarb/blob/a142df1dd521b8a78cd265877d1cd753a7aa437e/.github/workflows/ci.yml#L67-L78
Which runs tests from this package with older version of Scarb, to ensure it's still compatible in regards to metadata format. 
This older version of scarb will not accept `test-runner` as an argument.

ksew1: But as far as I understand, the normal tests won't pass. They need `--test-runner` as they also run this package.

ksew1:
> Would you be able to add any tests for that?

Regarding the tests, the task is harder than it seemed. We can't run this as `std::process::Command` as it's not `TTY`. I tried `PtyProcess`, which works on my computer but not on CI. Additionally, this is platform-dependent, so it won't build on Windows. I don't know if it is really worth it.

We can write unit tests by adjusting the function signature as follows:

```rust
pub fn ask_for_test_runner<F>(prompt: F) -> Result<TestRunner>
where
    F: FnOnce(Vec<&str>) -> InquireResult<&str>
```

and then mocking the prompt.

However, if you think e2e tests are really important, I can try to research this topic further and maybe find a solution to run them.

scarb-metadata/tests/command.rs
maciektr: Yeah, but this test is very important to me (it can warn us about unintentional breaking changes in the metadata format), so we need to think about some solution that works for both cases 😄 

You can always run `scarb init --help` and parse the output to see if it accepts the test runner flag or not. 

Another approach you can consider is reading the value from env - https://docs.rs/clap/latest/clap/struct.Arg.html#method.env
Say, `SCARB_INIT_TEST_RUNNER` or similar. It might be also useful in `snforge init` which calls Scarb under the hood.

ksew1: I think the env approach is better, but I won't be able to export it in CI with my permissions. So when I apply that fix, could you do that for me?

maciektr: Just add it here with `.env` on Command.

maciektr:
That's why I asked 😃 

If it's too hard, I can live without tests (is this a problem with getting output only, or are the files not created on CI? 🤔 ). Unit approach is not worth it imho, as it mocks exactly what we would want to test in this scenario.

ksew1:
> That's why I asked 😃
> 
> If it's too hard, I can live without tests (is this a problem with getting output only, or are the files not created on CI? 🤔 ). Unit approach is not worth it imho, as it mocks exactly what we would want to test in this scenario.

The issue is getting input; it seems it is incorrectly passed. I haven't debugged it, though.

Cargo.toml
```diff
@@ -81,6 +81,7 @@ ignore = "0.4"
 include_dir = "0.7"
 indicatif = "0.17"
 indoc = "2"
+inquire = "0.7.5"
```
mkaput, line 84: May I ask why is this using `inquire` instead of `dialoguer`?

`inquire` is based off `crossterm`, not `console`, so we have two crates for console APIs that are incompatible, and `scarb-ui` is doing some initial init for `console`. And, we're basically duplicating code here.

scarb-metadata/tests/command.rs
```diff
@@ -70,6 +70,7 @@ fn manifest_path() {
 fn init_project(t: &TempDir) {
     Command::new(scarb_bin())
         .args(["init", "--name", "hello"])
+        .env("SCARB_TEST_RUNNER", "cairo-test")
```
mkaput, line 73: why is this line needed everywhere? can this set of changes be avoided?

scarb/src/bin/scarb/args.rs
```diff
@@ -279,9 +285,9 @@ pub struct InitArgs {
     #[arg(long)]
     pub no_vcs: bool,
 
-    /// Use a Starknet Foundry Forge template.
-    #[arg(long)]
-    pub snforge: bool,
+    /// Test runner to use. Starts interactive session if not specified.
+    #[arg(long, env = "SCARB_TEST_RUNNER")]
```
mkaput, line 289:
```suggestion
    #[arg(long, env = "SCARB_INIT_TEST_RUNNER")]
```

isn't this more correct? without the `INIT` part it looks like a switch that forces test runner in runtime, not project creation

scarb/src/bin/scarb/commands/new.rs
```diff
@@ -18,7 +19,12 @@ pub fn run(args: NewArgs, config: &Config) -> Result<()> {
             } else {
                 VersionControl::Git
             },
-            snforge: args.init.snforge,
+            snforge: matches!(
+                Ok(args.init.test_runner)
+                    .transpose()
+                    .unwrap_or_else(ask_for_test_runner)?,
+                TestRunner::StarknetFoundry
+            ),
```
mkaput, line 27: duplicate change, please deduplicate this

scarb/src/bin/scarb/commands/init.rs
```diff
@@ -24,7 +25,12 @@ pub fn run(args: InitArgs, config: &Config) -> Result<()> {
             } else {
                 VersionControl::Git
             },
-            snforge: args.snforge,
+            snforge: matches!(
+                Ok(args.test_runner)
+                    .transpose()
+                    .unwrap_or_else(ask_for_test_runner)?,
```
mkaput, line 31: this sounds invalid. we cannot ask in non-interactive shell. is your library handling this under the hood or so?

mkaput (changes requested):

Cargo.toml
ksew1: Will switch to `dialoguer`

scarb-metadata/tests/command.rs
ksew1: This line is repeated twice in the functions that are the same. If you mean `args(CAIRO_NATIVE_RUNNER)`, this can be avoided if you want to be less explicit in tests. We can set it in the `pub fn std(self) -> StdCommand` function.

mkaput: If it's not needed for these tests, then I'd stick with default behaviour because this is out of interest of the test

scarb/src/bin/scarb/commands/init.rs
ksew1: We can check if it is a `TTY` device via the [`atty`](https://crates.io/crates/atty) crate. If it is not, we can either go with the default option or exit with an error. What do you think?

mkaput: `atty` has a security hole and in general is long-time abandoned in the ecosystem

https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html is the way to go :)

scarb/src/bin/scarb/commands/new.rs
ksew1: We can't use `TestRunner` as an argument to `ops::InitOptions`, as it is not available there. 

Is something like this okay?

```rust
matches!(
    get_or_ask_for_test_runner(args.init.test_runner)?,
    TestRunner::StarknetFoundry
)
```

scarb/src/bin/scarb/interactive.rs
```diff
@@ -1,9 +1,17 @@
...
-use inquire::Select;
+use anyhow::{ensure, Result};
+use dialoguer::theme::ColorfulTheme;
+use dialoguer::Select;
 use which::which;
 
 pub fn ask_for_test_runner() -> Result<TestRunner> {
+    ensure!(
+        io::stdout().is_terminal(),
+        "You are not running in terminal. Please provide the --test-runner flag."
```
mkaput, line 12: lowercase errors, no dots, use `note:` convention

scarb/src/bin/scarb/interactive.rs
```diff
@@ -13,9 +21,13 @@ pub fn ask_for_test_runner() -> Result<TestRunner> {
         ]
     };
 
-    let answer = Select::new("Which test runner do you want to set up?", options).prompt()?;
+    let selection = Select::with_theme(&ColorfulTheme::default())
```
mkaput, line 24: how is this working when colors are turned off? (Scarb UI is providing this information)

mkaput (changes requested):

scarb/src/bin/scarb/commands/new.rs
mkaput: yep

scarb/src/bin/scarb/interactive.rs
ksew1: When tested on `export TERM=dumb` it works normally (of course without colors).

mkaput: and what about these envs https://docs.rs/console/latest/console/fn.colors_enabled.html?

ksew1: As expected, there is no color, but nothing breaks.

mkaput: perfect!

mkaput (approved):

scarb/src/internal/serdex.rs
maciektr, line 1: Please rebase onto main

website/docs/writing-extensions/subcommands.md
maciektr, line 1: This change is invalid. 
This page describes env variables that are passed **by scarb** to subprocesses created by Scarb, regardless of user env configuration. 
So it's not input of Scarb, but rather some form of an output.

maciektr (commented):
Please make sure to use the new env var in `snforge init` after this is merged :D

website/docs/writing-extensions/subcommands.md
ksew1: Reverted

scarb/src/internal/serdex.rs
ksew1: Done 😄

maciektr (approved):

Issue #1457, opened 2024-07-17
Research adding support for `cairo` to mdbook


piotmag769:
We can see how `cairo-book` did it, tagged with `// TODO(#1457`

Arcticae:
This can also be added by defining a language for https://highlightjs.org/ - it's the mechanism that mdbook uses for syntax highlighting.
We could use fork [rust definition](https://github.com/highlightjs/highlight.js/blob/main/src/languages/rust.js) and modify it for cairo (shouldn't be too much of a problem), and then integrate it into mdbook - see [example implementation](https://github.com/pen-lang/pen/pull/395/files)

https://highlightjs.readthedocs.io/en/latest/language-guide.html

PR #1499, opened 2024-07-31
Create highlight.js file for mdBook cairo support

wawel37:
[Issue #1457](https://github.com/orgs/software-mansion/projects/12/views/1?pane=issue&itemId=71242578)

### DIFF
--- a/extensions/scarb-doc/theme/cairo.js
+++ b/extensions/scarb-doc/theme/cairo.js
@@ -1,0 +1,246 @@
+/*
+Language: Cairo
+Website: https://www.cairo-lang.org
+Category: common, system
+*/
+
+/** @type LanguageFn */
+
+export default function(hljs) {
+  const UNDERSCORE_IDENT_RE = hljs.UNDERSCORE_IDENT_RE;
+  const KEYWORDS = [
+    "as",
+    "break",
+    "const",
+    "continue",
+    "else",
+    "enum",
+    "extern",
+    "false",
+    "fn",
+    "if",
+    "impl",
+    "implicits",
+    "let",
+    "loop",
+    "match",
+    "mod",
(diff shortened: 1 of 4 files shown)


extensions/scarb-doc/theme/highlight.js
maciektr, line 1: Shouldn't this be `highlight.min.js` then? 🤔

wawel37: I mean, i based it upon the contribution guide to mdBook [here](https://github.com/rust-lang/mdBook/blob/master/CONTRIBUTING.md#updating-highlightjs). Also wanted to be specific, that the file after the mdBook generation, should be named `highlight.js` explicitly. I mean, we can change the name of it to `...min.js`, but it could suggest to the user, that he should also modify the name of the file inside the generated mdBook files. I don't have a strong opinion on that tbh

extensions/scarb-doc/theme/cairo.js
```diff
@@ -0,0 +1,246 @@
...
+        begin: hljs.IDENT_RE + "::",
+        keywords: {
+          keyword: "Self",
+          built_in: BUILTINS,
+          type: TYPES,
+        },
+      },
+      {
+        className: "punctuation",
+        begin: "->",
```
piotmag769, line 242: What is this one?

mkaput: this one is gonna be used in closures syntax:

https://github.com/starkware-libs/cairo/blob/106bb50f54e615ac7339e9ca1cb18743d0114579/crates/cairo-lang-semantic/src/expr/test_data/closure#L8

Arcticae (approved):

PR #1511, opened 2024-08-07
Doc/features support

wawel37:


mkaput (commented):
.

extensions/scarb-doc/src/lib.rs
```diff
@@ -29,15 +34,35 @@ pub struct AdditionalMetadata {
...
+                .features
+                .unwrap_or_default();
+
+        let initial_cfg_set = initial_cfg_set();
+        let cfg_wtih_features = get_cfg_with_features(
+            initial_cfg_set.clone(),
+            &features_manifest,
+            &enabled_features,
+            false,
+        )?;
```
mkaput, line 54: isn't this already available from scarb metadata?

extensions/scarb-doc/src/lib.rs
```diff
@@ -29,15 +34,35 @@ pub struct AdditionalMetadata {
...
     metadata: &Metadata,
     metadata_for_packages: &[PackageMetadata],
-) -> Vec<PackageInformation> {
+    enabled_features: FeaturesOpts,
+) -> Result<Vec<PackageInformation>, anyhow::Error> {
     let mut packages_information = vec![];
     for package_metadata in metadata_for_packages {
         let authors = package_metadata.manifest_metadata.authors.clone();
 
+        let features_manifest: BTreeMap<FeatureName, Vec<FeatureName>> =
```
mkaput, line 43: doing this may lead to potentially incomplete information, because who knows when Scarb decides to add some custom logic to features collection/generation

it would be much better to collect all features via `scarb metadata`. if necessary fields are missing, the feel free to add them there (but please do it in separate PR)

extensions/scarb-doc/Cargo.toml
```diff
@@ -26,6 +26,7 @@ indoc.workspace = true
 itertools.workspace = true
 scarb-metadata = { path = "../../scarb-metadata" }
 scarb-ui = { path = "../../utils/scarb-ui" }
+scarb = { path = "../../scarb" }
```
mkaput, line 29: please don't do this, we want scarb to be a binary in the future, unusable as a crate

I'll add comments in all usage places how to do them differently

extensions/scarb-doc/Cargo.toml
```diff
@@ -35,3 +36,4 @@ smol_str.workspace = true
 assert_fs.workspace = true
 scarb-test-support = { path = "../../utils/scarb-test-support" }
 walkdir.workspace = true
+multizip = "0.1.0"
```
mkaput, line 39: why not https://docs.rs/itertools/latest/itertools/fn.multizip.html that we already have pulled-in?

extensions/scarb-doc/tests/code/code_3.cairo
```diff
@@ -0,0 +1,10 @@
+//! Sub-package code (with feature)
+
+#[cfg(feature: 'test_feature')]
```
piotmag769, line 3: Vide supra

extensions/scarb-doc/tests/code/code_4.cairo
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/// use into_trait
+use core::traits::Into as into_trait;
+use core::traits::TryInto;
+
+/// FOO constant with value 42
+const FOO: u32 = 42;
+
+#[cfg(feature: "test_feature")]
```
piotmag769, line 15: Vide supra (comment here)

extensions/scarb-doc/src/lib.rs
```diff
@@ -48,16 +73,22 @@ pub fn generate_packages_information(
...
+    Ok(packages_information)
 }
 
 fn generate_language_elements_tree_for_package(
     package_name: String,
     project_config: ProjectConfig,
+    features_cfg: CfgSet,
 ) -> Crate {
-    let db = ScarbDocDatabase::new(Some(project_config));
+    let db: ScarbDocDatabase =
```
piotmag769, line 84: No need to explicitly type it

extensions/scarb-doc/src/lib.rs
```diff
@@ -29,15 +34,35 @@ pub struct AdditionalMetadata {
...
     for package_metadata in metadata_for_packages {
         let authors = package_metadata.manifest_metadata.authors.clone();
 
+        let features_manifest: BTreeMap<FeatureName, Vec<FeatureName>> =
+            TomlManifest::read_from_path(&package_metadata.manifest_path)?
+                .features
+                .unwrap_or_default();
+
+        let initial_cfg_set = initial_cfg_set();
+        let cfg_wtih_features = get_cfg_with_features(
```
piotmag769, line 49: typo

extensions/scarb-doc/src/lib.rs
```diff
@@ -48,16 +73,22 @@ pub fn generate_packages_information(
...
+        ScarbDocDatabase::new(&package_name, Some(project_config), features_cfg);
 
     let main_crate_id = db.intern_crate(CrateLongId::Real(package_name.into()));
 
     Crate::new(&db, main_crate_id)
 }
+
+fn initial_cfg_set() -> CfgSet {
+    CfgSet::from_iter([Cfg::name("doc")])
+}
```
piotmag769, line 94: Should be in `db.rs` imo and hidden in db constructor, since we always want to extend the set with `doc`

extensions/scarb-doc/src/lib.rs
piotmag769: scarb metadata cannot determine what features should be enabled, no? I see there is no info about specified features for each package even :/ cc @maciektr

extensions/scarb-doc/src/lib.rs
```diff
@@ -29,15 +34,35 @@ pub struct AdditionalMetadata {
...
         let authors = package_metadata.manifest_metadata.authors.clone();
 
+        let features_manifest: BTreeMap<FeatureName, Vec<FeatureName>> =
+            TomlManifest::read_from_path(&package_metadata.manifest_path)?
+                .features
+                .unwrap_or_default();
+
+        let initial_cfg_set = initial_cfg_set();
+        let cfg_wtih_features = get_cfg_with_features(
+            initial_cfg_set.clone(),
```
piotmag769, line 50: Why clone here?

extensions/scarb-doc/src/lib.rs
```diff
@@ -29,15 +34,35 @@ pub struct AdditionalMetadata {
...
+            None => initial_cfg_set,
+        };
+
         let project_config = get_project_config(metadata, package_metadata);
+
         let crate_ = generate_language_elements_tree_for_package(
             package_metadata.name.clone(),
-            project_config,
+            project_config.clone(),
+            cfg.clone(),
```
piotmag769, line 65: Why clone here?

piotmag769 (changes requested):

extensions/scarb-doc/src/lib.rs
wawel37: Your right, i looked through the code and will create separate PR before merging this one, that will collect features manifest for each package in metadata command

scarb/src/ops/resolve.rs
```diff
@@ -372,12 +373,14 @@ fn cairo_compilation_unit_for_target(
     })
 }
 
-fn get_cfg_with_features(
+pub fn get_cfg_with_features(
     mut cfg_set: CfgSet,
     features_manifest: &BTreeMap<FeatureName, Vec<FeatureName>>,
     enabled_features: &FeaturesOpts,
+    // When generating docs, we don't want to be strict whether a package has defined features in manifest.
```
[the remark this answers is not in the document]
wawel37: That's true, this function was forcing it to be strict about whether the scarb.toml got the features defined or not. With this flag, we can disable that check (as it's done in cargo doc).

extensions/scarb-doc/src/lib.rs
wawel37: Vsc extension is putting it as i accidently click the infered type hint xd 
image

extensions/scarb-doc/src/db.rs
```diff
@@ -71,6 +77,20 @@ impl ScarbDocDatabase {
...
+
+        if let Some(mut crate_config) = root_crate_config {
+            crate_config.settings.cfg_set = Some(features_cfg);
+            self.set_crate_config(root_crate_id, Some(crate_config.clone()));
+        }
+    }
+
+    fn initial_cfg_set() -> CfgSet {
+        CfgSet::from_iter([Cfg::name("doc")])
+    }
```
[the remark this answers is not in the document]
wawel37: moved it back

extensions/scarb-doc/src/db.rs
```diff
@@ -31,14 +32,19 @@ pub struct ScarbDocDatabase {
...
+    ) -> Self {
         let mut db = Self {
             storage: Default::default(),
         };
 
         init_files_group(&mut db);
 
-        db.set_cfg_set(Self::initial_cfg_set().into());
+        let initial_cfg = Self::initial_cfg_set();
+        db.set_cfg_set(initial_cfg.into());
```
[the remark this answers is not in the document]
wawel37: Do you mean that i created separate variable for that or why do we set the global Cfg for db? Because I'm wondering rn if we even need this global cfg at all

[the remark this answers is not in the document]
piotmag769: The first one. And yes, we need the global config, if a package has `cfg_set: None` then it will be used

extensions/scarb-doc/src/lib.rs
piotmag769: Consult any changes to `scarb metadata` with @maciektr

extensions/scarb-doc/src/lib.rs
```diff
@@ -29,15 +34,35 @@ pub struct AdditionalMetadata {
...
+        let cfg_wtih_features = get_cfg_with_features(
+            initial_cfg_set.clone(),
+            &features_manifest,
+            &enabled_features,
+            false,
+        )?;
+        let cfg = match cfg_wtih_features {
+            Some(cfg_with_features) => cfg_with_features,
+            None => initial_cfg_set,
+        };
```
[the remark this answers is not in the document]
wawel37: Will push change where i put it inside the project config. Could you check if that's what you meant here?

extensions/scarb-doc/src/lib.rs
wawel37: After discussion with @maciektr, we decided to use the metadata command directly, and make the behaviour of the missing features inside the `Scarb.toml` for separate task

extensions/scarb-doc/src/lib.rs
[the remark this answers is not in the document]
wawel37: It's irrelevant now as we got rid of manual feature cfg creation

scarb/src/ops/resolve.rs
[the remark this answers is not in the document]
wawel37: Irrelevant now, as we use features from metadata command directly

extensions/scarb-doc/src/lib.rs
wawel37: Removed it since we are using the features metadata from metadata command.

extensions/scarb-doc/Cargo.toml
wawel37: Removed it completely as we don't need it right now (using metadata command for features)

piotmag769 (approved):

extensions/scarb-doc/tests/code/code_3.cairo
```diff
@@ -0,0 +1,18 @@
+//! Sub-package code (with feature)
+
+/// Function that prints "test" to stdout with endline.
+/// Can invoke it like that:
+/// ```cairo
+///     fn main() {
+///         test();
+///     }
+/// ```
+#[cfg(feature: 'test_feature')]
```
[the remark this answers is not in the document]
wawel37: It also works, added test to cover that behaviour

extensions/scarb-doc/tests/test.rs
```diff
@@ -1,117 +1,44 @@
...
-            fn it_works() {
-                assert(fib_function(16) == 987, 'it works!');
-            }
-        }
-    "#
-};
+const FIBONACCI_CODE_WITHOUT_FEATURE: &str = include_str!("code/code_1.cairo");
+const FIBONACCI_CODE_WITH_FEATURE: &str = include_str!("code/code_4.cairo");
+const COMMON_CODE_WITHOUT_FEATURE: &str = include_str!("code/code_2.cairo");
+const COMMON_CODE_WITH_FEATURE: &str = include_str!("code/code_3.cairo");
```
maciektr, line 34: Personally, I am not a big fan of test fixtures like this, as I am not sure we will reuse it across multiple test cases and the names do not help either. If I was doing it, I would probably just `indoc` them here directly, maybe splitting features tests cases into a separate file (and thus also the code - hiding it there from my eyes). 
However, it's only me, so I am open to leave it if you have a contrary opinion :D

maciektr (commented):
Looks great! 
Attached some comments below - only tests.

extensions/scarb-doc/tests/test.rs
```diff
@@ -159,3 +86,343 @@ fn markdown_output() {
         }
     }
 }
+
+#[test]
+fn test_workspace_no_features() {
+    let root_dir = TempDir::new().unwrap();
+    let child_dir = root_dir.child(SUB_PACKAGE_NAME);
+    fs::create_dir(child_dir.path()).expect("Couldn't create a sub package directory.");
```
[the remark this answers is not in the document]
wawel37: You are right about this 94 line. The project builder will throw an error if the child_dir doesn't exist. The project is still built using the ProjectBuilder. The only thing we are doing here is creating directory for it. If we want to handle creation of the child directory in the project builder, imo we should also move the root_dir creation there.

extensions/scarb-doc/tests/test.rs
```diff
@@ -159,3 +86,343 @@ fn markdown_output() {
...
+            expect_file.assert_eq(&content);
+        }
+    }
+}
+
+#[test]
+fn test_workspace_without_features_in_manifest() {
+    let root_dir = TempDir::new().unwrap();
+    let child_dir = root_dir.child(SUB_PACKAGE_NAME);
+    fs::create_dir(child_dir.path()).expect("Couldn't create a sub package directory.");
```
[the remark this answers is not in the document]
wawel37: my comment above also refers to this

extensions/scarb-doc/tests/test.rs
wawel37: Imo ur right about hiding it under a certain variable, but this way we are separating them, because they are very large. Just wanted to make a test.rs file just about testing logic, not about certain "snapshots". We could also separate different test regarding features in a multiple different files, but I think they are so much connected to each other. Putting them in separate files would make it just unclear. Imho I would leave it like that, because there is so much code in those snapshots, that scrolling up to see what the code actually do and finding the example would take more time, than peaking inside the snapshot code file

maciektr (approved):
Looks great!

PR #1543, opened 2024-08-25
starknet_multiple_contracts example refurbishment

stranger80:
The https://github.com/software-mansion/scarb/blob/main/examples/starknet_multiple_contracts/src/forty_two.cairo sample included a 'shorthand' styled contract code (no explicitly specified contract interface), and the `answer()` method is interpreted as a 'write' method by the compiler, even though it does not alter stae. 

The proposed contract code is hopefully less confusing.


maciektr (approved):

maciektr:
Thank you for pointing this out @stranger80 ! 🔥

PR #1556, opened 2024-08-28
Support more ModuleItems in proc macro attribute expansion

### LINKED ISSUE #1532: Support proc macro attributes on module item impl
Reproduction: 

```
#[starknet::contract]
mod one {
    use starknet::storage::{StoragePointerReadAccess, StoragePointerWriteAccess};
    use starknet::get_contract_address;
    use super::{IInc, IIncDispatcher, IIncDispatcherTrait};

    #[storage]
    struct Storage {
        counter: u128
    }

    #[abi(embed_v0)]
    impl IncImpl of IInc<ContractState> {
        #[nonreentrant]
        fn infinite_inc(ref self: ContractState) {
            self.counter.write(self.counter.read() + 1);
            IIncDispatcher { contract_address: get_contract_address() }.infinite_inc();
        }
    }
}
```

The `nonreentrant` macro on `infinite_inc` is not run.

maciektr:
fix #1532

### DIFF
--- a/scarb/src/compiler/plugin/proc_macro/host.rs
+++ b/scarb/src/compiler/plugin/proc_macro/host.rs
@@ -4,7 +4,7 @@ use crate::compiler::plugin::proc_macro::{
 use crate::core::{Config, Package, PackageId};
 use anyhow::{ensure, Result};
 use cairo_lang_defs::ids::{ModuleItemId, TopLevelLanguageElementId};
-use cairo_lang_defs::patcher::PatchBuilder;
+use cairo_lang_defs::patcher::{PatchBuilder, RewriteNode};
 use cairo_lang_defs::plugin::{
     DynGeneratedFileAuxData, GeneratedFileAuxData, MacroPlugin, MacroPluginMetadata,
     PluginGeneratedFile, PluginResult,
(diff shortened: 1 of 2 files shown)


scarb/tests/build_cairo_plugin.rs
```diff
@@ -1431,3 +1431,175 @@ fn can_be_expanded() {
...
+    CairoPluginProjectBuilder::default()
+        .lib_rs(indoc! {r##"
+            use cairo_lang_macro::{attribute_macro, ProcMacroResult, TokenStream};
+
+            #[attribute_macro]
+            pub fn some(_attr: TokenStream, token_stream: TokenStream) -> ProcMacroResult {
+                ProcMacroResult::new(TokenStream::new(
+                    token_stream.to_string()
+                    .replace("hello", "world")
+                    .replace("12", "34")
```
wawel37, line 1448: nit: we replace literals here, it's also good to test whether they were replaced after running the changed function. Similar way as it's done for impl test below.

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -155,6 +156,186 @@ impl ProcMacroHostPlugin {
...
+    fn expand_inner_attr(
+        &self,
+        db: &dyn SyntaxGroup,
+        item_ast: ast::ModuleItem,
+    ) -> InnerAttrExpansionResult {
+        let mut context = InnerAttrExpansionContext::new(self);
+        let mut item_builder = PatchBuilder::new(db, &item_ast);
+
+        match item_ast.clone() {
+            ast::ModuleItem::Trait(trait_ast) => {
```
wawel37, line 168: nit: trait and impl logic are very similar. Maybe we could move some of the logic to separate function, so we just don't copy the code. Treat it as just a suggestion, i don't have a strong opinion on that

wawel37 (commented):
LGTM, but it would be great that someone with better context would look at it.

scarb/src/compiler/plugin/proc_macro/host.rs
maciektr: Good point, done.

PR #1570, opened 2024-09-02
[Draft] Add package publish

THenry14:


scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
+            Some(_) => Ok(true),
+            None => Ok(false),
+        }
     }
 
-    async fn publish(&self, _package: Package, _tarball: FileLockGuard) -> Result<()> {
-        todo!("Publishing to HTTP registries is not implemented yet.")
+    async fn publish(&self, package: Package, tarball: FileLockGuard) -> Result<RegistryUpload> {
+        let auth_token =
+            env::var("SCARB_AUTH_TOKEN").map_err(|_| anyhow!("Missing authentication token."))?;
```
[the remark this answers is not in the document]
maciektr: Maybe `SCARB_REGISTRY_AUTH_TOKEN`? 🤔

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
+            .send()
+            .await?;
+
+        match response.status() {
+            StatusCode::UNAUTHORIZED => Err(RegistryUpload::Unauthorized)
+                .map_err(|_| anyhow!("Invalid authentication token.")),
+            StatusCode::FORBIDDEN => Err(RegistryUpload::CannotPublish)
+                .map_err(|_| anyhow!("Missing upload permissions or not the package owner.")),
+            StatusCode::BAD_REQUEST => Err(RegistryUpload::VersionExists)
+                .map_err(|_| anyhow!("Package {} already exists.", &package.id)),
```
maciektr, line 210:
```suggestion
                .map_err(|_| anyhow!("package `{}` already exists", &package.id)),
```
etc

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
+            StatusCode::UNPROCESSABLE_ENTITY => {
+                Err(RegistryUpload::Corrupted).map_err(|_| anyhow!("File corrupted during upload."))
+            }
+            StatusCode::OK => Ok(RegistryUpload::Success),
+            _ => Err(RegistryUpload::Failed).map_err(|_| {
+                anyhow!(
+                    "Upload failed with an unexpected error (trace-id: {:?})",
+                    response
+                        .headers()
+                        .get("x-cloud-trace-context")
```
maciektr, line 220: Not sure if it is customary to include in user message 🤔

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
+            StatusCode::UNAUTHORIZED => Err(RegistryUpload::Unauthorized)
+                .map_err(|_| anyhow!("Invalid authentication token.")),
+            StatusCode::FORBIDDEN => Err(RegistryUpload::CannotPublish)
+                .map_err(|_| anyhow!("Missing upload permissions or not the package owner.")),
+            StatusCode::BAD_REQUEST => Err(RegistryUpload::VersionExists)
+                .map_err(|_| anyhow!("Package {} already exists.", &package.id)),
+            StatusCode::UNPROCESSABLE_ENTITY => {
+                Err(RegistryUpload::Corrupted).map_err(|_| anyhow!("File corrupted during upload."))
+            }
+            StatusCode::OK => Ok(RegistryUpload::Success),
```
maciektr, line 214: Why not just `()`, if all errors are returned as `Err()` anyway?

THenry14: Just a habit, I like being explicit. If you'd like me to change this I'm fine with it.

scarb/src/core/registry/client/http.rs
THenry14: sorry, fixed https://github.com/software-mansion/scarb/pull/1570/commits/8f2bc26e7086b4298db6de0331d69341a5ca6827

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
+            "cannot upload package - file does not exist at path: {}",
+            &path
+        );
+
+        let file = TokioFile::open(&path).await?;
+        let metadata = file.metadata().await?;
+        ensure!(
+            metadata.len() < 5 * 1024 * 1024,
+            "package size ({} B) is too large (max 5MB allowed)",
+            &metadata.len()
```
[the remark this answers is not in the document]
THenry14: sure, changed here https://github.com/software-mansion/scarb/pull/1570/commits/63b2a1661642c1346a1d8eff1028207b8976d498

scarb/src/bin/scarb/args.rs
```diff
@@ -483,7 +483,7 @@ pub struct PackageArgs {
 #[derive(Parser, Clone, Debug)]
 pub struct PublishArgs {
     /// Registry index URL to upload the package to.
-    #[arg(long, value_name = "URL")]
+    #[arg(long, value_name = "URL", default_value = "https://scarbs.xyz/")]
     pub index: Url,
```
mkaput, line 487:
```suggestion
    #[arg(long, value_name = "URL")]
    pub index: Option<Url>,
```

Wdyt about this approach? With this we can probably use package's source id to get registry url and thus avoid duplication.

scarb/src/core/registry/client/mod.rs
```diff
@@ -33,6 +33,23 @@ pub enum RegistryDownload<T> {
...
+    CannotPublish,
+    /// Package version already exists.
+    VersionExists,
+    /// File corrupted during upload.
+    Corrupted,
+    /// Upload failed for other reasons.
+    Failed,
+    /// Upload successful.
+    Success,
+}
```
mkaput, line 51: shouldn't all these error cases be kept as `anyhow::Error` objects? you seem to never match on the variants of this struct

scarb/src/ops/publish.rs
```diff
@@ -51,10 +52,18 @@ pub fn publish(package_id: PackageId, opts: &PublishOpts, ws: &Workspace<'_>) ->
...
+        let upload = registry_client.publish(package, tarball).await;
+        match upload {
+            Ok(RegistryUpload::Success) => {
+                ws.config().ui().print(Status::new(
+                    "Published",
+                    format!("{}", &dest_package_id).as_str(),
+                ));
+                Ok(())
+            }
+            _ => upload.map(|_| ()),
```
mkaput, line 64: do I see correctly that you're ignoring any errors and you're not informing the user about the cause?

utils/scarb-test-support/src/simple_http_server.rs
```diff
@@ -45,14 +46,18 @@ pub struct HttpLog {
 }
 
 impl SimpleHttpServer {
-    pub fn serve(dir: PathBuf) -> Self {
+    pub fn serve(dir: PathBuf, post_status: Option<u16>) -> Self {
```
mkaput, line 49: nitpick, I think that instead of passing and requiring this argument in a constructor I'd rather keep it as an `AtomicU16` as struct field and expose following method to use only when needed: `fn set_post_status(&mut self, status: u16)`

scarb/src/core/registry/client/http.rs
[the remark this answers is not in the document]
maciektr: Oh, I just meant `help: you can set your auth token in SCARB_RELEASE_AUTH_TOKEN env variable` or smth similar

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
 
-    async fn publish(&self, _package: Package, _tarball: FileLockGuard) -> Result<()> {
-        todo!("Publishing to HTTP registries is not implemented yet.")
+    async fn publish(&self, package: Package, tarball: FileLockGuard) -> Result<RegistryUpload> {
+        let auth_token =
+            env::var("SCARB_AUTH_TOKEN").map_err(|_| anyhow!("Missing authentication token."))?;
+
+        let path = tarball.path().to_owned();
+        // we need to drop, because windows file locking is very strict
+        drop(tarball);
```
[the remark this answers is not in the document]
mkaput: You can convert `FileLockGuard` into `AsyncFileLockGuard` which you can deref into `tokio::fs::File` which is what you eventually want I think?

`LocalRegistry` uses `fsx::copy` which makes a hard link on fs instead of copying data -- saves user storage. Here we have a different use-case.

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
+        let form = Form::new().part("file", file_part);
+
+        let response = self
+            .config
+            .online_http()?
+            .post(
+                index_config
+                    .upload
+                    .clone()
+                    .ok_or_else(|| anyhow!("Upload URL is missing"))?,
```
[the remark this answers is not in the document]
maciektr: Yep, would be ideal 👌

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
+        match index_config.upload {
+            Some(_) => Ok(true),
+            None => Ok(false),
+        }
     }
 
-    async fn publish(&self, _package: Package, _tarball: FileLockGuard) -> Result<()> {
-        todo!("Publishing to HTTP registries is not implemented yet.")
+    async fn publish(&self, package: Package, tarball: FileLockGuard) -> Result<RegistryUpload> {
+        let auth_token = env::var("SCARB_REGISTRY_AUTH_TOKEN")
```
mkaput, line 162: is this temporary?

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
-        todo!("Publishing to HTTP registries is not implemented yet.")
+    async fn publish(&self, package: Package, tarball: FileLockGuard) -> Result<RegistryUpload> {
+        let auth_token =
+            env::var("SCARB_AUTH_TOKEN").map_err(|_| anyhow!("Missing authentication token."))?;
+
+        let path = tarball.path().to_owned();
+        // we need to drop, because windows file locking is very strict
+        drop(tarball);
+
+        ensure!(
```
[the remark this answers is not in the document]
maciektr: ups, sorry, I got confused.

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
+            Path::new(&path).exists(),
+            "cannot upload package - file does not exist at path: {}",
+            &path
+        );
+
+        let file = TokioFile::open(&path).await?;
+        let metadata = file.metadata().await?;
+        ensure!(
+            metadata.len() < 5 * 1024 * 1024,
```
mkaput, line 178: why are we hardcoding a seemingly arbitrary number on client side?

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
+            .online_http()?
+            .post(
+                index_config
+                    .upload
+                    .clone()
+                    .ok_or_else(|| anyhow!("upload URL is missing"))?,
+            )
+            .header(AUTHORIZATION, auth_token)
+            .multipart(form)
+            .timeout(Duration::from_secs(60))
```
mkaput, line 200: https://brr.fyi/posts/engineering-for-slow-internet - consider this

scarb/src/core/registry/client/http.rs
mkaput: I think we should skip the `(trace-id: ...)` part altogether if data is missing. Otherwise, this makes an impression that something is missing by error.

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +151,77 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
     }
 
     async fn supports_publish(&self) -> Result<bool> {
-        // TODO(mkaput): Publishing to HTTP registries is not implemented yet.
-        Ok(false)
+        let index_config = self.index_config.load().await?;
+        match index_config.upload {
+            Some(_) => Ok(true),
+            None => Ok(false),
+        }
```
mkaput, line 158:
```suggestion
        Ok(self.index_config.load().await?.is_some())
```

scarb/src/core/registry/client/http.rs
THenry14: It takes ~ 60 seconds to upload 5MB file with 700Kb/s internet connection
EDGE connectivity standard (2.5G) approaches speeds up to 1Mb/s (less than that on average for a typical connection) - this is second slowest standard, right above GPRS (GSM) with 30 – 80 kb/s.
From what I gather slowest internet speeds are mostly in Africa, with lowest probably in the region of Yemen with ~700Kb/s

With all this context an educated guess was to do 60s. We can raise it to eg 120s if you prefer so :+1:

scarb/src/core/registry/client/mod.rs
THenry14: Why? What would be the advantage of doing this?
I went with this because find it more specific and easy to propagate. It also holds all possible results, not only Errors. But I am willing to change my mind on this, given strong arguments :+1:

scarb/src/ops/publish.rs
THenry14: Err are propagated, they are not consumed or modified in any way. You can see tests from registry_upload.rs (added i this PR)

utils/scarb-test-support/src/simple_http_server.rs
THenry14: This would have to be modified in router probably, that is defined in serve function. Not sure this will be this trivial, but if you insist I can try and research this.

scarb/src/bin/scarb/args.rs
THenry14: Yup, I've considered this but at the end opted to stick with clap's default value, for no particular reason. 
If you prefer it this way (and deduplication is always a valid argument) we can do that no prob. 
Changed in: https://github.com/software-mansion/scarb/pull/1570/commits/c6a5c0fd058c402b3fd37028513c9b33d9cf77fe

scarb/src/core/registry/client/http.rs
THenry14: Changed in https://github.com/software-mansion/scarb/pull/1570/commits/148f40b897c820a0b647814661325600e2b92de9

scarb/src/core/registry/client/http.rs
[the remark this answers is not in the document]
THenry14: Don't thin deref will work, as stream needs an object BUT this is actually very cool idea! I have added an `into_file` to async lock to assume the lock ad used it alongside the idea of converting it to async and it seems to be working :D thanks for this pointer, really appreciate this 🙏 been trying to figure this out longer than I should 😅 

Added change in: https://github.com/software-mansion/scarb/pull/1570/commits/6c399981822e0417f0bcf18ce91b3e5e134efe80

scarb/src/core/registry/client/http.rs
THenry14: Hope I understood you correctly, here's the change, please take a look: https://github.com/software-mansion/scarb/pull/1570/commits/b23ad3dbf0e027e474bf70343fe558b67499692d

scarb/src/core/registry/client/http.rs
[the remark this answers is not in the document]
THenry14: Sure, thanks! changed in https://github.com/software-mansion/scarb/pull/1570/commits/67df1c878ec11e1e2c21efc6d73048b0464ca9f0

utils/scarb-test-support/src/simple_http_server.rs
mkaput: not insisting, do as you wish. just finding amount of stuff you're changing alerting

scarb/src/core/registry/client/http.rs
mkaput: 1. clap support env vars
2. I asked whether this is temporary because expecting people to set env manually is bad UX imho

scarb/src/core/registry/client/http.rs
mkaput: I can imagine one creating a custom registry with a higher package size limit. In such scenario, we'd be the bad guys for hardcoding some arbitrary limit client-side.

Won't relying on the registry to reject post requests with huge content-size be enough? (+/-)

scarb/src/core/registry/client/http.rs
mkaput: I would rather have no timeout at all and instead show a message that upload is in progress but it's running slow after some amount of time

scarb/src/core/registry/client/http.rs
THenry14: Hmm, ok I thought about it from devops and security pov - I wouldn't ever want my tokens or passwords to be passed as flags to any program. Just having this as plain text in bash_history is bad enough, so I actually disagree with you on that points. They are valid in most cases, but not sure in this particular one.

extensions/scarb-snforge-test-collector/src/main.rs
```diff
@@ -52,7 +52,7 @@ fn main() -> Result<()> {
 
         // artifact saved to `{target_dir}/{profile_name}/{package_name}.sierra.json`
         let output_path =
-            snforge_target_dir.join(&format!("{}.snforge_sierra.json", package_metadata.name));
+            snforge_target_dir.join(format!("{}.snforge_sierra.json", package_metadata.name));
```
mkaput, line 55: unrelated change?

scarb/src/flock.rs
```diff
@@ -100,6 +100,12 @@ impl AsyncFileLockGuard {
         self.lock_kind
     }
 
+    pub fn into_file(mut self) -> tokio::fs::File {
+        self.file
+            .take()
+            .expect("failed to take ownership of the file")
+    }
+
```
mkaput, line 108: this is illegal. you will forget to call `unlock` on the file, which is done in `Drop` for (Async)FileLockGuard and is the exact purpose of existence of these types.

I think what you should do is to open the file as a stream and then make a `Body` out of it with https://docs.rs/reqwest/latest/reqwest/struct.Body.html#method.wrap_stream

mkaput (changes requested):

scarb/src/core/registry/client/http.rs
THenry14: The registry would then probably need to accept the file, in order to check its metadata, right? so the resources would already be wasted. It's not a big deal, but still. 
Looking at the package sizes atm we are comfortably in the < 100Kb region, just for context. I know things may change, but on the other hand I really don't think they will anytime soon, if ever.

extensions/scarb-snforge-test-collector/src/main.rs
THenry14: CI fails because of this, not sure why. I did a rebase to main, and it still failed: https://github.com/software-mansion/scarb/actions/runs/10735766063/job/29773752730

scarb/src/flock.rs
THenry14: tried to do this after your previous comment, but was unable to, as the stream needs an ownership of the file (and lock), and all I could do was a reference. I also tried to do a ReadStream out of it, but no dice :( 
Perhaps I did something wrong, I do not exclude this possibility, but this solution sadly did not work for windows :(

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +150,81 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
+        Ok(self.index_config.load().await?.upload.is_some())
     }
 
-    async fn publish(&self, _package: Package, _tarball: FileLockGuard) -> Result<()> {
-        todo!("Publishing to HTTP registries is not implemented yet.")
+    async fn publish(&self, package: Package, tarball: FileLockGuard) -> Result<RegistryUpload> {
+        let auth_token = env::var("SCARB_REGISTRY_AUTH_TOKEN").map_err(|_| {
+            anyhow!(
+                "missing authentication token. \
+            help: make sure SCARB_RELEASE_AUTH_TOKEN environment variable is set"
```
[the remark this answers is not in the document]
THenry14: already done in https://github.com/software-mansion/scarb/pull/1570/commits/c6d58254dc142349900f9651920625879f6e50de

scarb/src/core/registry/client/http.rs
THenry14: > And if there’s already a server-side limit, adding the same check client-side feels redundant.

Please correct me if I'm wrong, but I don't think there is such limit atm

> but we're already doing that when checking that:

I think those are quite different cases. 
Moreover I'm pretty sure that checks for missing dependencies are done locally (granted - registry is being queried, but this is not something that registry does when package is uploaded, which is the case with publish). Version duplication or keywords limit is not resource intensive. 

I do understand your overall points though, and I do not disagree with them. We can try and look into file upload validation on django side, to see it we can somehow manage to do that before upload starts (I imagine this is quite common problem, so perhaps a solution alredy exists). I can raise an issue, and remove this check from scarb for time being if you wish so. sounds good?

website/docs/registries/publishing.md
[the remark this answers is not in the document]
THenry14: Sure, changed in https://github.com/software-mansion/scarb/pull/1570/commits/4762f55d195871a390508b16ea89f374f8adbf2e

website/docs/registries/publishing.md
```diff
@@ -1,38 +1,25 @@
...
-If you want to speed up the packaging process, you can disable this step using the `--no-verify` flag.
+## Publishing the package
 
-> [!WARNING]
-> This is a dangerous operation as it can lead to uploading a corrupted package to the registry.
-> Please use with caution.
+To upload your package, use the scarb publish command. By default, this command will publish your package to the official [scarbs.xyz](https://scarbs.xyz) registry. The publish command automatically [packages and verifies](#packaging-your-package) your package, so there is no need to run `scarb package` beforehand.
 
-After successfully completing the whole process, the `{name}-{version}.tar.zst` archive waits in the `target/package` directory for being uploaded, where both `name` and `version` correspond to the values in `Scarb.toml`.
+To publish your package to a registry that supports package publishing, you need to authenticate using an API token with the `publish` scope. First, log in to the registry and generate this API token. Scarb will use the token to authenticate and complete the publishing process. The token must be provided via the ``SCARB_REGISTRY_AUTH_TOKEN` environment variable.
```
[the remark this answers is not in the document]
THenry14: good, added in https://github.com/software-mansion/scarb/pull/1570/commits/163dea1f2fae17537b2525ba4cbe8154280d6e41

scarb/src/core/registry/client/http.rs
THenry14: Maybe this is an idea for a "feature"? :) 
regarding exports etc, yup this still leaks stuff. But what I meant is more of an "automatic" env, like some VMs created with env variables, CIs etc :+1:

scarb/src/core/registry/client/http.rs
```diff
@@ -146,12 +150,81 @@ impl<'c> RegistryClient for HttpRegistryClient<'c> {
...
+            .send()
+            .await?;
+
+        match response.status() {
+            StatusCode::UNAUTHORIZED => Err(RegistryUpload::Unauthorized)
+                .map_err(|_| anyhow!("invalid authentication token")),
+            StatusCode::FORBIDDEN => Err(RegistryUpload::CannotPublish)
+                .map_err(|_| anyhow!("missing upload permissions or not the package owner")),
+            StatusCode::BAD_REQUEST => Err(RegistryUpload::VersionExists)
+                .map_err(|_| anyhow!("package `{}` already exists", &package.id)),
```
DelevoXDG, line 206: This seems problematic because there can be a mismatch between the actual error and what Scarb displays. For example, uploading package with 11 keywords using CURL will result in the following error:
```json
{"status":400,"error":"['Expected at most 10 keywords per package.']"}
```

But uploading the same package via `scarb publish` results a different, misleading error message:
```sh
error: package `access_control v0.3.5 ([..]/access_control/Scarb.toml)` already exists
```

Maybe we should consider @mkaput’s [suggestion](https://github.com/software-mansion/scarb/pull/1570/files#r1745749154) to use the `anyhow::Error` objects and propagate errors directly from the registry instead of producing our own.

Otherwise, even if we address this case, future errors returned by registry could still end up getting converted into generic `“file corrupted during upload.”`.

What do you think?

scarb/src/core/registry/client/http.rs
DelevoXDG: > Maybe this is an idea for a "feature"? :)

Sure, just an idea to address both your and Marek's concerns, i.e. to both have no leaks and good UX. Added #1579 to further discuss this.

> But what I meant is more of an "automatic" env, like some VMs created with env variables, CIs etc 👍

Yeah, I understand, make sense

scarb/src/core/registry/client/http.rs
THenry14: Hmm, that might be a good reason to apply the suggestion. Indeed keywords checks were added in the meantime, and they utilise the same error code. 

I do wonder however, if we should reuse error codes for completely different potential errors 🤔 I guess registry should return different codes for different problems, right?

scarb/src/core/registry/client/http.rs
DelevoXDG: > Please correct me if I'm wrong, but I don't think there is such limit atm

Yes, we don't have it atm 😅 what I meant is that we definitely *should have it* on the registry side, so there will be unnecessary duplication anyway

> I can raise an issue, and remove this check from scarb for time being if you wish so. sounds good?

Sounds good

scarb/src/bin/scarb/args.rs
THenry14: Unresolving this because I needed to make some further changes. It appeared that source id taken from package is being wrong (it's a path). So in order to get the default index I did https://github.com/software-mansion/scarb/pull/1570/commits/ef6b9ad6ba85e264b44f52ccdab3dd2111c5e424

scarb/src/core/registry/client/http.rs
DelevoXDG: > I do wonder however, if we should reuse error codes for completely different potential errors 🤔 I guess registry should return different codes for different problems, right?

I get where you’re coming from, but IIUC we're dealing with HTTP errors codes here, which are meant to capture wide range of errors under the same error code, and it's [common](https://stackoverflow.com/questions/3290182/which-status-code-should-i-use-for-failed-validations-or-invalid-duplicates) to use `400 Bad Request` for validation errors.

Even if consider using `422` as suggested in the thread I linked above, we'd still likely to encounter same error codes for different errors, including those generated internally by Django.

We might consider defining our own non-http error codes for our custom errors (~~starknet~~ scarbs spec 😄), but I'm not sure that's really necessary right now.

scarb/src/core/registry/client/http.rs
mkaput: I meant, you can expect env vars in clap arguments without exposing them as parameters, just don't use `short` nor `long` in relevant `#[arg]` use

scarb/src/core/registry/client/http.rs
mkaput: yeah, that TODO was from times I wasn't aware of slow internet problems

this is not urgent

scarb/src/flock.rs
mkaput: hmm, maybe you could use https://docs.rs/tokio/latest/tokio/fs/struct.File.html#method.try_clone then? 

Under the hood, this boils down to call to `fcntl` (`man fcntl`) with `F_DUPFD_CLOEXEC` on Unix and [`DuplicateHandle`](https://learn.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-duplicatehandle) on Windows. From quickly scanning docs it appears to me that this can only fail when we'll exceed the open file descriptor limit in the process, which is very unlikely.

scarb/src/core/registry/client/mod.rs
mkaput: see @DelevoXDG argument, it'd be easier to just pass error messages from the registry etc.

scarb/src/flock.rs
THenry14: Hmm, interesting, it creates a new instance 🤔 seems to be working on windows as well, thanks :+1: https://github.com/THenry14/scarb/actions/runs/10740797646/job/29789733604
will read into this a bit more to better understand what happens :+1: Will of course make a change shortly

scarb/src/core/registry/client/http.rs
THenry14: Sure, I think this is a bit misleading then

> I asked whether this is temporary because expecting people to set env manually is bad UX imho

as users will still have to set this manually, right? so actually nothing changes from UX perspective?

scarb/src/core/registry/client/http.rs
THenry14: great, created an issue https://github.com/software-mansion/scarb/issues/1580
and removed the timeout in https://github.com/software-mansion/scarb/pull/1570/commits/c0077eaf95f63b9f925b0538909b246bff4fe4d3

scarb/src/flock.rs
THenry14: Done in https://github.com/software-mansion/scarb/pull/1570/commits/687ab6004fec9cc8b82acfa6fe3d889f043678ad

scarb/src/core/registry/client/http.rs
THenry14: Great. Created an issue: https://github.com/software-mansion/scarbs.xyz/issues/352
and removed said code: https://github.com/software-mansion/scarb/pull/1570/commits/9d900b90d48df046bb6f3d5aa3d658e4b6420352

mkaput (approved):
lgtm from my side, though I didn't review this carefully (only picked stuff that interested me)

scarb/src/core/registry/client/http.rs
THenry14: Carrying over from similar thread with Marek.
Yeah, you're right. It's better to just pass through the errors, and depend on codes and messages from the registry. I have refactored the enum and adjusted the code https://github.com/software-mansion/scarb/pull/1570/commits/a9f383a41251ca3d139048aa2bab8640fefbf2c3

maciektr (approved):

PR #1571, opened 2024-09-02
Fix unnecessary downloads of cached registry packages

### LINKED ISSUE #1544: Potentially unnecessary downloads from registry when rebuilding
### Problem

From Alex:

Compiling the same project, w/o changing a single bit, sometimes re-downloads packages from registry (sometimes).
While this made some sense with github tags which could have changed, there is no intention for registry tags to have any different code.
Besides slower build times, this also blocks building if the user goes offline

First build: downloads deps

```
[~/Code/Cairo1/Foundry/starknet-foundry-template]> scarb build
 Downloading openzeppelin_token v0.15.1
 Downloading openzeppelin_introspection v0.15.1
 Downloading openzeppelin_account v0.15.1
 Downloading openzeppelin_governance v0.15.1
 Downloading openzeppelin_utils v0.15.1
 Downloading openzeppelin_access v0.15.1
   Compiling snforge_scarb_plugin v0.1.0 (git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.28.0#4dfe39d96690ed6b3d56971512700de3f58288ea)
    Finished `release` profile [optimized] target(s) in 0.16s
   Compiling lib(token_sender) token_sender v0.5.0 (/home/amanusk/Code/Cairo1/Foundry/starknet-foundry-template/Scarb.toml)
   Compiling starknet-contract(token_sender) token_sender v0.5.0 (/home/amanusk/Code/Cairo1/Foundry/starknet-foundry-template/Scarb.toml)
    Finished release target(s) in 25 seconds
```

Second build: no download :slightly_smiling_face:

```
[~/Code/Cairo1/Foundry/starknet-foundry-template]> scarb build
   Compiling snforge_scarb_plugin v0.1.0 (git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.28.0#4dfe39d96690ed6b3d56971512700de3f58288ea)
    Finished `release` profile [optimized] target(s) in 0.17s
   Compiling lib(token_sender) token_sender v0.5.0 (/home/amanusk/Code/Cairo1/Foundry/starknet-foundry-template/Scarb.toml)
   Compiling starknet-contract(token_sender) token_sender v0.5.0 (/home/amanusk/Code/Cairo1/Foundry/starknet-foundry-template/Scarb.toml)
    Finished release target(s) in 22 seconds
```    
    
Third build : downloads again :disappointed:

```
[~/Code/Cairo1/Foundry/starknet-foundry-template]> scarb build
 Downloading openzeppelin_token v0.15.1
 Downloading openzeppelin_introspection v0.15.1
 Downloading openzeppelin_account v0.15.1
 Downloading openzeppelin_governance v0.15.1
 Downloading openzeppelin_utils v0.15.1
 Downloading openzeppelin_access v0.15.1
   Compiling snforge_scarb_plugin v0.1.0 (git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.28.0#4dfe39d96690ed6b3d56971512700de3f58288ea)
    Finished `release` profile [optimized] target(s) in 0.19s
   Compiling lib(token_sender) token_sender v0.5.0 (/home/amanusk/Code/Cairo1/Foundry/starknet-foundry-template/Scarb.toml)
```

Forth build w/o wifi: build fails :disappointed:

```
[~/Code/Cairo1/Foundry/starknet-foundry-template]> scarb build
error: failed to lookup for `openzeppelin_token ^0.15.1` in registry: registry+https://scarbs.xyz/

Caused by:
    0: error sending request for url (https://scarbs.xyz/api/v1/index/op/en/openzeppelin_token.json): error trying to connect: dns error: failed to lookup address information: Name or service not known
    1: error trying to connect: dns error: failed to lookup address information: Name or service not known
    2: dns error: failed to lookup address information: Name or service not known
    3: failed to lookup address information: Name or service not known
```

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

_No response_

DelevoXDG:
Closes #1544

### DIFF
--- a/scarb/src/core/registry/client/cache.rs
+++ b/scarb/src/core/registry/client/cache.rs
@@ -145,7 +145,7 @@ impl<'c> RegistryClientCache<'c> {
             let tarball_name = package.tarball_name();
             let file = self
                 .dl_fs
-                .open_rw(&tarball_name, &tarball_name, self.config)?;
+                .open_ro(&tarball_name, &tarball_name, self.config)?;
             let checksum = self.get_record_maybe_uncached(package).await?.checksum;
             return Ok((file, checksum));
         }
--- a/scarb/tests/http_registry.rs
+++ b/scarb/tests/http_registry.rs
@@ -309,6 +309,15 @@ fn caching() {
         .success()
         .stdout_eq("");
 
+    Scarb::quick_snapbox()
+        .arg("fetch")
+        .env("SCARB_CACHE", cache_dir.path())
+        .current_dir(&t)
+        .timeout(Duration::from_secs(10))
+        .assert()
+        .success()
+        .stdout_eq("");
+
     let expected = expect![[r#"
         GET /api/v1/index/config.json
         accept: */*
@@ -352,6 +361,19 @@ fn caching() {
         content-type: application/octet-stream
         etag: ...
         last-modified: ...
+
+        ###
+
+        GET /index/3/b/bar.json
+        accept: */*
+        accept-encoding: gzip, br, deflate
+        host: ...
+        if-none-match: ...
+        user-agent: ...
+
+        304 Not Modified
+        content-length: 0
+        etag: ...
 
         ###
 


scarb/src/core/registry/client/cache.rs
```diff
@@ -145,7 +145,7 @@ impl<'c> RegistryClientCache<'c> {
             let tarball_name = package.tarball_name();
             let file = self
                 .dl_fs
-                .open_rw(&tarball_name, &tarball_name, self.config)?;
+                .open_ro(&tarball_name, &tarball_name, self.config)?;
```
DelevoXDG, line 148: This is necessary as `open_rw` truncates the existing contents of the file, effectively destroying the cached tarball: 

>    /// This function will create a file at `path` if it doesn't already exist (including
>    /// intermediate directories) else if it does exist, it will be truncated.
>    ...
>    pub fn open_rw(

Since all tests pass and `scarb fetch`, `scarb build` work without issue, write access might not be necessary when accessing a cached archive. We could consider creating an alternative method that provides write access without truncating the file, but unless there's a specific need for this, it seems to me that read-only access should be sufficient in this context.

DelevoXDG: Also, it's not very intuitive from the naming that `open_rw` is supposed to truncate anything 🤔

mkaput: You can fire git blame and see that truncating was added after this function was created :) There was a relatively significant amount of consideration going into that name and the conclusion was that in Scarb context: you will (almost) never want to write to a file without overwriting it.

maciektr (approved):
💪 💪 💪

scarb/src/core/registry/client/cache.rs
DelevoXDG: > You can fire git blame and see that truncating was added after this function was created :) 

Thanks for clarification! Although I assume the discussion happened offline rather then in the [PR](https://github.com/software-mansion/scarb/pull/388).

> you will (almost) never want to write to a file without overwriting it.

I understand the rationale behind *typically* wanting to overwrite files, but I think "almost" is the key and the difference between intended and actual use can lead to bugs that can tricky to spot 😅 . Maybe we should consider changing it to something like `create_open_rw` or `create_rw`, just to fully convey what this does and be safe in the future. Thoughts?

mkaput: I'm fine with that

maciektr: This can be a separate PR, please merge this one 🙏

PR #1578, opened 2024-09-04
Add `cargo xtask set-cairo-version --path`

mkaput:
This small utility seems very helpful when setting up custom Scarb
checkout for testing some work on the Cairo compiler.


maciektr (approved):
💪

PR #1592, opened 2024-09-13
Make paths displed on left nav bar if duplicated

wawel37:


extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -153,6 +169,26 @@ impl MarkdownDocItem for Trait {
...
+    let mut names_counter = HashMap::<String, u32>::new();
+    for item in items {
+        *names_counter.entry(item.name().to_string()).or_insert(0) += 1;
+    }
+
+    items
+        .iter()
+        .map(|item| {
+            let is_duplicated = match names_counter.get(item.name()) {
+                Some(value) => *value > 1,
```
maciektr, line 184: Why not just use `HashSet` instead of a map?

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -14,12 +15,27 @@ pub trait TopLevelMarkdownDocItem: MarkdownDocItem + TopLevelDocItem {
...
+        path_tree_elements.pop();
+        path_tree_elements.join("::")
     }
 
-    fn generate_markdown_list_item(&self) -> String {
-        format!("- {}\n", self.md_ref())
+    fn md_ref(&self, display_parent_path: bool) -> String {
+        if display_parent_path {
+            format!(
+                "[{}  ({})](./{})",
```
maciektr, line 27: Wouldn't it actually be more readable, to just print the path? (like `hello_world::sub_module::duplicated_function_name` instead of this weird brackets `duplicated_function_name (hello_world::sub_module)`)

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -14,12 +15,27 @@ pub trait TopLevelMarkdownDocItem: MarkdownDocItem + TopLevelDocItem {
...
+                self.name(),
+                self.parent_path(),
```
maciektr, line 29: Maybe instead of printing full path on duplicates, we could print only the path from the node where duplicate happens? Like in case of `hello:first::dup` and `hello::second::dup` we print `first::dup` and `second::dup`? WDYT?

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
wawel37: We need to know whether there are more than 1 item with the same name. With Set, we are unable to tell if there's a single item (not duplicated), or more of them. Or I missed something? We would need 2 sets, the `observed_set` and `duplicated_set`

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
wawel37: You mean just printing whole path (without the name, as it's inside the path)? If so, I think it might work, will check this out and see how does it look like

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
wawel37: Hmmm, I think we could do that. Will work on that

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
maciektr: Hmm, I am not sure what I envisioned. HashMap is good :D

maciektr (approved):
LGTM! 🦾

PR #1593, opened 2024-09-13
Fix single package features in a workspace

### LINKED ISSUE #1516: Make features work the way as in `cargo doc`
As for now, we strictly check whether the sub packages have the features from args inside the manifests. We want to make it work as in `cargo doc`. We want to ignore them (if they are missing in manifest), rather than throwing an error.

maciektr:
fix #1516 #1599

### DIFF
--- a/scarb/src/ops/resolve.rs
+++ b/scarb/src/ops/resolve.rs
@@ -173,7 +173,12 @@ pub fn generate_compilation_units(
     ws: &Workspace<'_>,
 ) -> Result<Vec<CompilationUnit>> {
     let mut units = Vec::with_capacity(ws.members().size_hint().0);
-    for member in ws.members().filter(|member| !member.is_cairo_plugin()) {
+    let members = ws
+        .members()
+        .filter(|member| !member.is_cairo_plugin())
+        .collect_vec();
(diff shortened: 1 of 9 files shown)


maciektr:
@wawel37 should this fix https://github.com/software-mansion/scarb/issues/1516?

scarb/src/ops/resolve.rs
```diff
@@ -207,6 +212,25 @@ pub fn generate_compilation_units(
...
+pub fn validate_features(members: &[Package], enabled_features: &FeaturesOpts) -> Result<()> {
+    // Check if any member has features defined.
+    if let FeaturesSelector::Features(features) = &enabled_features.features {
+        for feature in features {
+            if !members
+                .iter()
+                .any(|member| member.manifest.features.contains_key(feature))
+            {
+                bail!(
+                    "none of selected packages contains `{}` feature\n\
```
wawel37, line 224:
```suggestion
                    "none of the selected packages contains `{}` feature\n\
```

wawel37 (approved):
LGTM

wawel37:
> @wawel37 should this fix #1516?

Yup! If we check for any member with given feature, it will work for sure!

Issue #1509, opened 2024-08-06
Build signatures without trivia


piotmag769:
Currently signatures in `scarb-doc` sometimes contain trivia, particularly comments. Example:

```rs
struct MyStruct{
  /// Field.
  field: u8
}
```

This happens because the logic of fetching the signatures is basically getting all the text from the node with item definition. We should instead: traverse the subtree of the node, get text without trivia from each leaf, concatenate it together (remember about adding whitespaces where neccesary) and format it.

Note: if we decide to do #1426 first/in near future then this task becomes irrelevant

PR #1601, opened 2024-09-17
Add signature for struct's member

wawel37:
Fixes [#1509](https://github.com/software-mansion/scarb/issues/1509). It waits for [this](https://github.com/starkware-libs/cairo/pull/6387) to be merged in Cairo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -568,7 +568,7 @@ dependencies = [
 [[package]]
 name = "cairo-lang-casm"
 version = "2.8.2"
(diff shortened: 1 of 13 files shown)


piotmag769:
@wawel37 are you sure it resolves #1509 after the PR to compiler is merged? Wouldn't trivia be still there for other signatures?

wawel37:
> @wawel37 are you sure it resolves #1509 after the PR to compiler is merged? Wouldn't trivia be still there for other signatures? I also believe it would still be there for structs bcs you don't call for the signature of members when generating struct signature

You're right. Will create separate Cairo PR to support signatures for structs as a whole.

maciektr (approved):

PR #1604, opened 2024-09-18
Remove redundant call to DiagnosticsReporter::ensure

maciektr:


scarb/src/compiler/compilers/test.rs
```diff
@@ -32,11 +32,9 @@ impl Compiler for TestCompiler {
                 && plugin.package.id.source_id == SourceId::for_std()
         });
 
-        let mut diagnostics_reporter =
+        let diagnostics_reporter =
             build_compiler_config(db, &unit, &test_crate_ids, ws).diagnostics_reporter;
```
wawel37, line 36: Nit: Maybe not related to this PR, but as I looked into the `build_compiler_config` func, there are 2 big logic parts, one for creating diagnostics reporter, and the second one is responsible for extracting different fields from CompilationUnit. Maybe we can make a diagnostic creation logic to separate function, so we don't have to call whole `build_compiler_config` function just to get the diagnostic_reporter. I am aware that there isn't a lot of complexity inside compiler config creation, but in future, how knows.

wawel37 (approved):
LGTM! No point to check for diagnostic errors twice :D

scarb/src/compiler/compilers/test.rs
maciektr: Yeah, although it basically boils down to creating three unnecessary booleans. I see your point though and I would say the compiler api of the `compile_test_prepared_db` is generally a bit cumbersome. We might want to change it in a future.

PR #1616, opened 2024-09-26
Dependencies crate settings

wawel37:


extensions/scarb-doc/src/metadata/compilation.rs
```diff
@@ -131,10 +135,30 @@ fn get_crate_settings_for_package(
...
     };
 
+    let dependencies = package
+        .dependencies
+        .iter()
+        .map(|dependency| {
+            let version = packages
+                .iter()
+                .find(|package| package.name == dependency.name)
+                .unwrap()
```
maciektr, line 145: I would rather do `compilation_unit_metadata .components` instead of `metadata.packages` here and skip the package if find returns `None`. My point is, that the `components` is what is fed to the compiler as crate roots, not `metadata.packages` (which can be a superset of `components`).

extensions/scarb-snforge-test-collector/src/metadata.rs
```diff
@@ -226,10 +231,30 @@ fn get_crate_settings_for_package(
...
     };
 
+    let dependencies = package
+        .dependencies
+        .iter()
+        .map(|dependency| {
+            let version = packages
+                .iter()
+                .find(|package| package.name == dependency.name)
+                .unwrap()
```
maciektr, line 241: vide supra

scarb/src/compiler/db.rs
```diff
@@ -134,12 +136,39 @@ fn build_project_config(unit: &CairoCompilationUnit) -> Result<ProjectConfig> {
...
+                .manifest
+                .summary
+                .full_dependencies()
+                .collect();
+
+            // We iterate over all of the compilation unit components to get dependency's version.
+            let dependencies = unit
+                .components
+                .iter()
+                .filter(|component_as_dependency| {
```
maciektr, line 151: Have you verified that we don't need to add the component itself as the dep? (i.e. import stuff as `use package_name::...` in said package) Because it would not pop up in `full_dependencies`, right? 🤔

extensions/scarb-doc/src/metadata/compilation.rs
wawel37: Need to use packages because the compilation unit component has not version info.

extensions/scarb-snforge-test-collector/src/metadata.rs
wawel37: same as above

scarb/src/compiler/db.rs
wawel37: Right, will change that

maciektr (approved):

PR #1620, opened 2024-09-30
Doc/support reexported modules

wawel37:


extensions/scarb-doc/src/types.rs
```diff
@@ -21,6 +23,21 @@ use cairo_lang_syntax::node::{ast, SyntaxNode};
...
+    Vec<EnumId>,
+    Vec<ModuleTypeAliasId>,
+    Vec<ImplAliasId>,
+    Vec<TraitId>,
+    Vec<ImplDefId>,
+    Vec<ExternTypeId>,
+    Vec<ExternFunctionId>,
+    Vec<SubmoduleId>,
+    Vec<CrateId>,
+);
```
maciektr, line 39: I would enclose this and the extraction logic into a new struct instead, for better readability.  
For instance something like this: https://pastebin.swmansion.com/?f93d580e1791f3fc#2QyVDKnmrJyPnwdQjY3kQhnbgecmaFcksy9LibEoD13v

maciektr (commented):
I don't have any objections against the logic. I understand a lot of the changes will go away after rebasing onto main, right?

maciektr (approved):

PR #1626, opened 2024-10-02
feat: scarb fmt a subdirectory

### LINKED ISSUE #1534: feat: `scarb fmt` option to target a specific file.
### Problem

# Describe the Feature Request

Following https://github.com/starkware-libs/cairo/issues/4134, for integration with external tools, cairo-format should have an option to take a specific file as a target, instead of formatting the entire package all at once.

For reference, with the old CairoZero cairo-format, we could do cairo-format ${target} which would format one specific file.
Forge (foundry) supports forge fmt ${target} --check as well. This works great for integration with lint aggregators.

# Describe Preferred Solution

Add an optional target argument, referring to the file to format. The command scarb fmt ${target} -e stdout should be valid. If a target is specfied, the name of the file being formatted is not part of the stdout.

cairo-format actually supports this; but scarb fmt doesn't: see below

```
❯ cargo run --bin cairo-format -- --help
Formats a file or directory with the Cairo formatter.
Exits with 0/1 if the input is formatted correctly/incorrectly.

Usage: cairo-format [OPTIONS] [FILES]...

Arguments:
  [FILES]...  A list of files and directories to format. Use "-" for stdin

Options:
  -c, --check                 Check mode, don't write the formatted files, just output the diff between the original and the formatted file
  -r, --recursive             Format directories content recursively
  -v, --verbose               Print verbose output
  -p, --print-parsing-errors  Print parsing errors
  -s, --sort-mod-level-items  Enable sorting the module level items (imports, mod definitions...)
  -h, --help                  Print help
  -V, --version               Print version
  ```
  
  However it's not shown when doing `scarb fmt --help`
  
  ```
  Usage: scarb fmt [OPTIONS]

Options:
  -c, --check                  Only check if files are formatted, do not write the changes to disk
  -e, --emit <EMIT>            Emit the formatted file to stdout [possible values: stdout]
  -v, --verbose...             Increase logging verbosity.
      --no-color               Do not color output
  -q, --quiet...               Decrease logging verbosity.
  -p, --package <SPEC>         Packages to run this command on, can be a concrete package name (`foobar`) or a prefix glob (`foo*`) [env: SCARB_PACKAGES_FILTER=] [default: *]
      --verbosity <VERBOSITY>  Set UI verbosity level by name. [env: SCARB_UI_VERBOSITY=]
  -w, --workspace              Run for all packages in the workspace
  -h, --help                   Print help
  ```
  
  Delegating this to the scarb team
  
  

### Proposed Solution

Either modify scarb fmt to accept a specific file path OR expose cairo-format through scarb (like scarb cairo-run does)

### Notes

_No response_

### PR DESCRIPTION
**Pull request by enitrat**
Allows formatting a single file. Useful for integration with external tools.

Closes #1534

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -342,8 +342,9 @@ pub struct FmtArgs {
     /// Specify package(s) to format.
     #[command(flatten)]
     pub packages_filter: PackagesFilter,
-    #[clap(value_name = "TARGET")]
-    pub target: Option<PathBuf>,
+    /// Path to a file or directory to format. If provided, only this file or directory will be formatted.
+    #[clap(value_name = "PATH")]
+    pub target_path: Option<Utf8PathBuf>,
 }
 
 /// Arguments accepted by the `add` command.
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -343,7 +343,7 @@ pub struct FmtArgs {
     pub packages_filter: PackagesFilter,
     /// Path to a file or directory to format. If provided, only this file or directory will be formatted.
     #[clap(value_name = "PATH")]
-    pub target_path: Option<Utf8PathBuf>,
+    pub path: Option<Utf8PathBuf>,
 }
 
 /// Arguments accepted by the `add` command.
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -4,6 +4,7 @@
 
 use std::collections::BTreeMap;
 use std::ffi::OsString;
+use std::path::PathBuf;
 
 use anyhow::Result;
 use camino::Utf8PathBuf;
@@ -341,6 +342,8 @@ pub struct FmtArgs {
     /// Specify package(s) to format.
     #[command(flatten)]
     pub packages_filter: PackagesFilter,
+    #[clap(value_name = "TARGET")]
+    pub target: Option<PathBuf>,
 }
 
 /// Arguments accepted by the `add` command.
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -4,7 +4,6 @@
 
 use std::collections::BTreeMap;
 use std::ffi::OsString;
-use std::path::PathBuf;
 
 use anyhow::Result;
 use camino::Utf8PathBuf;
--- a/scarb/src/ops/fmt.rs
+++ b/scarb/src/ops/fmt.rs
@@ -58,10 +58,7 @@ pub fn format(opts: FmtOptions, ws: &Workspace<'_>) -> Result<bool> {
             format_package(&pkg, &opts, ws, &all_correct)?;
             return Ok(all_correct.load(Ordering::Acquire));
         } else {
-            return Err(anyhow::anyhow!(
-                "File {:?} is not part of the workspace.",
-                target_path
-            ));
+            anyhow::bail!("path {:?} is not part of the workspace.", target_path);
         };
     };
     for package_id in opts.packages.iter() {
@@ -112,7 +109,7 @@ fn format_single_file(
     let pkg = ws
         .members()
         .find(|member| target_file.starts_with(member.root()))
-        .ok_or_else(|| anyhow::anyhow!("File {:?} is not part of the workspace.", target_file))?;
+        .ok_or_else(|| anyhow::anyhow!("file {:?} is not part of the workspace.", target_file))?;
 
     let mut config = FormatterConfig::default();
     if let Some(overrides) = pkg.tool_metadata("fmt") {
@@ -128,6 +125,7 @@ fn format_single_file(
             target,
             ws,
             target_file.as_std_path(),
+            EmitMode::WithoutPath,
         ),
     };
 
@@ -235,18 +233,27 @@ impl Message for TextWithNewline {
     }
 }
 
+pub enum EmitMode {
+    WithPath,
+    WithoutPath,
+}
+
 pub trait Emittable {
-    fn emit(&self, ws: &Workspace<'_>, path: &Path, formatted: &str);
+    fn emit(&self, ws: &Workspace<'_>, path: &Path, formatted: &str, emit_mode: EmitMode);
 }
 
 impl Emittable for FmtEmitTarget {
-    fn emit(&self, ws: &Workspace<'_>, path: &Path, formatted: &str) {
-        match self {
-            Self::Stdout => ws.config().ui().print(TextWithNewline(format!(
+    fn emit(&self, ws: &Workspace<'_>, path: &Path, formatted: &str, emit_mode: EmitMode) {
+        match (self, emit_mode) {
+            (Self::Stdout, EmitMode::WithPath) => ws.config().ui().print(TextWithNewline(format!(
                 "{}:\n\n{}",
                 path.display(),
                 formatted
             ))),
+            (Self::Stdout, EmitMode::WithoutPath) => ws
+                .config()
+                .ui()
+                .print(TextWithNewline(format!("{}", formatted))),
         }
     }
 }
@@ -256,14 +263,15 @@ fn emit_formatted_file(
     target: &dyn Emittable,
     ws: &Workspace<'_>,
     path: &Path,
+    emit_mode: EmitMode,
 ) -> bool {
     match fmt.format_to_string(&path) {
         Ok(FormatOutcome::Identical(original)) => {
-            target.emit(ws, path, &original);
+            target.emit(ws, path, &original, emit_mode);
             true
         }
         Ok(FormatOutcome::DiffFound(diff)) => {
-            target.emit(ws, path, &diff.formatted);
+            target.emit(ws, path, &diff.formatted, emit_mode);
             false
         }
         Err(parsing_error) => {
@@ -320,7 +328,7 @@ impl<'t> ParallelVisitor for PathFormatter<'t> {
             FmtAction::Fix => format_file_in_place(self.fmt, self.opts, self.ws, path),
             FmtAction::Check => check_file_formatting(self.fmt, self.opts, self.ws, path),
             FmtAction::Emit(target) => {
-                emit_formatted_file(self.fmt, target, self.ws, path)
+                emit_formatted_file(self.fmt, target, self.ws, path, EmitMode::WithPath)
             }
         };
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: scarb/src/ops/fmt.rs [hunk 1]
```diff
@@ -66,6 +71,28 @@ pub fn format(opts: FmtOptions, ws: &Workspace<'_>) -> Result<bool> {
     Ok(result)
 }
 
+fn format_single_file(
+    target_file: &Path,
+    opts: &FmtOptions,
+    ws: &Workspace<'_>,
+    all_correct: &AtomicBool,
+) -> Result<bool> {
+    let config = FormatterConfig::default();
```
maciektr, line 80: This should also do 
```
        let mut config = FormatterConfig::default();
        if let Some(overrides) = package.tool_metadata("fmt") {
            config = toml_merge(&config, overrides)?;
        }
```
To allow setting formatting opts through scarb manifest.

scarb/src/bin/scarb/args.rs
```diff
@@ -341,6 +342,8 @@ pub struct FmtArgs {
     /// Specify package(s) to format.
     #[command(flatten)]
     pub packages_filter: PackagesFilter,
+    #[clap(value_name = "TARGET")]
```
maciektr, line 345: This should have a doc that will be shown on `scarb fmt --help`. 
Maybe something like: 
```suggestion
    /// The target file for the formatter. If provided, only this file will be formatted.  
    #[clap(value_name = "TARGET")]
```

scarb/src/bin/scarb/args.rs
```diff
@@ -341,6 +342,8 @@ pub struct FmtArgs {
...
+    pub target: Option<PathBuf>,
```
maciektr, line 346:
```suggestion
    pub target: Option<Utf8PathBuf>,
```
Please use `Utf8PathBuf` instead

maciektr (commented):
Thanks for contributing! 👏

scarb/src/bin/scarb/args.rs
mkaput, line 346: I don't like this name. It collides with Scarb Targets.

Wdyt about `--file`?

enitrat:
Pushed a commit to avoid emitting the file path in emit mode. Behvaior is similar to foundry.

However I have an issue: it seems that in emit mode, the output has an extra empty line at the end. I can't find where this originates from?

scarb/src/bin/scarb/args.rs
enitrat: actually what about making this `PATH` and instead formatting the entire path - if that path happens to be a file, only a single file will be formatted?

scarb/src/bin/scarb/args.rs
enitrat: did something similar

enitrat:
> However I have an issue: it seems that in emit mode, the output has an extra empty line at the end. I can't find where this originates from?

I think it's just that `get_formatted_file` auto-inserts an empty line at EOF ?

scarb/src/ops/fmt.rs
enitrat: what should be the package passed here? do I need to iterate over all packages and infer from the provided path?

scarb/src/bin/scarb/args.rs
mkaput: sounds cool

scarb/src/ops/fmt.rs
maciektr: Yes, that's what I would do 
```
        let target_file = target_file.canonicalize_utf8()?;
        let pkg = ws
            .members()
            .find(|member| target_file.starts_with(member.root()));
```

scarb/src/ops/fmt.rs
```diff
@@ -155,16 +193,13 @@ fn check_file_formatting(
...
 
 impl Emittable for FmtEmitTarget {
-    fn emit(&self, ws: &Workspace<'_>, path: &Path, formatted: &str) {
+    fn emit(&self, ws: &Workspace<'_>, formatted: &str) {
         match self {
-            Self::Stdout => ws
-                .config()
-                .ui()
-                .print(format!("{}:\n{}", path.display(), formatted)),
+            Self::Stdout => ws.config().ui().print(format!("{}", formatted)),
```
maciektr, line 202: > I think it's just that get_formatted_file auto-inserts an empty line at EOF ?

Note that `.ui().print( {smth} )` will call `println!`. 

Not sure if this is a proper solution,  but you can override the behaviour with following: 
```
struct TextWithNewline(String);
impl Message for TextWithNewline {
    fn print_text(self)
    where
        Self: Sized,
    {
        print!("{}", self.0);
    }
}

(...)

ws.config().ui().print(TextWithNewline(format!("{}", formatted)));
```

scarb/src/ops/fmt.rs
```diff
@@ -178,7 +213,7 @@ fn emit_formatted_file(
     match fmt.format_to_string(&path) {
         Ok(FormatOutcome::Identical(_)) => true,
         Ok(FormatOutcome::DiffFound(diff)) => {
-            target.emit(ws, path, &diff.formatted);
+            target.emit(ws, &diff.formatted);
```
maciektr, line 216: Don't we need the paths printed,  when formatting multiple files at once? How do we know which is which?

enitrat: when in emit mode the standard behavior from `forge` is to just concatenate each file without any info on the origin.   I do not know of any tool that uses the multiple files at once directly, generally they just do `tool fmt <filepath>` and expect to find the re-formatted file in stdout.

enitrat: also: even if the file is addentical, in emit mode, it should be emitted to stdout

scarb/src/ops/fmt.rs
```diff
@@ -51,7 +59,12 @@ pub fn format(opts: FmtOptions, ws: &Workspace<'_>) -> Result<bool> {
         }
         let fmt = CairoFormatter::new(config);
 
-        let walk = fmt.walk(package.root().as_std_path());
+        let walk = if let Some(target_path) = &opts.target_path {
+            fmt.walk(target_path.as_std_path())
```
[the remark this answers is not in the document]
enitrat: indeed! good catch, I fixed this in the last commit

scarb/src/ops/fmt.rs
enitrat: thanks, did this

enitrat:
Pushed a version that works fine when integrated with Trunk :)

scarb/src/ops/fmt.rs
maciektr: Hmm, can we still print paths in case there is multiple files under the path? 
The `--emit` has been implemented as a community request. Not sure if any tools rely on this, but this is how `cargo fmt -- --emit stdout` does it as well.

enitrat: Done - Now the behavior is like cargo's:

```
path/to/file.cairo:

<file content>
```

scarb/src/ops/fmt.rs
```diff
@@ -42,28 +45,97 @@ pub fn format(opts: FmtOptions, ws: &Workspace<'_>) -> Result<bool> {
...
         let package = ws.fetch_package(package_id)?;
+        format_package(package, &opts, ws, &all_correct)?;
+    }
 
-        let mut config = FormatterConfig::default();
-        if let Some(overrides) = package.tool_metadata("fmt") {
-            config = toml_merge(&config, overrides)?;
-        }
-        let fmt = CairoFormatter::new(config);
+    Ok(all_correct.load(Ordering::Acquire))
```
enitrat, line 72: is this the right way of using `all_correct`?

scarb/src/bin/scarb/args.rs
```diff
@@ -341,6 +341,9 @@ pub struct FmtArgs {
     /// Specify package(s) to format.
     #[command(flatten)]
     pub packages_filter: PackagesFilter,
+    /// Path to a file or directory to format. If provided, only this file or directory will be formatted.
+    #[clap(value_name = "PATH")]
+    pub target_path: Option<Utf8PathBuf>,
```
mkaput, line 346: I would seriously consider avoiding the word `target`. I think `--path` would be just fine here, wdyt?

scarb/src/ops/fmt.rs
```diff
@@ -42,28 +45,97 @@ pub fn format(opts: FmtOptions, ws: &Workspace<'_>) -> Result<bool> {
...
+    target_file: &Utf8PathBuf,
+    opts: &FmtOptions,
+    ws: &Workspace<'_>,
+    all_correct: &AtomicBool,
+) -> Result<bool> {
+    let target_file = target_file.canonicalize_utf8()?;
+    let pkg = ws
+        .members()
+        .find(|member| target_file.starts_with(member.root()))
+        .ok_or_else(|| anyhow::anyhow!("File {:?} is not part of the workspace.", target_file))?;
```
maciektr, line 115:
```suggestion
        .ok_or_else(|| anyhow::anyhow!("file {:?} is not part of the workspace", target_file))?;
```
nit:

scarb/src/ops/fmt.rs
maciektr: Seems alright to me :)

scarb/src/ops/fmt.rs
```diff
@@ -42,28 +45,97 @@ pub fn format(opts: FmtOptions, ws: &Workspace<'_>) -> Result<bool> {
...
+        let pkg = ws
+            .members()
+            .find(|member| target_path.starts_with(member.root()));
+
+        if let Some(pkg) = pkg {
+            format_package(&pkg, &opts, ws, &all_correct)?;
+            return Ok(all_correct.load(Ordering::Acquire));
+        } else {
+            return Err(anyhow::anyhow!(
+                "File {:?} is not part of the workspace.",
```
maciektr, line 62:
```suggestion
                "path {:?} is not part of the workspace",
```

maciektr: There is a shorthand  for`return Err(anyhow!` called `anyhow::bail` :)

scarb/src/ops/fmt.rs
maciektr: FYI not emitting the path if formatting a single file only is okay to me :) Leaving the decision to you.

scarb/src/ops/fmt.rs
enitrat: done

scarb/src/ops/fmt.rs
enitrat: done

scarb/src/ops/fmt.rs
enitrat: ok, let's go with 2 then. It's easier to handle

scarb/src/bin/scarb/args.rs
mkaput: exactly, thx

maciektr (approved):
Looks good to me! 😄 
Thanks so much for your contribution! ❤️

PR #1630, opened 2024-10-03
Update scarb docs

### LINKED ISSUE #1512: Update dependencies guide to use registries
Update dependencies guide to refer to registries as the default way of adding dependencies to the registry.

https://github.com/software-mansion/scarb/blob/main/website/docs/guides/dependencies.md

Let's consider making this a separate guide, to preserve current guide detailing all nuances of Git-based dependency management  as is.

maciektr:
fixes  #1512

### DIFF
--- a/website/docs/guides/dependencies.md
+++ b/website/docs/guides/dependencies.md
@@ -6,37 +6,37 @@ To add a dependency, simply declare it in your `Scarb.toml`.
 > [!WARNING]
 > Using Git repositories as a foundation for package management is not the recommended approach anymore.
 > Instead, [registries](../registries/overview.md) are now the primary way to manage dependencies.
-> This guide will be updated to reflect the new approach after official registry is out of beta and considered stable.
-> For details on how to specify dependencies from the official registry, see [here](../reference/specifying-dependencies#specifying-dependencies-from-official-registry).
+> For details on how to specify dependencies from the official registry,
+> see [here](../reference/specifying-dependencies#specifying-dependencies-from-official-registry).
+> If you want to rely on git dependencies instead of the recommended way, you can learn more
+> [here](../reference/specifying-dependencies#specifying-dependencies-from-git-repositories).
 
 ## Adding a dependency
 
-If your `Scarb.toml` doesn't already have a `[dependencies]` section, add it, then list the package name and the URL to
-its Git repository.
-This example adds a dependency on the [`alexandria`](https://github.com/keep-starknet-strange/alexandria) package (note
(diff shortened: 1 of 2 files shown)


website/docs/guides/dependencies.md
```diff
@@ -6,56 +6,62 @@ To add a dependency, simply declare it in your `Scarb.toml`.
...
 ```toml
 [dependencies]
-alexandria_math = { git = "https://github.com/keep-starknet-strange/alexandria.git" }
+openzeppelin_merkle_tree = "0.17.0"
 ```
 
-You can pin a Git dependency to concrete commit, branch or a tag using one of the following extra fields that can be
-passed along `git`: `branch`, `tag` and `rev`.
+Using `"0.17.0"` as version requirement means, that you want to use a version `0.17.0` or newer, but not `0.18.0` or
+newer. To accept only a specific version, you can use `"=0.17.0"`. You can learn more about specifying version
```
THenry14, line 30: nitpick: I think phrasing here could be better. Maybe something like "... that you want to use a version `0.17.0` or newer, up until `0.18.0`" or something similar?

website/docs/guides/dependencies.md
```diff
@@ -6,56 +6,62 @@ To add a dependency, simply declare it in your `Scarb.toml`.
...
-If your `Scarb.toml` doesn't already have a `[dependencies]` section, add it, then list the package name and the URL to
-its Git repository.
-This example adds a dependency on the [`alexandria`](https://github.com/keep-starknet-strange/alexandria) package (note
-that Alexandria is a collection of multiple packages, and we will use `alexandria_math` as an example in this guide):
+If your `Scarb.toml` doesn't already have a `[dependencies]` section, add it, then list the package name and the version
+required. You can search for packages to use through the [scarbs.xyz registry website](https://scarbs.xyz/).
+This example adds a dependency on the [`openzeppelin_merkle_tree`](https://github.com/OpenZeppelin/cairo-contracts)
+package (note that OpenZeppelin is a collection of multiple packages, and we will use only one of them as an example in
+this guide). To see all available versions of some package, you can see
+its [registry page](https://scarbs.xyz/packages/openzeppelin_merkle_tree). At the time of writing this guide, the latest
```
THenry14, line 21: Should we also mention to go to "versions" pane, or is it so obvious it's redundant to do so in your opinion?

THenry14 (approved):
both comments are rather nitpicks, so I approve :+1:

website/docs/guides/dependencies.md
maciektr: Yes, thanks!

PR #1634, opened 2024-10-07
Do not allow importing by test target name in integration tests

maciektr:


scarb/src/compiler/db.rs
```diff
@@ -169,6 +169,7 @@ fn build_project_config(unit: &CairoCompilationUnit) -> Result<ProjectConfig> {
                         .any(|target| {
                             target.group_id.clone().unwrap_or(target.name.clone())
                                 == component.package.id.name.to_smol_str()
+                            && component_as_dependency.cairo_package_name() != component.cairo_package_name()
```
DelevoXDG, line 172: Why this change?

maciektr: To make sure, we do not add dependency on self here.

scarb/src/compiler/db.rs
```diff
@@ -184,13 +185,21 @@ fn build_project_config(unit: &CairoCompilationUnit) -> Result<ProjectConfig> {
...
 
             // Adds itself to dependencies
-            dependencies.insert(
-                component.package.id.name.to_string(),
-                DependencySettings {
-                    version: (component.package.id.name.to_string() != *CORELIB_CRATE_NAME)
-                        .then_some(component.package.id.version.clone()),
-                },
-            );
+            if component.first_target().kind.is_test() {
```
DelevoXDG, line 188: Correct me if I'm wrong, but the title of this PR and previous state of things suggest we only want to filter out integration tests, not all non-tests?

maciektr: Good catch! Submitted a fix :D

DelevoXDG (approved):

piotmag769:
@enitrat a heads up cause I know you use this import method in integration tests for some projects: pls migrate to `use crate::`

enitrat:
thanks 🙏

Issue #1132, opened 2024-02-07
Implement full TokenStream internals 


maciektr:
### Problem

Cairo should be represented as tokenized into TokenTrees

PR #1639, opened 2024-10-10
TokenStream implementation for host-macro communication

wawel37:
Related to #1132 .

Wait for [this](https://github.com/starkware-libs/cairo/pull/6475) to be merged on Cairo side


plugins/cairo-lang-macro/src/types/token.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub fn new(tokens: Vec<Token>) -> Self {
+        Self {
+            tokens,
+            metadata: TokenStreamMetadata::default(),
+        }
+    }
+
+    pub fn from_string(str: String) -> Self {
+        let db = SimpleParserDatabase::default();
+        // Sometimes this will return an error, when the user pass the simple literal value like "34".
```
Draggu, line 59: This will fail if there is passed expression instead of `ModuleItem`, when parsing virtual file it's kind have to be correctly set

plugins/cairo-lang-macro/src/types/token.rs
```diff
@@ -0,0 +1,130 @@
...
+        // In this case, we create the Token manually with its span.
```
Draggu, line 60: This can be arbitrary complex expression, not just single token

mkaput:
how does this relate to https://github.com/starkware-libs/cairo/pull/6388?

wawel37:
> how does this relate to [starkware-libs/cairo#6388](https://github.com/starkware-libs/cairo/pull/6388)?

I don't think the [#6388](https://github.com/starkware-libs/cairo/pull/6388) directly relates to this PR. The changes there only change the Syntax Tree for macros, so it won't affect this certain PR. Yes, it might change something when it comes to Scarb macro host, but i'm not so sure about that. Even if it does, it's better to adjust to cairo changes in a separate change, as this is not related

plugins/cairo-lang-macro/Cargo.toml
maciektr, line 1: Generally speaking, it's crucial to have as little dependencies here as possible, as this structs are linked into each and every procedural macro. This package also need to be as stable as possible, as often cairo-lang-macro on Scarb side will have different version, than cairo-lang-macro on proc macro side.

plugins/cairo-lang-macro/src/types/token.rs
```diff
@@ -0,0 +1,149 @@
...
+    span: TextSpan,
+    content: String,
+}
+
+impl Token {
+    pub fn new(content: String, span: TextSpan) -> Self {
+        Self { content, span }
+    }
+
+    pub fn from_syntax_node(db: &dyn SyntaxGroup, node: SyntaxNode) -> Self {
```
maciektr, line 27: This package should not depend on Cairo compiler at all. Instead of accepting SyntaxNode and building stream for the user, it should simply define the structs that user needs to build himself.

plugins/cairo-lang-macro/src/types/token.rs
```diff
@@ -0,0 +1,149 @@
...
+        Token::new(node.get_text(db), node.span(db))
+    }
+}
+
+/// An abstract stream of Cairo tokens.
+///
+/// This is both input and part of an output of a procedural macro.
+#[derive(Debug, Default, Clone)]
+pub struct TokenStream {
+    pub tokens: Vec<Token>,
```
maciektr, line 37: Shouldn't we go for `struct TokenStream -> Vec<enum TokenTree> -> TokenTree::Ident()` already? To be more future safe? 🤔

plugins/cairo-lang-macro/src/types/token.rs
```diff
@@ -0,0 +1,149 @@
...
+    #[doc(hidden)]
+    pub fn new(tokens: Vec<Token>) -> Self {
+        Self {
+            tokens,
+            metadata: TokenStreamMetadata::default(),
+        }
+    }
+
+    pub fn from_string(str: String) -> Self {
+        let db = SimpleParserDatabase::default();
```
maciektr, line 64: As stated earlier, I don't think this should happen in this package.

maciektr (changes requested):

maciektr:
Replaced with #1699

PR #1646, opened 2024-10-14
Enable features for main package during integration tests compilation

### LINKED ISSUE #1637: Enable features for main package during integration tests compilation
### Problem

Modify https://github.com/software-mansion/scarb/blob/eec9b4af3bdbc2ec9c391201fd530ba2b39e98ba/scarb/src/ops/resolve.rs#L284 so that `features` passed to scarb with `--features` flag are enabled for components representing both `src/` and `tests/` directories, when compiling integration tests.  

### Proposed Solution

_No response_

### Notes

_No response_

DelevoXDG:
Closes #1637

### DIFF
--- a/scarb/src/ops/resolve.rs
+++ b/scarb/src/ops/resolve.rs
@@ -377,11 +377,9 @@ fn cairo_compilation_unit_for_target(
         components.push(CompilationUnitComponent::try_new(
             member.clone(),
             vec![target],
-            get_cfg_with_features(
-                no_test_cfg_set.unwrap(),
-                &member.manifest.features,
-                enabled_features,
-            )?,
+            no_test_cfg_set.and_then(|cfg_set| {
+                get_cfg_with_features(cfg_set, &member.manifest.features, enabled_features).unwrap()
+            }),
         )?);
 
         // Set test package as main package for this compilation unit.
--- a/scarb/src/ops/resolve.rs
+++ b/scarb/src/ops/resolve.rs
@@ -377,7 +377,11 @@ fn cairo_compilation_unit_for_target(
         components.push(CompilationUnitComponent::try_new(
             member.clone(),
             vec![target],
-            no_test_cfg_set,
+            get_cfg_with_features(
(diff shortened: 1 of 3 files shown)


scarb/src/ops/resolve.rs
```diff
@@ -377,7 +377,9 @@ fn cairo_compilation_unit_for_target(
         components.push(CompilationUnitComponent::try_new(
             member.clone(),
             vec![target],
-            no_test_cfg_set,
+            no_test_cfg_set.and_then(|cfg_set| {
```
maciektr, line 380: This should include features even for an empty set, right? 🤔

scarb/src/ops/resolve.rs
```diff
@@ -377,7 +377,12 @@ fn cairo_compilation_unit_for_target(
         components.push(CompilationUnitComponent::try_new(
             member.clone(),
             vec![target],
-            no_test_cfg_set,
+            get_cfg_with_features(
+                no_test_cfg_set.unwrap_or(cfg_set.clone()),
```
maciektr, line 381: It cannot use `cfg_set` as default - it can only be used in the main compilation unit. Please use an empty set instead.

DelevoXDG: Sure 🫡  [b26e4b4](https://github.com/software-mansion/scarb/pull/1646/commits/b26e4b40f323e5680a933eb52051d6c990964164)

scarb/src/ops/resolve.rs
```diff
@@ -377,7 +377,12 @@ fn cairo_compilation_unit_for_target(
         components.push(CompilationUnitComponent::try_new(
             member.clone(),
             vec![target],
-            no_test_cfg_set,
+            get_cfg_with_features(
+                no_test_cfg_set.unwrap_or(CfgSet::new()),
+                &member.manifest.features,
+                enabled_features,
+            )
+            .unwrap(),
```
maciektr, line 385: Why unwrap instead of bubble the Result?

maciektr (approved):

PR #1648, opened 2024-10-15
Add CI to publish builtin plugins

### LINKED ISSUE #1619: Add CI to upload builtin Cairo plugins
Add CI to publish builtin Cairo plugins upon release

DelevoXDG:
Test run: https://github.com/software-mansion/scarb/actions/runs/11365042651/job/31612347755?pr=1649

Closes #1619


.github/workflows/publish_plugins.yml
```diff
@@ -0,0 +1,63 @@
...
+  release:
+    types:
+      - published
+
+jobs:
+  publish:
+    name: Publish plugins
+    runs-on: ubuntu-latest
+    env:
+      SCARB_REGISTRY_AUTH_TOKEN: ${{ secrets.SCARB_REGISTRY_AUTH_TOKEN }}
```
DelevoXDG, line 12: What token do you think we should keep in our GitHub secrets? Do we want to use some sort of an organization account, or will my personal token do for now? :D

Also, do you think we need any expiration date for it?

.github/workflows/publish_plugins.yml
```diff
@@ -0,0 +1,63 @@
...
+        run: |
+          set -eo pipefail
+          mkdir -p $HOME/pkg
+          pushd $HOME/pkg
+          scarb init --no-vcs
+          echo "CAIRO_VERSION=$(scarb metadata --format-version 1 | jq -r '.app_version_info.cairo.version')" >> $GITHUB_OUTPUT
+        env:
+          SCARB_INIT_TEST_RUNNER: cairo-test
+
+      - name: Check if plugin versions exist in the registry
```
maciektr, line 32: Why do we need to check it here? Can't we just try to publish and let the pipeline fail if it has already been uploaded? (Generally it expect it not to be, right?)

.github/workflows/publish_plugins.yml
```diff
@@ -0,0 +1,63 @@
...
+          echo "assert_macros_exists=$(curl -s https://scarbs.xyz/api/v1/index/as/se/assert_macros.json | jq --arg version "${{ steps.scarb-init.outputs.CAIRO_VERSION }}" '[.[] | select(.v == $version)] | length > 0')" >> $GITHUB_OUTPUT
+          echo "cairo_run_exists=$(curl -s https://scarbs.xyz/api/v1/index/ca/ir/cairo_run.json | jq --arg version "${{ steps.scarb-init.outputs.CAIRO_VERSION }}" '[.[] | select(.v == $version)] | length > 0')" >> $GITHUB_OUTPUT
+          echo "starknet_exists=$(curl -s https://scarbs.xyz/api/v1/index/st/ar/starknet.json | jq --arg version "${{ steps.scarb-init.outputs.CAIRO_VERSION }}" '[.[] | select(.v == $version)] | length > 0')" >> $GITHUB_OUTPUT
+          echo "cairo_test_exists=$(curl -s https://scarbs.xyz/api/v1/index/ca/ir/cairo_test.json | jq --arg version "${{ steps.scarb-init.outputs.CAIRO_VERSION }}" '[.[] | select(.v == $version)] | length > 0')" >> $GITHUB_OUTPUT
+
+      - name: Publish starknet
+        if: steps.check-plugins.outputs.starknet_exists != 'true'
+        run: |
+          pushd $HOME/.cache/scarb/registry/std/v${{ steps.scarb-init.outputs.CAIRO_VERSION }}/starknet
+          scarb publish --no-verify
```
maciektr, line 45: Why no verify?

.github/workflows/publish_plugins.yml
```diff
@@ -0,0 +1,63 @@
...
+          set -eo pipefail
+          echo "assert_macros_exists=$(curl -s https://scarbs.xyz/api/v1/index/as/se/assert_macros.json | jq --arg version "${{ steps.scarb-init.outputs.CAIRO_VERSION }}" '[.[] | select(.v == $version)] | length > 0')" >> $GITHUB_OUTPUT
+          echo "cairo_run_exists=$(curl -s https://scarbs.xyz/api/v1/index/ca/ir/cairo_run.json | jq --arg version "${{ steps.scarb-init.outputs.CAIRO_VERSION }}" '[.[] | select(.v == $version)] | length > 0')" >> $GITHUB_OUTPUT
+          echo "starknet_exists=$(curl -s https://scarbs.xyz/api/v1/index/st/ar/starknet.json | jq --arg version "${{ steps.scarb-init.outputs.CAIRO_VERSION }}" '[.[] | select(.v == $version)] | length > 0')" >> $GITHUB_OUTPUT
+          echo "cairo_test_exists=$(curl -s https://scarbs.xyz/api/v1/index/ca/ir/cairo_test.json | jq --arg version "${{ steps.scarb-init.outputs.CAIRO_VERSION }}" '[.[] | select(.v == $version)] | length > 0')" >> $GITHUB_OUTPUT
+
+      - name: Publish starknet
+        if: steps.check-plugins.outputs.starknet_exists != 'true'
+        run: |
+          pushd $HOME/.cache/scarb/registry/std/v${{ steps.scarb-init.outputs.CAIRO_VERSION }}/starknet
```
maciektr, line 44: This should read path from Scarb, instead of hardcoding it here.

.github/workflows/publish_plugins.yml
```diff
@@ -0,0 +1,63 @@
...
+    runs-on: ubuntu-latest
+    env:
+      SCARB_REGISTRY_AUTH_TOKEN: ${{ secrets.SCARB_REGISTRY_AUTH_TOKEN }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: software-mansion/setup-scarb@v1
+        with:
+          scarb-version: ${{ github.ref_name }}
```
maciektr, line 19: Won't this be `refs/tags/<tag_name>`? https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

.github/workflows/publish_plugins.yml
DelevoXDG: Right, I forgot we don't verify builins anyway :D 
Removed

.github/workflows/publish_plugins.yml
DelevoXDG: Afaik, `ref_name` is a short name and is not the same as `GITHUB_REF`. 

`ref_name` should be a name of the tag that triggered the workflow: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

.github/workflows/publish_plugins.yml
DelevoXDG: > (Generally it expect it not to be, right?)

Since some Cairo versions correspond to multiple Scarb versions, this is likely to happen from time to time. 

> Can't we just try to publish and let the pipeline fail if it has already been uploaded?

I think this is the question of deciding how do we want the action to resolve in such instances. We can `continue-on-error`, and let entire workflow fail, or we can skip the uploads. It seems to me that if action is resolved within *what is expected* should not be deemed "failed".

maciektr: Makes sense, thanks!

maciektr (approved):
🦾

PR #1650, opened 2024-10-16
Add add_redeposit_gas to compiler config

maciektr:


website/docs/reference/manifest.md
```diff
@@ -302,6 +302,20 @@ If `avoid` strategy is set, the compiler will only inline function annotated wit
...
 > You can use profile settings overwriting, for more granular control of which builds use the avoid strategy.
 
+### `add-redeposit-gas`
+
+> [!WARNING]
+> This is experimental and unstable flag intended to be used for Cairo compiler research.
+> It is advised not to use it for other purposes than that.
+
+If enabled, during the project compilation Scarb will add redeposit gas libfunc calls in some locations.
+By default, this flag is disabled.
```
DelevoXDG, line 312: Nit: 

All contracts should be built with this enabled starting starknet 0.13.3, perhaps it'd be reasonable to mention that?

DelevoXDG (approved):
I think we can add "Add add_redeposit_gas to compiler config" to merge queue

website/docs/reference/manifest.md
```diff
@@ -302,6 +302,20 @@ If `avoid` strategy is set, the compiler will only inline function annotated wit
...
+
+```toml
+[cairo]
+add-redeposit-gas = false
+```
```
wawel37, line 317: Nit: If default is `false`, isn't it better to make it `true` in the example, as making it `false` won't have any effect?

wawel37 (approved):
Love when PRs got so clean tests

website/docs/reference/manifest.md
maciektr: Nope, it will require action from us to change defaults and docs anyway.

website/docs/reference/manifest.md
maciektr: Done

PR #1652, opened 2024-10-16
Stop using process:exit for error code handling

maciektr:


scarb/src/bin/scarb/main.rs
```diff
@@ -39,25 +40,27 @@ fn main() {
...
 
-fn exit_with_error(err: Error, ui: &Ui) {
+fn exit_with_error(err: Error, ui: &Ui) -> ExitCode {
     debug!("exit_with_error; err={:?}", err);
 
     if let Some(ErrorWithExitCode { source, exit_code }) = err.downcast_ref::<ErrorWithExitCode>() {
         if let Some(source_err) = source {
             ui.anyhow(source_err);
         }
-        std::process::exit(*exit_code);
```
maciektr, line 53: >Note that because this function never returns, and that it terminates the process, no destructors on the current stack or any other thread’s stack will be run. If a clean shutdown is needed it is recommended to only call this function at a known point where there are no more destructors left to run; or, preferably, simply return a type implementing [Termination](https://doc.rust-lang.org/stable/std/process/trait.Termination.html) (such as [ExitCode](https://doc.rust-lang.org/stable/std/process/struct.ExitCode.html) or Result) from the main function and avoid this function altogether:

https://doc.rust-lang.org/stable/std/process/fn.exit.html#

piotmag769 (approved):

PR #1655, opened 2024-10-17
Add component dependencies to metadata

piotmag769:


scarb/src/compiler/compilation_unit.rs
```diff
@@ -92,6 +96,20 @@ pub struct CompilationUnitCairoPlugin {
...
+/// It may be not sufficient in the future depending on changes to the compilation model.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct CompilationUnitComponentId {
+    pub package_id: PackageId,
+}
+
+impl CompilationUnitComponentId {
+    pub fn to_discriminator(&self) -> String {
+        self.package_id.to_serialized_string()
+    }
```
piotmag769, line 110: Changed discriminator to be serialized `PackageId` instead of `Version` to be aligned with the comment to `CompilationUnitComponentDependencyMetadata::id`

piotmag769: This requires changes to Cairo to work ^

scarb/src/ops/resolve.rs
```diff
@@ -390,6 +391,61 @@ fn cairo_compilation_unit_for_target(
         member.id
     };
 
+    // Collect dependencies for the components.
```
maciektr, line 394: This logic is quite indirect in this context, because you have access to `WorkspaceResolve`, which keeps a graph of all packages in CU.

scarb/src/ops/metadata.rs
```diff
@@ -386,6 +399,14 @@ fn wrap_source_id(id: SourceId) -> m::SourceId {
     }
 }
 
+fn wrap_compilation_unit_component_id(
```
maciektr, line 402: Why not just `id.to_discriminator().into()`?

scarb/src/compiler/db.rs
```diff
@@ -145,66 +145,25 @@ fn build_project_config(unit: &CairoCompilationUnit) -> Result<ProjectConfig> {
...
+                    let compilation_unit_component = unit.components.iter().find(|component| component.id == *compilation_unit_component_id)
+                        .expect("Dependency of a component is guaranteed to exist in compilation unit components");
                     (
                         compilation_unit_component.package.id.name.to_string(),
                         DependencySettings {
                             discriminator: (compilation_unit_component.package.id.name.to_string()
                                 != *CORELIB_CRATE_NAME)
-                                .then_some(compilation_unit_component.package.id.version.clone())
-                                .map(|v| v.to_smolstr()),
+                                .then_some(compilation_unit_component.id.clone())
```
maciektr, line 160: I wonder if we should hide the corelib edge case in cu too.

scarb/src/compiler/db.rs
```diff
@@ -145,66 +145,25 @@ fn build_project_config(unit: &CairoCompilationUnit) -> Result<ProjectConfig> {
...
-                        .any(|target| {
-                            target.group_id.clone().unwrap_or(target.name.clone())
-                                == component.package.id.name.to_smol_str()
-                            && component_as_dependency.cairo_package_name() != component.cairo_package_name()
-                        })
-                })
-                .map(|compilation_unit_component| {
+                .map(|compilation_unit_component_id| {
+                    let compilation_unit_component = unit.components.iter().find(|component| component.id == *compilation_unit_component_id)
+                        .expect("Dependency of a component is guaranteed to exist in compilation unit components");
```
maciektr, line 154:
```suggestion
                        .expect("dependency of a component is guaranteed to exist in compilation unit components");
```

scarb-metadata/src/lib.rs
```diff
@@ -356,6 +380,28 @@ pub struct CompilationUnitComponentMetadata {
     /// If not specified, the one from `CompilationUnit` will be used.
     #[serde(default)]
     pub cfg: Option<Vec<Cfg>>,
+    /// Identifier of this component. It is unique in its compilation unit.
+    pub id: Option<CompilationUnitComponentId>,
+    /// Dependencies of this component.
+    pub dependencies: Option<Vec<CompilationUnitComponentDependencyMetadata>>,
```
maciektr, line 386: Do we need to differentiate between `Option::None` and empty `Vec`?

mkaput: Yes, because crate with no dependencies (empty vec) is a valid thing. And scarb metadata from old days (None) is also a valid thing and in such cache all components should be assumed as dependencies

scarb/src/compiler/db.rs
mkaput: Definitely. I don't think we want all tools to think about this case

scarb-metadata/src/lib.rs
```diff
@@ -106,6 +106,30 @@ impl fmt::Display for CompilationUnitId {
     }
 }
 
+/// An "opaque" identifier for a compilation unit component.
+/// It is possible to inspect the `repr` field if the need arises,
+/// but its precise format is an implementation detail and is subject to change.
+///
+/// [`Metadata`] can be indexed by [`CompilationUnitComponentId`].
```
mkaput, line 113: you haven't added that indexing operation

mkaput (changes requested):

scarb/src/compiler/compilation_unit.rs
piotmag769: I mean `ProjectConfig` is a terrible structure that sometimes has duplicated information, but sometimes lacks it (like in this case), so rn in `RootDatabaseBuilder` we set the discriminator to always be equal to `CrateSetting.version`

https://github.com/starkware-libs/cairo/blob/932767340c4b8a762140bf9eba305f437587ac1b/crates/cairo-lang-compiler/src/project.rs#L148-L152

scarb/src/compiler/db.rs
piotmag769: To be fair it doesn't even matter from the compiler's point of view if the corelib discriminator is `None` or not as long as `core` is a dependency for each package (which it is in Scarb)

https://github.com/starkware-libs/cairo/blob/932767340c4b8a762140bf9eba305f437587ac1b/crates/cairo-lang-semantic/src/resolve/mod.rs#L1077

I agree though that we should keep it as `None` for consistency, will change it in metadata and Scarb's internal model

scarb/src/ops/metadata.rs
piotmag769: It was just for consistency with how it's done for `PackageId` and `SourceId`. Can change it

scarb/src/compiler/db.rs
maciektr: Are you sure it does not matter? I know this has not been working without this `if` when we first introduced it. If we can omit this logic, it would be best to do so I think 🤔

scarb-metadata/src/lib.rs
piotmag769: Added indexing to `CompilationUnitMetadata`

mkaput (approved):
Looks good from `scarb-metadata` perspective. I haven't reviewed Scarb internals changes.

scarb/src/ops/resolve.rs
maciektr: Discussed offline - will be changed later by Scarb team

Cargo.toml
```diff
@@ -35,30 +35,30 @@ anyhow = "1"
...
-cairo-lang-sierra-to-casm = { git = "https://github.com/starkware-libs/cairo", rev = "7cfecf38801416c19a431e3a5c21b7d68615ce93" }
-cairo-lang-sierra-type-size = { git = "https://github.com/starkware-libs/cairo", rev = "7cfecf38801416c19a431e3a5c21b7d68615ce93" }
-cairo-lang-starknet = { git = "https://github.com/starkware-libs/cairo", rev = "7cfecf38801416c19a431e3a5c21b7d68615ce93" }
-cairo-lang-starknet-classes = { git = "https://github.com/starkware-libs/cairo", rev = "7cfecf38801416c19a431e3a5c21b7d68615ce93" }
-cairo-lang-syntax = { git = "https://github.com/starkware-libs/cairo", rev = "7cfecf38801416c19a431e3a5c21b7d68615ce93" }
-cairo-lang-test-plugin = { git = "https://github.com/starkware-libs/cairo", rev = "7cfecf38801416c19a431e3a5c21b7d68615ce93" }
-cairo-lang-test-runner = { git = "https://github.com/starkware-libs/cairo", rev = "7cfecf38801416c19a431e3a5c21b7d68615ce93" }
-cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo", rev = "7cfecf38801416c19a431e3a5c21b7d68615ce93", features = ["env_logger"] }
+cairo-lang-casm = { git = "https://github.com/starkware-libs/cairo", rev = "80df2b83f55537d1dd9dee68c82d7745b11612fa" }
```
piotmag769, line 38: Will be changed when I merge it

scarb/src/ops/compile.rs
```diff
@@ -4,12 +4,11 @@ use cairo_lang_compiler::diagnostics::DiagnosticsError;
...
 use indoc::formatdoc;
 use itertools::Itertools;
-use smol_str::{SmolStr, ToSmolStr};
-use std::thread;
-
 use scarb_ui::args::FeaturesSpec;
 use scarb_ui::components::Status;
 use scarb_ui::HumanDuration;
+use smol_str::{SmolStr, ToSmolStr};
+use std::thread;
```
mkaput, line 11: irrelevant changes

mkaput (approved):

scarb/src/ops/compile.rs
piotmag769: Probably an autoformatter, will leave it as it is

Cargo.toml
```diff
@@ -34,30 +34,30 @@ anyhow = "1"
...
-cairo-lang-sierra-to-casm = { git = "https://github.com/starkware-libs/cairo", rev = "68938b757d8b3960f7588f27931ce716c5e914bc" }
-cairo-lang-sierra-type-size = { git = "https://github.com/starkware-libs/cairo", rev = "68938b757d8b3960f7588f27931ce716c5e914bc" }
-cairo-lang-starknet = { git = "https://github.com/starkware-libs/cairo", rev = "68938b757d8b3960f7588f27931ce716c5e914bc" }
-cairo-lang-starknet-classes = { git = "https://github.com/starkware-libs/cairo", rev = "68938b757d8b3960f7588f27931ce716c5e914bc" }
-cairo-lang-syntax = { git = "https://github.com/starkware-libs/cairo", rev = "68938b757d8b3960f7588f27931ce716c5e914bc" }
-cairo-lang-test-plugin = { git = "https://github.com/starkware-libs/cairo", rev = "68938b757d8b3960f7588f27931ce716c5e914bc" }
-cairo-lang-test-runner = { git = "https://github.com/starkware-libs/cairo", rev = "68938b757d8b3960f7588f27931ce716c5e914bc" }
-cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo", rev = "68938b757d8b3960f7588f27931ce716c5e914bc", features = ["env_logger"] }
+cairo-lang-casm = { git = "https://github.com/starkware-libs/cairo", rev = "71eaf8a2a02a9e50747261dd9c862862a9631374" }
```
maciektr, line 37: This needs to be merged on main first before merging this

extensions/scarb-doc/src/metadata/compilation.rs
```diff
@@ -152,43 +129,25 @@ fn get_crate_settings_for_package(
...
-                        .then_some(version)
-                        .flatten()
-                        .map(|v| v.to_smolstr());
-                    (
-                        dependency.name.clone(),
-                        DependencySettings { discriminator },
-                    )
-                })
+        .map(|CompilationUnitComponentDependencyMetadata {id, .. } | {
+            let dependency_component = unit.components.iter().find(|component| component.id.as_ref().unwrap() == id)
```
wawel37, line 138: nit: expect here plz instead of unwrap

extensions/scarb-doc/src/metadata/compilation.rs
```diff
@@ -64,26 +57,15 @@ fn package_compilation_unit(
...
-) -> OrderedHashMap<SmolStr, PathBuf> {
+) -> OrderedHashMap<CrateIdentifier, PathBuf> {
     compilation_unit_metadata
         .components
         .iter()
-        .filter(|du| du.name != CORELIB_CRATE_NAME)
         .map(|cu| {
             (
-                cu.name.to_smolstr(),
+                cu.id.as_ref().unwrap().into(),
```
wawel37, line 68: nit: expect rather than unwrap 🙏

extensions/scarb-doc/src/metadata/compilation.rs
```diff
@@ -152,43 +129,25 @@ fn get_crate_settings_for_package(
             .contains(&String::from("coupons")),
     };
 
-    let mut dependencies: BTreeMap<String, DependencySettings> = package
+    let dependencies = component
         .dependencies
+        .as_ref()
+        .unwrap()
```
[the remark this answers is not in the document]
piotmag769: It cannot if it is used with new Scarb (which is the only reasonable way to use `scarb-doc` binary imo)

maciektr (approved):

PR #1657, opened 2024-10-18
Remove code mappings for attribute expand (temp)

maciektr:


piotmag769 (approved):

wawel37 (commented):
Just to be sure, we only want to disable it for attributes, right? Because i see that we are still returning mappings for [Derives](https://github.com/software-mansion/scarb/blob/6deb5020e3686f67332848ef696e5b86c6242623/scarb/src/compiler/plugin/proc_macro/host.rs#L574).

wawel37 (approved):

PR #1666, opened 2024-10-22
Add `--ignore-cairo-version` flag

### LINKED ISSUE #1661: Add `—ignore-cairo-version` flag which allows build even if `cairo-version` does not match

DelevoXDG:
Closes #1661


scarb/src/ops/resolve.rs
```diff
@@ -592,12 +603,17 @@ fn check_cairo_version_compatibility(packages: &[Package], ws: &Workspace<'_>) -
                     pkg.id.name,
                     package_version,
                     current_version
-                ));
+                );
+                if ignore_mismatch {
+                    ws.config().ui().warn(msg);
+                } else {
+                    ws.config().ui().error(msg);
```
DelevoXDG, line 610: Should we add a help message suggesting to use newly added `--ignore-cairo-version` flag? 🤔

DelevoXDG: As for the error messages, they do in fact suggest setting a flag. However, no warnings are displayed on mismatch 🤔

maciektr: So we can suggest setting it too :)

DelevoXDG: Yea, I added that 😄 

My question was rather about the warnings: should we keep them or do as Cargo does? :D

maciektr: Keep please

scarb/tests/build.rs
```diff
@@ -293,6 +293,42 @@ fn compile_with_incompatible_cairo_version() {
...
+        .assert()
+        .success()
+        .stdout_matches(indoc! {r#"
+            warn: the required Cairo version of package hello is not compatible with current version
+            Cairo version required: ^33.33.0
+            Cairo version of Scarb: [..]
+
+            warn: the required Cairo version of package hello is not compatible with current version
+            Cairo version required: ^33.33.0
+            Cairo version of Scarb: [..]
```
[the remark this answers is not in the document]
DelevoXDG: We previously discussed offline the idea of storing the fact that the warning occurred to avoid duplication. 

However, on the second look, it seems to me a cleaner solution would be to just move the check to `generate_compilation_units`. 

What do you think?

scarb/src/ops/resolve.rs
```diff
@@ -178,6 +179,8 @@ pub fn generate_compilation_units(
         .filter(|member| !member.is_cairo_plugin())
         .collect_vec();
     validate_features(&members, enabled_features)?;
+    check_cairo_version_compatibility(&members, ws, ignore_cairo_version)?;
```
maciektr, line 182: This cannot be run on members only. Members are all package that are part of the workspace - but **without their dependencies**. This should be run on all dependencies too.

maciektr (changes requested):

scarb/src/ops/resolve.rs
DelevoXDG: Good catch. Used a different approach with `PackageSolutionCollector.warnings`, let me know what you think

PR #1668, opened 2024-10-22
Support UI verbosity in scarb-doc

maciektr:
commit-id:cbcee713


extensions/scarb-doc/src/lib.rs
```diff
@@ -70,7 +71,7 @@ pub fn generate_packages_information(
             .find(|unit| unit.package == package_metadata.id);
 
         let mut diagnostics_reporter =
-            setup_diagnostics_reporter(&db, main_crate_id, package_compilation_unit)
+            setup_diagnostics_reporter(&db, main_crate_id, package_compilation_unit, ui.clone())
```
wawel37, line 74: Nit: Better to just pass the reference to `setup_diagnostics_reporter`.

wawel37 (approved):

PR #1677, opened 2024-10-24
`cairo-lang-macro` fix null-byte panics

Draggu:


plugins/cairo-lang-macro/src/types/conversion.rs
```diff
@@ -95,7 +95,9 @@ impl TokenStream {
     ///
     /// # Safety
     #[doc(hidden)]
-    pub fn into_stable(self) -> StableTokenStream {
+    pub fn into_stable(mut self) -> StableTokenStream {
+        // Remove 0 bytes so conversion will always work.
```
mkaput, line 99: have you managed to find out why these null bytes where there?

Draggu: Sadly no, this is workaround for it.

mkaput: Have you tried logging all TokenStream allocations (even plain dbg! is fine) to a file and mining that data to find a culprit? If not, can you please send me such dump to analyze myself please? :)

PR #1680, opened 2024-10-24
Proc macro server: Load `ProcMacroHost`

Draggu:
**Stack**:
- #1686
- #1685
- #1684
- #1683
- #1718
- #1682
- #1681
- #1680 ⬅


scarb/src/bin/scarb/commands/proc_macro_server.rs
```diff
@@ -1,7 +1,52 @@
...
+        &FeaturesOpts {
+            features: FeaturesSelector::AllFeatures,
+            no_default_features: false,
+        },
+        &ws,
+    )?;
+
+    for unit in &compilation_units {
+        if let CompilationUnit::ProcMacro(_) = unit {
+            ops::compile_unit(unit.clone(), &ws)?;
```
mkaput, line 23: why are you compiling everything here?

mkaput (changes requested):

scarb/src/bin/scarb/commands/proc_macro_server.rs
```diff
@@ -1,7 +1,52 @@
...
+    compiler::{plugin::proc_macro::ProcMacroHost, CairoCompilationUnit, CompilationUnit},
+    core::{Config, Workspace},
+    ops::{self, FeaturesOpts, FeaturesSelector},
+};
 
 #[tracing::instrument(skip_all, level = "info")]
-pub fn run(_config: &Config) -> Result<()> {
-    unimplemented!()
+pub fn run(config: &mut Config) -> Result<()> {
+    let ws = ops::read_workspace(config.manifest_path(), config)?;
```
Arcticae, line 10: What if this changes in time (i.e. more crates are added)

mkaput: We will detect this on a client (LS) and just restart the server

scarb/src/bin/scarb/commands/proc_macro_server.rs
Draggu: Not sure what could be omitted. If you have better inside can you type here what could be removed?

mkaput: Compiling creates the `RootDatabase`, runs full compilation and outputs artefacts. That's nothing we can make use of, so we are safe to skip this.

The rest of the code here all looks okay for me here. We basically need to `scarb fetch` before booting up, so you just need to:
1. Read the workspace
2. Resolve it, including downloading all dependencies.
3. Generate compilation units to gather all compiler db parameters.
4. Create proc macro host.

maciektr: @mkaput but those are only macros -> per match one line prior

mkaput: > @mkaput but those are only macros -> per match one line prior

I don't understand, could you elaborate please?

maciektr:
```
        if let CompilationUnit::ProcMacro(_) = unit {
            ops::compile_unit(unit.clone(), &ws)?;
```
Unit is only compiled, if it fulfills the `let CompilationUnit::ProcMacro(_) = unit` match, thus is guaranteed to represent a procedural macro.

mkaput: 🤦🏻 I didn't notice this. @Draggu Could you please add a comment elaborating that we are compiling procedural macros here?

mkaput (approved):

piotmag769 (approved):

PR #1681, opened 2024-10-24
Add Proc macro server

Draggu:
**Stack**:
- #1686
- #1685
- #1684
- #1683
- #1718
- #1682
- #1681 ⬅


scarb/src/ops/proc_macro_server/connection.rs
```diff
@@ -0,0 +1,98 @@
...
+                    eprintln!("Error occurred while deserializing request, used input:\n{line}");
+
+                    break;
+                };
+
+                if reader_sender.send(request).is_err() {
+                    eprintln!("Error occurred while sending request to worker threads");
+
+                    break;
+                }
```
mkaput, line 48: a lot of vertical noise here imo

```suggestion
                if stdin.read_line(&mut line).is_err() {
                    eprintln!("Error occurred while reading from stdin");
                    break;
                }

                if line.is_empty() {
                    continue;
                }

                let Ok(request) = serde_json::from_str(&line) else {
                    eprintln!("Error occurred while deserializing request, used input:\n{line}");
                    break;
                };

                if reader_sender.send(request).is_err() {
                    eprintln!("Error occurred while sending request to worker threads");
                    break;
                }
```

scarb/src/ops/proc_macro_server/connection.rs
```diff
@@ -0,0 +1,98 @@
...
+            let mut line = String::new();
+
+            loop {
+                line.clear();
+
+                if stdin.read_line(&mut line).is_err() {
+                    eprintln!("Error occurred while reading from stdin");
+
+                    break;
+                }
```
mkaput, line 32: 1. don't you want to build errors here & use anyhow contexts?
2. you're logging to stderr here, so use `tracing` macros

scarb/src/ops/proc_macro_server/connection.rs
```diff
@@ -0,0 +1,98 @@
+use crossbeam_channel::{Receiver, Sender};
+use proc_macro_server_api::{RpcRequest, RpcResponse};
+use std::{
+    io::{BufRead, Write},
+    thread::JoinHandle,
+};
```
mkaput, line 6: we don't use nested imports in our codebases

mkaput (changes requested):

scarb/src/ops/proc_macro_server/connection.rs
```diff
@@ -0,0 +1,101 @@
...
+pub struct Connection {
+    pub sender: Sender<RpcResponse>,
+    pub receiver: Receiver<RpcRequest>,
+    io_threads: IoThreads,
+}
+
+impl Connection {
+    pub fn new() -> Self {
+        let (reader_sender, reader_receiver) = crossbeam_channel::bounded(0);
+        let (writer_sender, writer_receiver) = crossbeam_channel::bounded(0);
```
mkaput, line 17: you can use `std::sync::mpsc` here as well, it's baked by crossbeam under the hood anyway, yet less deps to think about

mkaput (changes requested):

scarb/src/ops/proc_macro_server/connection.rs
Draggu: I think it is different implementation https://github.com/crossbeam-rs/crossbeam/blob/10f0296a274f5c4b904a9e6bf95b0ca7e37708cb/crossbeam-channel/README.md?plain=1#L16

mkaput: https://www.reddit.com/r/rust/comments/18z6bcz/comment/kgfkgne/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

Draggu: Good to know, but after looking at our case we need `std::sync::mpmc` (we have multiple consumers and single producer for channel that is reading input and sending it to worker threads) not `std::sync::mpsc` and `std::sync::mpmc` is nightly only now.

mkaput (approved):

piotmag769 (approved):

PR #1684, opened 2024-10-24
Add `ExpandAttribute` handler

Draggu:
**Stack**:
- #1686
- #1685
- #1684 ⬅

### DIFF
--- a/scarb/src/ops/proc_macro_server/methods/expand_attribute.rs
+++ b/scarb/src/ops/proc_macro_server/methods/expand_attribute.rs
@@ -1,0 +1,29 @@
+use std::sync::Arc;
+
+use anyhow::Result;
+use scarb_proc_macro_server_types::methods::{expand::ExpandAttribute, ProcMacroResult};
+
+use super::Handler;
+use crate::compiler::plugin::proc_macro::{ExpansionKind, ProcMacroHost};
+
+impl Handler for ExpandAttribute {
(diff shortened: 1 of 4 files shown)


scarb/src/ops/proc_macro_server/methods/expand_attribute.rs
mkaput, line 1: missing tests

mkaput (changes requested):

scarb/tests/proc_macro_server.rs
```diff
@@ -36,3 +39,37 @@ fn defined_macros() {
...
+    let response = proc_macro_server
+        .request_and_wait::<ExpandAttribute>(ExpandAttributeParams {
+            attr: "some".to_string(),
+            args: TokenStream::empty(),
+            item: item.clone(),
+        })
+        .ok();
+
+    assert_eq!(response.diagnostics, vec![]);
+    assert_eq!(response.token_stream, item);
```
piotmag769, line 74: Can we do anything in the macro so this test actually tests if the logic executes?

mkaput: I think asserting on the debug-print will be enough

mkaput (changes requested):
blocking due to Piotr's comment

scarb/tests/proc_macro_server.rs
Draggu: >I think asserting on the debug-print will be enough

Will not as these are not proc-macros

piotmag769 (approved):

PR #1688, opened 2024-10-25
Feat: Support --print-detailed-resources parameter in cairo-run

maciejka:
This PR introduces an optional CLI parameter --print-detailed-resources for scarb cairo-run extension that prints detailed used resources report:
image

### DIFF
--- a/extensions/scarb-cairo-run/src/main.rs
+++ b/extensions/scarb-cairo-run/src/main.rs
@@ -233,7 +233,7 @@ struct Summary {
     result: RunResultStarknet,
     print_full_memory: bool,
     gas_defined: bool,
-    detailed_resources: bool
+    detailed_resources: bool,
 }
 
 impl Message for Summary {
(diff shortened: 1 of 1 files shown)


extensions/scarb-cairo-run/src/main.rs
```diff
@@ -43,6 +43,10 @@ struct Args {
     #[arg(long, default_value_t = false)]
     print_full_memory: bool,
 
+    /// Print detailed resources.
+    #[arg(long, default_value_t = false)]
+    print_detailed_resources: bool,
```
maciektr, line 48:
```suggestion
    print_resource_usage: bool,
```
Just to use the same name, as in cairo-test

extensions/scarb-cairo-run/src/main.rs
```diff
@@ -150,6 +154,7 @@ fn main_inner(ui: &Ui, args: Args) -> Result<()> {
         result,
         print_full_memory: args.print_full_memory,
         gas_defined: available_gas.is_defined(),
+        detailed_resources: args.print_detailed_resources,
```
maciektr, line 157:
```suggestion
        detailed_resources: args.print_resource_usage,
```

maciektr (approved):
Thanks for your contribution! 💯

PR #1689, opened 2024-10-25
Allow hiding warnings in Scarb output

### LINKED ISSUE #1660: Add option to hide warnings but not errors in Scarb output

DelevoXDG:
Add `no-warning` verbosity option, available as `--no-warning` flag or env var setting

Closes #1660

### DIFF
--- a/utils/scarb-ui/src/args/verbosity.rs
+++ b/utils/scarb-ui/src/args/verbosity.rs
@@ -24,6 +24,15 @@ pub struct VerbositySpec {
     quiet: u8,
 
     #[arg(
+    long = "no-warnings",
+    action = clap::ArgAction::SetTrue,
+    global = true,
+    help = "Decrease logging verbosity, hiding warnings but still showing errors.",
+    conflicts_with_all = &["verbose", "quiet"]
+    )]
+    no_warnings: bool,
+
+    #[arg(
         long,
         global = true,
         help = "Set UI verbosity level by name.",
@@ -35,7 +44,8 @@ pub struct VerbositySpec {
 impl Verbosity {
     fn level_value(level: Self) -> i8 {
         match level {
-            Self::Quiet => -1,
+            Self::Quiet => -2,
+            Self::NoWarnings => -1,
             Self::Normal => 0,
             Self::Verbose => 1,
         }
@@ -43,16 +53,17 @@ impl Verbosity {
 }
 
 impl VerbositySpec {
-    /// Whether any verbosity flags (either `--verbose` or `--quiet`)
+    /// Whether any verbosity flags (either `--verbose`, `--quiet`, or `--no-warnings`)
     /// are present on the command line.
     pub fn is_present(&self) -> bool {
-        self.verbose != 0 || self.quiet != 0
+        self.verbose != 0 || self.quiet != 0 || self.no_warnings
     }
 
     /// Convert the verbosity specification to a [`tracing_core::LevelFilter`].
     pub fn as_trace(&self) -> String {
         let level = match self.integer_verbosity() {
-            i8::MIN..=-1 => tracing_core::LevelFilter::OFF,
+            i8::MIN..=-2 => tracing_core::LevelFilter::OFF,
+            -1 => tracing_core::LevelFilter::ERROR,
             0 => tracing_core::LevelFilter::ERROR,
             1 => tracing_core::LevelFilter::WARN,
             2 => tracing_core::LevelFilter::INFO,
@@ -63,13 +74,22 @@ impl VerbositySpec {
     }
 
     fn integer_verbosity(&self) -> i8 {
-        let int_level = (self.verbose as i8) - (self.quiet as i8);
-        if self.is_present() {
-            int_level
+        if self.no_warnings {
+            -1
         } else {
-            self.verbosity
-                .map(Verbosity::level_value)
-                .unwrap_or(int_level)
+            let int_level = if self.no_warnings {
+                -1
+            } else {
+                (self.verbose as i8) - (self.quiet as i8 * 2)
+            };
+
+            if self.is_present() {
+                int_level
+            } else {
+                self.verbosity
+                    .map(Verbosity::level_value)
+                    .unwrap_or(int_level)
+            }
         }
     }
 }
@@ -77,7 +97,8 @@ impl VerbositySpec {
 impl From<VerbositySpec> for Verbosity {
     fn from(spec: VerbositySpec) -> Self {
         match spec.integer_verbosity() {
-            v if v < 0 => Verbosity::Quiet,
+            v if v < -1 => Verbosity::Quiet,
+            -1 => Verbosity::NoWarnings,
             0 => Verbosity::Normal,
             _ => Verbosity::Verbose,
         }
@@ -92,6 +113,7 @@ mod tests {
     use crate::Verbosity;
 
     #[test_case(Verbosity::Quiet)]
+    #[test_case(Verbosity::NoWarnings)]
     #[test_case(Verbosity::Normal)]
     #[test_case(Verbosity::Verbose)]
     fn verbosity_serialization_identity(level: Verbosity) {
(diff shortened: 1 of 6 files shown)


utils/scarb-ui/src/verbosity.rs
```diff
@@ -14,6 +14,10 @@ pub enum Verbosity {
     ///
     /// String representation: `quiet`.
     Quiet,
+    /// Avoid printing warnings to standard output.
+    ///
+    /// String representation: `no-warnings`.
+    NoWarnings,
```
maciektr, line 20: This can be
```suggestion
    Error,
```
Indicating that it will print errors but not warnings

utils/scarb-ui/src/lib.rs
maciektr, line 1: You also need to change https://github.com/software-mansion/scarb/blob/dd56c334f0689f1deea70d4355879497aaf8b6c8/utils/scarb-ui/src/lib.rs#L106
so it is `> Verbosity::Quiet`.

utils/scarb-ui/src/lib.rs
```diff
@@ -140,7 +140,9 @@ impl Ui {
 
     /// Print a warning to the user.
     pub fn warn(&self, message: impl AsRef<str>) {
-        self.print(TypedMessage::styled("warn", "yellow", message.as_ref()))
+        if self.verbosity > Verbosity::NoWarnings {
```
maciektr, line 143: Why `>` and not `>=`? 🤔

utils/scarb-ui/src/args/verbosity.rs
```diff
@@ -100,29 +122,33 @@ mod tests {
...
-    #[test_case(1, 0, Verbosity::Quiet, tracing_core::LevelFilter::OFF)]
-    #[test_case(0, 0, Verbosity::Normal, tracing_core::LevelFilter::ERROR)]
-    #[test_case(0, 1, Verbosity::Verbose, tracing_core::LevelFilter::WARN)]
-    #[test_case(0, 2, Verbosity::Verbose, tracing_core::LevelFilter::INFO)]
-    #[test_case(0, 3, Verbosity::Verbose, tracing_core::LevelFilter::DEBUG)]
-    #[test_case(0, 4, Verbosity::Verbose, tracing_core::LevelFilter::TRACE)]
-    #[test_case(0, 5, Verbosity::Verbose, tracing_core::LevelFilter::TRACE)]
+    #[test_case(2, 0, false, Verbosity::Quiet, tracing_core::LevelFilter::OFF)]
+    #[test_case(1, 0, false, Verbosity::Quiet, tracing_core::LevelFilter::OFF)]
+    #[test_case(0, 0, false, Verbosity::Normal, tracing_core::LevelFilter::ERROR)]
```
maciektr, line 133: Why we do not need to change this tests, if we do: 
```
impl From<VerbositySpec> for Verbosity {
    fn from(spec: VerbositySpec) -> Self {
        match spec.integer_verbosity() {
            v if v < -1 => Verbosity::Quiet,
            -1 => Verbosity::NoWarnings,
```
Shouldn't `quiet=1, verbose=0` be the new one then? 🤔

utils/scarb-ui/src/args/verbosity.rs
```diff
@@ -63,21 +74,31 @@ impl VerbositySpec {
...
-        if self.is_present() {
-            int_level
+        if self.no_warnings {
+            -1
         } else {
-            self.verbosity
-                .map(Verbosity::level_value)
-                .unwrap_or(int_level)
+            let int_level = if self.no_warnings {
+                -1
```
maciektr, line 81: Why check twice? 🤔 
Can we do 
``` 
if self.no_warnings {
    return -1;
}
``` 
at the beginning of this function, and avoid changing anything else?

utils/scarb-ui/src/verbosity.rs
DelevoXDG: That was my initial thought too, but we have some outputs that aren’t warnings or errors, like:
```rs
    ws.config()ui().print(Status::new("Packaging", &pkg_id.to_string()));
```
I figured we probably want to keep these messages and only mute warnings, so `NoWarnings` seemed more fitting.

utils/scarb-ui/src/lib.rs
DelevoXDG: We can, but why? 😄

utils/scarb-ui/src/lib.rs
DelevoXDG: The goal here is to avoid printing warnings when verbosity is set to `NoWarnings`. 

Using `> Verbosity::NoWarnings` directly communicates that intent, unlike using `>= Verbosity::Normal` for this purpose.

utils/scarb-ui/src/args/verbosity.rs
DelevoXDG: I'm not sure I understand, `quiet=1, verbose=0` shouldn't produce any different verbosity level than before? That's not something this PR aims to change 🤔

utils/scarb-ui/src/lib.rs
maciektr: So we have:
- Quite: never prints
- Error: print() + error()
- Normal: print() + error() + warn()
- Verbose: print() + error() + warn() + verbose()
Otherwise, you would omit `print()` with --no-warnings, right? 🤔

utils/scarb-ui/src/lib.rs
maciektr: Right, sorry :D

utils/scarb-ui/src/args/verbosity.rs
DelevoXDG: Yep, I initially added the check at the beginning of the function, but later thought setting `int_level` first is just more in line with existing approach. Removed the extra check 👍🏻

utils/scarb-ui/src/args/verbosity.rs
maciektr: But you are changing `impl From<VerbositySpec> for Verbosity` which is used to parse it, right?

DelevoXDG: I am, but only to create conditions for setting `NoWarnings` verbosity level, without affecting impact of `--quiet` and `--verbose` flags on settings their respectful verbosity level.

utils/scarb-ui/src/verbosity.rs
maciektr: Yes, I would still use `ERROR` as it's more customary

utils/scarb-ui/src/lib.rs
DelevoXDG: Snap, you're right 😄 

Good catch, fixed it

utils/scarb-ui/src/verbosity.rs
DelevoXDG: Are you sure it's more customary? Unlike `--quiet`, `--verbose` and `--no-warnings`, `--error` imo doesn't seem like a flag directly related to verbosity, and reading docs will be required to understand what it actually does.

utils/scarb-ui/src/args/verbosity.rs
```diff
@@ -63,7 +74,12 @@ impl VerbositySpec {
     }
 
     fn integer_verbosity(&self) -> i8 {
-        let int_level = (self.verbose as i8) - (self.quiet as i8);
+        let int_level = if self.no_warnings {
+            -1
+        } else {
+            (self.verbose as i8) - (self.quiet as i8 * 2)
```
maciektr, line 80: Let's just conditionally subtract 1 if quiet is used instead of multiplication

utils/scarb-ui/src/verbosity.rs
maciektr: Hmm, I see what you mean, but personally I would stick with Error even if this does not fully reflect what we do, instead of using a negation 🤔 wdyt?

utils/scarb-ui/src/args/verbosity.rs
maciektr: Makes sense

maciektr (commented):
Hi! 👋 
I'd like to get back to this :D

utils/scarb-ui/src/args/verbosity.rs
DelevoXDG: Do you mean something like:

```suggestion
            (self.verbose as i8) - (self.quiet as i8) - (if self.quiet > 0 { 1 } else { 0 })
```

Not sure if it's better, but maybe we can just do:
```suggestion
            self.verbose as i8) - (self.quiet as i8) - (self.quiet as i8)
```
if you specifically don't like multiplication

utils/scarb-ui/src/verbosity.rs
DelevoXDG: I think it's useful for the flag to be directly related to what we do. 

> instead of using a negation 🤔 wdyt?

I'm not sure I see a problem in negation per se, we use it all the time already: `--no-build`, `--no-verify`, `--no-deps`, `--no-color` - the latter is directly related to output same as `--no-warnings`. 

For me, CURL comes to mind with:
- `--silent` -  hides progress / info output, i.e. hides all messages.
- `--show-error` - ensures errors are printed despite silence, while rest of the messages are silent.

I feel like if flag `--show-error` would actually keep both errors **and** regular messages, it would be misleading. Similarly here, I fee like `--error` flag would be misleading.

Thoughts?

utils/scarb-ui/src/args/verbosity.rs
maciektr: I just want this level to represent the number of times `-q` has been selected, starting from `-2`, so it's `-q = -2`, `-qq = -3`, `-qqq = -4` etc. The values we get now do not represent the actual integer level we mean by the input.

utils/scarb-ui/src/verbosity.rs
maciektr: I don't have a problem with `--no-warnings` CLI flag - it serves the purpose well. However, I don't particularly like the `NoWarnings` level of `Verbosity` (i.e. `SCARB_UI_VERBOSITY` env value). I still fell like usually verbosity levels are modeled around nouns and not negations (e.g. https://docs.rs/clap-verbosity/latest/clap_verbosity/enum.Level.html).

DelevoXDG: > I still fell like usually verbosity levels are modeled around nouns and not negations (e.g. https://docs.rs/clap-verbosity/latest/clap_verbosity/enum.Level.html).

Sure, but they mean something different in this context - if logging level is set to `Error`, it's expected **nothing** is printed except for errors.

If we're intending to keep logging level structure similar to the one you linked, we might consider handling `--no-warnings` separately altogether (with a separate env var as well).

We might want to add `ERROR` level in the future, and it would filter both everything that's below error level. 
Otherwise, having `SCARB_UI_VERBOSITY = ERROR` right now that is not behaving how you would **normally** expect it to seems misleading to me.

maciektr: > if logging level is set to Error, it's expected nothing is printed except for errors

I don't agree with this statement. If logging verbosity is set to `WARN` it doesn't mean that you only print warnings, but warning and errors. And for `INFO`, you would print three types of messages, etc. 
We categorize our messages into levels of verbosity and impose some arbitrary ordering between them. 
If "ERROR" sounds misleading to you, maybe `PRINT` or something like this? 
Otherwise, by extending your logic, the next level should be `NoWarningsNoErrorsNoPrints`, etc. 

> If we're intending to keep logging level structure similar to the one you linked, we might consider handling --no-warnings separately altogether (with a separate env var as well).

I thought about it as well, but IMHO it complicates this on both ends unnecessarily. 

>  Also, we might want to add level in the future that would filter everything that's below error level, and there's no other name than ERROR for that 😅

You can suggest different noun :D Apart from this very philosophical discussion the PR looks ready to me :D

maciektr (approved):

maciektr (approved):

PR #1690, opened 2024-10-25
Scarb Doc: cross item linkage

wawel37:
Waiting for [this](https://github.com/starkware-libs/cairo/pull/6522) to be merged.


Cargo.toml
```diff
@@ -35,30 +35,30 @@ anyhow = "1"
...
+cairo-lang-sierra = { path = "../cairo/crates/cairo-lang-sierra" }
+cairo-lang-sierra-generator = { path = "../cairo/crates/cairo-lang-sierra-generator" }
+cairo-lang-sierra-to-casm = { path = "../cairo/crates/cairo-lang-sierra-to-casm" }
+cairo-lang-sierra-type-size = { path = "../cairo/crates/cairo-lang-sierra-type-size" }
+cairo-lang-starknet = { path = "../cairo/crates/cairo-lang-starknet" }
+cairo-lang-starknet-classes = { path = "../cairo/crates/cairo-lang-starknet-classes" }
+cairo-lang-syntax = { path = "../cairo/crates/cairo-lang-syntax" }
+cairo-lang-test-plugin = { path = "../cairo/crates/cairo-lang-test-plugin" }
+cairo-lang-test-runner = { path = "../cairo/crates/cairo-lang-test-runner" }
+cairo-lang-utils = { path = "../cairo/crates/cairo-lang-utils" }
```
wawel37, line 61: This will get changed when changes are merged on the Cairo side

wawel37:
> Waiting for [this](https://github.com/starkware-libs/cairo/pull/6522) to be merged.

Review ready

extensions/scarb-doc/src/types.rs
```diff
@@ -378,6 +409,80 @@ impl Module {
...
+            ids.insert(trait_.item_data.id, trait_.item_data.clone());
+            trait_.trait_constants.iter().for_each(|item| {
+                ids.insert(item.item_data.id, item.item_data.clone());
+            });
+            trait_.trait_functions.iter().for_each(|item| {
+                ids.insert(item.item_data.id, item.item_data.clone());
+            });
+            trait_.trait_types.iter().for_each(|item| {
+                ids.insert(item.item_data.id, item.item_data.clone());
+            });
```
maciektr, line 464: Maybe those types should also implement `get_all_item_ids`, so we do not need to remember to update this function whenever they change.

maciektr (commented):
Not sure if it's only me or not, but passing this very specific `included_items` everywhere just bothers me a bit 😅 Maybe we can at least hide it behind some abstraction? Something like this is coming to my mind, probably not ideal though: https://gist.github.com/2028017ce043ca137530cd0281fdf55b.git
LMKWYT

wawel37:
> Not sure if it's only me or not, but passing this very specific `included_items` everywhere just bothers me a bit 😅 Maybe we can at least hide it behind some abstraction? Something like this is coming to my mind, probably not ideal though: https://gist.github.com/2028017ce043ca137530cd0281fdf55b.git LMKWYT

I totally agree. Moved the included items to the context. Tbh I don't see any more elegant solve for this problem here, as we need to keep track of those included items throughout creating types and later generating docs. Maybe in the future we can make some refactor, which would keep the context in mind

maciektr (approved):
💪

PR #1697, opened 2024-10-28
Remove Clone from TokenStream

maciektr:
**Stack**:
- #1698
- #1697 ⬅
- #1696
- #1695


plugins/cairo-lang-macro/src/types/token.rs
```diff
@@ -3,7 +3,7 @@ use std::fmt::Display;
 /// An abstract stream of Cairo tokens.
 ///
 /// This is both input and part of an output of a procedural macro.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
```
maciektr, line 6: Just so we can store Bump arena in TokenStream later.

PR #1700, opened 2024-10-28
Avoid unnecessary allocations when freeing stable structs

maciektr:
commit-id:55ec12c0

---

**Stack**:
- #1734
- #1722
- #1704
- #1700 ⬅

### DIFF
--- a/plugins/cairo-lang-macro/src/types/conversion.rs
+++ b/plugins/cairo-lang-macro/src/types/conversion.rs
@@ -39,7 +39,7 @@ impl ProcMacroResult {
 
     /// Convert to native Rust representation, without taking the ownership of the string.
     ///
-    /// Note that you still need to free the memory by calling `from_owned_stable`.
+    /// Note that you still need to free the memory by calling `free_owned_stable`.
     ///
     /// # Safety
     #[doc(hidden)]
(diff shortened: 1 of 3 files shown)


plugins/cairo-lang-macro/src/types/conversion.rs
```diff
@@ -109,11 +101,8 @@ impl TextSpan {
...
 
     #[doc(hidden)]
-    pub fn from_owned_stable(span: StableTextSpan) -> Self {
-        Self {
-            start: span.start,
-            end: span.end,
-        }
+    pub fn free_owned_stable(span: StableTextSpan) {
+        let _ = span;
     }
```
wawel37, line 106: Shouldn't this be in the StableTextSpan's impl?

scarb/src/compiler/plugin/proc_macro/ffi.rs
```diff
@@ -147,8 +147,8 @@ impl ProcMacroInstance {
         // Free the memory allocated by the `stable_token_stream`.
         // This will call `CString::from_raw` under the hood, to take ownership.
         unsafe {
-            TokenStream::from_owned_stable(stable_result.input);
-            TokenStream::from_owned_stable(stable_result.input_attr);
+            TokenStream::free_owned_stable(stable_result.input);
```
wawel37, line 150: Change comment in [this](https://github.com/software-mansion/scarb/blob/308f593261681a0c03a718ae3f985257fab34c67/scarb/src/compiler/plugin/proc_macro/ffi.rs#L136C53-L136C70) line, from `from_owned_stable` to `free_owned_stable`

wawel37 (approved):
Looks good. I was thinking about calling the [std::mem::drop](https://doc.rust-lang.org/std/mem/fn.drop.html) explicitly, but I don't think that's necessary

plugins/cairo-lang-macro/src/types/conversion.rs
maciektr: We implement all conversions in `cairo-lang-macro` not `cairo-lang-macro-stable`,  so the latter only defines the binary representation that is passed through the ffi barrier (so it changes as rarely as possible).

wawel37 (approved):

PR #1686, opened 2024-10-24
Add `ExpandInline` handler

Draggu:

PR #1685, opened 2024-10-24
Add `ExpandDerive` handler

Draggu:
**Stack**:
- #1684
- #1683
- #1682
- #1681
- #1680
- #1679
- #1678
- #1677
- #1676

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #1683, opened 2024-10-24
Add `DefinedMacros` handler

Draggu:
**Stack**:
- #1686
- #1685
- #1684
- #1683 ⬅
- #1718

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #1718, opened 2024-11-05
Add test helpers

Draggu:
**Stack**:
- #1686
- #1685
- #1684
- #1683
- #1718 ⬅

### DIFF
--- a/utils/scarb-test-support/src/proc_macro_server.rs
+++ b/utils/scarb-test-support/src/proc_macro_server.rs
@@ -1,0 +1,153 @@
+use crate::command::Scarb;
+use scarb_proc_macro_server_types::jsonrpc::RequestId;
+use scarb_proc_macro_server_types::jsonrpc::ResponseError;
+use scarb_proc_macro_server_types::jsonrpc::RpcRequest;
+use scarb_proc_macro_server_types::jsonrpc::RpcResponse;
+use scarb_proc_macro_server_types::methods::Method;
+use std::collections::HashMap;
+use std::io::BufRead;
+use std::io::BufReader;
+use std::io::Lines;
+use std::io::Write;
+use std::marker::PhantomData;
+use std::path::Path;
+use std::process::Child;
+use std::process::ChildStdin;
+use std::process::ChildStdout;
+use std::process::Stdio;
+
+pub const SIMPLE_MACROS: &str = r#"
+use cairo_lang_macro::{
+    ProcMacroResult,
+    TokenStream,
+    attribute_macro,
+    inline_macro,
+    derive_macro,
+    executable_attribute
+};
+
+executable_attribute!("some_executable");
+
+#[attribute_macro]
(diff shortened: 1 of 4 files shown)


scarb/tests/proc_macro_server_helpers/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+        serde_json::from_value(self.raw.result.unwrap()).unwrap()
+    }
+
+    #[allow(dead_code)]
+    pub fn error(self) -> ResponseError {
+        self.raw.error.unwrap()
+    }
+}
+
+pub struct TestProcMacroServer {
```
mkaput, line 84:
```suggestion
pub struct ProcMacroClient {
```

wdyt?

scarb/tests/proc_macro_server_helpers/mod.rs
mkaput, line 1: this should live in https://github.com/software-mansion/scarb/tree/main/utils/scarb-test-support/src I think?

scarb/tests/proc_macro_server_helpers/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+            } else {
+                self.responses.insert(raw_response.id, raw_response);
+            }
+        }
+    }
+}
+
+impl Drop for TestProcMacroServer {
+    fn drop(&mut self) {
+        self.inner.wait().unwrap();
```
mkaput, line 157: how are you terminating the process? This `Drop` is called before `TestProcMacroServer` fields are dropped: https://doc.rust-lang.org/reference/destructors.html#destructors.operation

scarb/tests/proc_macro_server_helpers/mod.rs
mkaput, line 1: I would like to see usages of this code in a single PR. Right now, you just added a bunch of dead code here.

mkaput (changes requested):

scarb/tests/proc_macro_server_helpers/mod.rs
mkaput, line 1: there no `mod` decl for this file

mkaput (changes requested):

utils/scarb-test-support/src/proc_macro_server.rs
```diff
@@ -0,0 +1,166 @@
...
+impl<M: Method> PendingRequest<M> {
+    fn new(id: RequestId) -> Self {
+        Self {
+            id,
+            _data: Default::default(),
+        }
+    }
+}
+
+pub struct Response<M: Method> {
```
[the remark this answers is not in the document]
Draggu: >Cannot we just always return Result<<M as Method>::Response, ResponseError> or RpcResponse? This seems unnecessary

This will not work if we want to do something more complex than 1 by 1 execution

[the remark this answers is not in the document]
Draggu: My bad, had `PendingRequest` in mind, you are right

mkaput (approved):

PR #1719, opened 2024-11-07
Add `RpcResponse` result helper

Draggu:


utils/scarb-proc-macro-server-types/src/jsonrpc.rs
```diff
@@ -18,6 +18,26 @@ pub struct RpcResponse {
...
+    /// or the error that occurred during the RPC request.
+    ///
+    /// # Returns
+    /// * `Ok(serde_json::Value)` if the RPC request was successful and contains a result.
+    /// * `Err(ResponseError)` if the RPC request failed and contains an error.
+    ///
+    /// # Panics
+    /// Panics if both `result` and `error` are present or both are `None`,
+    /// as this represents an invalid or inconsistent state of `RpcResponse`.
+    pub fn extract_result(self) -> Result<serde_json::Value, ResponseError> {
```
mkaput, line 32:
```suggestion
    pub fn into_result(self) -> Result<serde_json::Value, ResponseError> {
```

wdyt?

utils/scarb-proc-macro-server-types/src/jsonrpc.rs
```diff
@@ -18,6 +18,26 @@ pub struct RpcResponse {
...
+        match (self.result, self.error) {
+            (Some(response), None) => Ok(response),
+            (None, Some(error)) => Err(error),
+            (_, _) => unreachable!("RpcResponse must not have both result and error or neither."),
```
mkaput, line 36: that's risky assumption that this is unreachable, given that this struct construciton does not prohibit such case

I think it's better to assume precedence here and I think `error` should be preferred over response

utils/scarb-proc-macro-server-types/src/jsonrpc.rs
```diff
@@ -18,6 +18,26 @@ pub struct RpcResponse {
...
+        }
```
mkaput, line 37:
```suggestion
        match (self.result, self.error) {
            (_, Some(error)) => Err(error),
            (Some(response), None) => Ok(response),
        }
```

mkaput (approved):

PR #1736, opened 2024-11-12
Add `no_op_attribute` macro to `cairo-lang-macro-attributes`

DelevoXDG:
### Stack
- #1736 
- #1737 
- #1738

PR #1737, opened 2024-11-12
Prepare `cairo-lang-macro-attributes` release `0.1.1`

DelevoXDG:
### Stack
- #1736 
- #1737 
- #1738

### DIFF
--- a/plugins/cairo-lang-macro/Cargo.toml
+++ b/plugins/cairo-lang-macro/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cairo-lang-macro"
-version = "0.1.0"
+version = "0.1.1"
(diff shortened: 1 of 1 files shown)


maciektr (changes requested):
We do not need this change. `"0.1"` accepts all versions `>=0.1.0, <0.2.0`, so releasing `cairo-lang-macro-attributes` is enough. See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

DelevoXDG:
> We do not need this change. "0.1" accepts all versions >=0.1.0, <0.2.0, so releasing cairo-lang-macro-attributes is enough.

Sure, just thought it'd be better to be explicit, but I suppose it's really unnecessary here.

PR #1738, opened 2024-11-12
Allow defining attributes with no logic as proc macros

### LINKED ISSUE #1627: Allow defining attributes with no logic as proc macros
This may be useful to mark some element of larger entity modified with another proc macro - for instance mark field of an Enum as default, etc.

DelevoXDG:
### Stack
- #1736 
- #1737 
- #1738

Closes #1627

### DIFF
--- a/plugins/cairo-lang-macro/Cargo.toml
+++ b/plugins/cairo-lang-macro/Cargo.toml
@@ -13,8 +13,8 @@ readme = "README.md"
 repository.workspace = true
 
 [dependencies]
-cairo-lang-macro-attributes = "0.1.1"
-cairo-lang-macro-stable = "1"
+cairo-lang-macro-attributes = { path = "../cairo-lang-macro-attributes" }
+cairo-lang-macro-stable = { path = "../cairo-lang-macro-stable" }
 linkme.workspace = true
 serde = { workspace = true, optional = true }
 
--- a/scarb/Cargo.toml
+++ b/scarb/Cargo.toml
@@ -23,7 +23,7 @@ cairo-lang-filesystem.workspace = true
 cairo-lang-formatter.workspace = true
 cairo-lang-lowering.workspace = true
 cairo-lang-macro = { path = "../plugins/cairo-lang-macro" }
-cairo-lang-macro-stable = "1"
+cairo-lang-macro-stable = { path = "../plugins/cairo-lang-macro-stable" }
 cairo-lang-parser.workspace = true
 cairo-lang-semantic.workspace = true
 cairo-lang-sierra-to-casm.workspace = true
--- a/scarb/src/compiler/plugin/proc_macro/ffi.rs
+++ b/scarb/src/compiler/plugin/proc_macro/ffi.rs
@@ -111,7 +111,11 @@ impl ProcMacroInstance {
     pub fn declared_attributes_and_executables(&self) -> Vec<String> {
         self.get_expansions()
             .iter()
-            .filter(|e| e.kind == ExpansionKind::Attr || e.kind == ExpansionKind::Executable || e.kind == ExpansionKind::NoOp)
+            .filter(|e| {
+                e.kind == ExpansionKind::Attr
+                    || e.kind == ExpansionKind::Executable
+                    || e.kind == ExpansionKind::NoOp
+            })
             .map(|e| e.name.clone())
             .map(Into::into)
             .collect()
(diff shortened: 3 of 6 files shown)


scarb/src/compiler/plugin/proc_macro/ffi.rs
```diff
@@ -110,7 +111,11 @@ impl ProcMacroInstance {
     pub fn declared_attributes_and_executables(&self) -> Vec<String> {
```
DelevoXDG, line 111: Should we include no-op `declared_attributes` to `declared_attributes` as well, or is there a reason not to do so? 🤔

maciektr: We should not. The `declared_attributes_and_executables` declares attributes as allowed in the compiler, the `declared_attributes` is used by cairo-ls to obtain expandable attributes list.

plugins/cairo-lang-macro/Cargo.toml
```diff
@@ -13,8 +13,8 @@ readme = "README.md"
 repository.workspace = true
 
 [dependencies]
-cairo-lang-macro-attributes = "0.1.1"
-cairo-lang-macro-stable = "1"
+cairo-lang-macro-attributes = { path = "../cairo-lang-macro-attributes" }
+cairo-lang-macro-stable = { path = "../cairo-lang-macro-stable" }
```
maciektr, line 17:
```suggestion
```
Why change the stable one? 🤔 If we use `NO_OP_ATTR_PREFIX` to differentiate, there is no changed in the interface, right?

scarb/Cargo.toml
```diff
@@ -23,7 +23,7 @@ cairo-lang-filesystem.workspace = true
 cairo-lang-formatter.workspace = true
 cairo-lang-lowering.workspace = true
 cairo-lang-macro = { path = "../plugins/cairo-lang-macro" }
-cairo-lang-macro-stable = "1"
+cairo-lang-macro-stable = { path = "../plugins/cairo-lang-macro-stable" }
```
maciektr, line 26:
```suggestion
```
vide supra

maciektr (changes requested):

plugins/cairo-lang-macro/Cargo.toml
DelevoXDG: This is a temporary change so we can run tests, until the new version of `cairo-lang-macro-attributes` is released, as otherwise the same structs are not recognized due to one dependency specified as a path dependency and the other one as a registry dependency:
```
error[E0308]: arguments to this function are incorrect
   --> scarb/src/compiler/plugin/proc_macro/ffi.rs:182:13
    |
182 |             (self.plugin.vtable.expand)(item_name, stable_attr, stable_token_stream);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^            -----------  ------------------- expected `StableTokenStream`, found a different `StableTokenStream`
    |                                                    |
    |                                                    expected `StableTokenStream`, found a different `StableTokenStream`
    |
    = note: `StableTokenStream` and `StableTokenStream` have similar names, but are actually distinct types
note: `StableTokenStream` is defined in crate `cairo_lang_macro_stable`
   --> /Users/delevo/repos/scarb/plugins/cairo-lang-macro-stable/src/lib.rs:25:1
    |
25  | pub struct StableTokenStream {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: `StableTokenStream` is defined in crate `cairo_lang_macro_stable`
   --> /Users/delevo/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cairo-lang-macro-stable-1.0.0/src/lib.rs:25:1
```  

Afterwards, all changes that include use of path dependencies will be reverted.

Draggu (commented):
Are you sure we want to define these globally? no-op attributes are only metadata for other macros, as so placing them on code that is not using regular macros is pointless. I think we should make no-op assigned to regular macros so we can validate their placement.

Something like
```rs
#[attribute_macro(noops(a,b,c))] // Syntax can be changed, no opinion on that
fn my_macro(args: TokenStream, item: TokenStream) -> TokenStream {
    ...
}
```

Now we can validate if `a,b,c` are placed on code accessible to `my_macro`.

DelevoXDG:
Discussed offline, will be addressed on the compiler side

PR #1743, opened 2024-11-14
Implement quote macro.

wawel37:


wawel37:
It waits for [this](https://github.com/starkware-libs/cairo/pull/6707) to be merged

plugins/cairo-lang-quote/Cargo.toml
maciektr, line 1: Should this be re-exported from `cairo-lang-macro`?

maciektr (commented):
Pls rebase 🙏

scarb/tests/build_cairo_plugin.rs
```diff
@@ -1536,3 +1536,182 @@ fn can_expand_impl_inner_func_attrr() {
...
+          let db_val = SimpleParserDatabase::default();
+          let db = &db_val;
+          let code = r#"
+              fn main() -> felt252 {
+                5
+              }
+          "#;
+          let syntax_node = db.parse_virtual(code).unwrap();
+          let syntax_node_with_db = SyntaxNodeWithDb::new(&syntax_node, db);
+          let tokens = quote! {
```
mkaput, line 1689: I'm missing a test that is using/quoting Cairo-specific syntax, like `impl of`

plugins/cairo-lang-quote/Cargo.toml
wawel37: I mean we can do that for sure, but i thought we want to threat `quote` macro as a separate crate. Same way as it's done in rust. As for now, there is no problem with reexporting it from `cairo-lang-macro`, but in the future, some changes might come to quote, which would make `cairo-lang-macro` less stable. What do you think?

scarb/tests/build_cairo_plugin.rs
wawel37: Added tests for `impl of`, coming from either syntax_node, or putting it directly into the quote

plugins/cairo-lang-quote/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+    };
+
+    let token_iter = input.into_iter().peekable();
+    process_token_stream(token_iter, &mut parsed_input);
+
+    for quote_token in parsed_input.iter() {
+        match quote_token {
+            QuoteToken::Content(content) => {
+                output_token_stream.extend(rust_quote! {
+                  quote_macro_result.push_token(TokenTree::Ident(Token::new(#content.to_string(), None)));
```
Draggu, line 89: Use full paths, as this can break user imports. Paths starting with `::` are recognized as absolute paths starting from some crate root rather than a module. Add tests for colliding `use` statements in user code.

```rs
use my_crate::my_module::Token;

fn my_macro(){
    quote!(some content)
}
```
where `Token::new` has different signature would fail to compile

plugins/cairo-lang-quote/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+                });
+            }
+            QuoteToken::Var(ident) => {
+                output_token_stream.extend(rust_quote! {
+                  quote_macro_result.extend(TokenStream::from_primitive_token_stream(#ident.to_primitive_token_stream()));
```
Draggu, line 94: vide supra

plugins/cairo-lang-quote/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+                        output.push(QuoteToken::Var(var_ident))
+                    }
+                } else {
+                    output.push(QuoteToken::Content(punct.to_string()));
+                }
+                was_last_ident = false;
+            }
+            RustTokenTree::Ident(ident) => {
+                if was_last_ident {
+                    output.push(QuoteToken::Content(format!(" {}", ident)));
```
Draggu, line 59: Maybe it would be better to add `QuoteToken::Whitespace` and move whitespace logic to the place where you create a token stream out of it?

Draggu (changes requested):

mkaput (approved):

plugins/cairo-lang-quote/src/lib.rs
Draggu: https://doc.rust-lang.org/reference/paths.html#path-qualifiers

The current code can still break if there is `mod cairo_lang_macro` in user code and `cairo_lang_macro` crate is used with `as` alias. Prefix it with `::` to make sure that rustc will interpret `::cairo_lang_macro` as crate and NOT module.

Draggu (approved):

plugins/cairo-lang-quote/Cargo.toml
maciektr: Ok, makes sense :D

scarb/tests/build_cairo_plugin.rs
```diff
@@ -1573,3 +1574,287 @@ fn can_expand_impl_inner_func_attrr() {
...
         "#});
 }
+
+#[test]
+fn can_use_quote() {
+    let temp = TempDir::new().unwrap();
+    let t = temp.child("some");
+    let cairo_quote_path = CAIRO_LANG_QUOTE_PATH.to_string();
+    CairoPluginProjectBuilder::default()
+        .add_dep(formatdoc! {r#"cairo-lang-quote = {{ path={cairo_quote_path} }}"#})
```
maciektr, line 1584: Nit: could be hidden in `add_quote_dep()` or similar helper.

scarb/tests/build_cairo_plugin.rs
```diff
@@ -1573,3 +1574,287 @@ fn can_expand_impl_inner_func_attrr() {
...
+        .add_dep(r#"cairo-lang-primitive-token = "1.0.0""#)
```
maciektr, line 1585: Vide supra

scarb/tests/build_cairo_plugin.rs
```diff
@@ -1573,3 +1574,287 @@ fn can_expand_impl_inner_func_attrr() {
...
+        "#})
+        .build(&project);
+
+    Scarb::quick_snapbox()
+        .arg("cairo-run")
+        // Disable output from Cargo.
+        .env("CARGO_TERM_QUIET", "true")
+        .current_dir(&project)
+        .assert()
+        .success();
```
maciektr, line 1615: Can we add assert on returned value, to make sure it is 5 indeed?

plugins/cairo-lang-quote/Cargo.toml
maciektr: After reading more into this: I still wonder what's the benefit of using it separately? 🤔 
It does rely **heavily** on the `cairo-lang-macro` API, which is not indicated in the dependency system in any way right now 🤔

maciektr (approved):

maciektr:
💪 💪 💪

plugins/cairo-lang-quote/Cargo.toml
wawel37: Ok, after thinking about it I think you're right. The quote macro is tightly related to the `cairo-lang-macro`. I placed the reexports of `quote` inside the `cairo-lang-macro`

scarb/tests/build_cairo_plugin.rs
wawel37: I did assert the result on 2 last tests, as they are using the attribute macro. As it turns out, the `scarb expand` doesn't really work with inline macros

wawel37: In those inline macro tests, we test for the returned value and its type, as the `cairo-run` would throw an error, when the main function haven't returned the proper type

plugins/cairo-lang-macro/src/types/token.rs
```diff
@@ -278,6 +281,32 @@ impl TokenStream {
...
+                            start: stable_span.start as u32,
+                            end: stable_span.end as u32,
+                        }),
+                    ))
+                })
+                .collect(),
+        )
+    }
+
+    pub fn extend(&mut self, token_stream: Self) {
```
Draggu, line 303: https://doc.rust-lang.org/std/iter/trait.Extend.html

plugins/cairo-lang-quote/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+
+#[derive(Debug)]
+enum QuoteToken {
+    Var(Ident),
+    Content(String),
+    Whitespace,
+}
+
+impl QuoteToken {
+    pub fn from_delimiter(delimiter: Delimiter, is_end: bool) -> Self {
```
Draggu, line 17:
```suggestion
    pub fn from_delimiter(delimiter: Delimiter, delimiter_variant: DelimiterVariant) -> Self {
```

```
enum DelimiterVariant {
    Open,
    Close
}
```

up to you

plugins/cairo-lang-macro/src/types/token.rs
maciektr: good call!

PR #1748, opened 2024-11-14
Make TextSpan u32 to match the compiler one

maciektr:
commit-id:107b5e17

---

**Stack**:
- #1749
- #1748 ⬅

### DIFF
--- a/plugins/cairo-lang-macro/src/types/token.rs
+++ b/plugins/cairo-lang-macro/src/types/token.rs
@@ -95,12 +95,14 @@ impl TokenTree {
     }
 }
 
+pub type TextOffset = u32;
+
 /// A range of text offsets that form a span (like text selection).
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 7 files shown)


plugins/cairo-lang-macro/src/types/token.rs
```diff
@@ -95,12 +95,14 @@ impl TokenTree {
     }
 }
 
+pub type TextOffset = u32;
```
DelevoXDG, line 98: Should we add such alias if this `TextOffset` does not match compiler one anyway?

maciektr: We don't need to structure our spans exactly the same as compiler ones - we just need to be able to convert between those two. 

I don't think structuring offsets into new types gives us any advantage - let me know if you think otherwise. 
In practice, type alias acts just as a comment, so I don't see anything bad with that.

DelevoXDG: > I don't think structuring offsets into new types gives us any advantage - let me know if you think otherwise.

I suppose you're right, keeping it as a comment makes sense 👍🏻

DelevoXDG (approved):

PR #1781, opened 2024-11-26
feat: allow custom names for proc macros

remybar:
At the moment, the macro name is the name of the function tagged with `#[attribute_macro]/#[inline_macro]/#[derive_macro]`.

In Cairo, it is quite common to use the `::` separator in macro names, but as this separator is not allowed in function names, we cannot use it for Cairo proc macros.

This PR adds string litteral argument to allow to set a custom name for a proc macros.

For example, here is a proc macro to process `dojo::model` attributes:
```rust
#[attribute_macro("dojo::model")
pub fn dojo_model(args: TokenStream, token_stream: TokenStream) -> ProcMacroResult {
}
```


maciektr:
Replacing with: https://github.com/software-mansion/scarb/pull/1891

PR #1789, opened 2024-11-28
Aot macro compilation - magic script

wawel37:


scarb/src/ops/package.rs
```diff
@@ -168,12 +169,22 @@ fn package_one_impl(
...
     let num_files = recipe.len();
+    let target_dir = ws.target_dir().child("package");
+
+    if let Some(script_definition) = pkg.manifest.scripts.get("package") {
+        if pkg.is_cairo_plugin() {
+            ws.target_dir()
+                .child("scarb")
+                .child("cairo-plugin")
+                .path_existent()?;
+            ops::execute_script(script_definition, &[], ws, pkg.root(), None)?;
```
DelevoXDG, line 180: Should we maybe emit warning/info when we package binaries via the scripts? 🤔

scarb/src/ops/package.rs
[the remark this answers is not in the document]
mkaput: definitely

scarb/src/ops/package.rs
```diff
@@ -168,12 +169,22 @@ fn package_one_impl(
...
+        }
+    }
```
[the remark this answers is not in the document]
mkaput: I don't see why we should limit this to cairo plugins only

scarb/src/ops/package.rs
mkaput, line 180: the script should be called with `target_dir` as PWD, not package root

mkaput (changes requested):

scarb/src/ops/package.rs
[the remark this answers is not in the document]
wawel37: I thought we wanted to run this in case of target being the plugin, but sure, we can get rid of that limitation

scarb/src/ops/package.rs
```diff
@@ -168,12 +169,22 @@ fn package_one_impl(
...
     let recipe = prepare_archive_recipe(pkg, opts, ws)?;
     let num_files = recipe.len();
+    let target_dir = ws.target_dir().child("package");
+
+    if let Some(script_definition) = pkg.manifest.scripts.get("package") {
+        if pkg.is_cairo_plugin() {
+            ws.target_dir()
+                .child("scarb")
+                .child("cairo-plugin")
+                .path_existent()?;
```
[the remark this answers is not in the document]
wawel37: Maybe I misunderstood the concept of this. If it's the scripts responsibility of creating the directory, we sure can get rid of that 👍

scarb/src/ops/package.rs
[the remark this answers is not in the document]
wawel37: Added them 👍

mkaput (changes requested):

scarb/tests/package.rs
```diff
@@ -1525,3 +1525,57 @@ fn package_with_publish_disabled() {
...
+
+#[test]
+fn package_with_package_script() {
+    let t = TempDir::new().unwrap();
+
+    #[cfg(not(windows))]
+    let script_name = "script.sh";
+    #[cfg(not(windows))]
+    let script_code = indoc! { r#"
+      touch text.txt
```
mkaput, line 1537: would be nice if this script uses something from a package

mkaput (changes requested):

scarb/tests/package.rs
```diff
@@ -1535,6 +1535,8 @@ fn package_with_package_script() {
     #[cfg(not(windows))]
     let script_code = indoc! { r#"
       touch text.txt
+      cd ../..
```
mkaput, line 1538: why are you cd-ing? everything needed for building should be copied into package directory already 🤔

mkaput (changes requested):

scarb/tests/package.rs
```diff
@@ -1525,3 +1525,68 @@ fn package_with_publish_disabled() {
...
+        .arg("package")
+        .current_dir(&t)
+        .assert()
+        .success();
+
+    assert!(Path::new(
+        &t.to_path_buf()
+            .join("target")
+            .join("package")
+            .join("text.txt"),
```
DelevoXDG, line 1588: Maybe we should check whether something gets produces in `target/scarb/cairo-plugin'? 

I've tried running this locally and it doesn't seem to produce anything, possibly due to `prepare_archive_recipe` not copying contents of `target`.

I've also tried changing `.gitignore` to exclude `target/scarb/cairo-plugin`, but it didn't seem to help. This is probably also something we should address:
1. The main purpose of the package script is to produce binaries in `target/scarb/cairo-plugin`. We might want to include warning on this directory being empty
2. I think we should include `target/scarb/cairo-plugin` files regardless of whether they are .gitignore'd, as we probably don't want to force users to include binaries in their repos.

scarb/tests/package.rs
wawel37: Moved all included items into the `target/package`

scarb/tests/package.rs
wawel37: Previous version didn't have the packaging the result dlls implemented yet, but right now it's done

scarb/src/core/manifest/mod.rs
```diff
@@ -75,6 +75,7 @@ pub struct ManifestMetadata {
     pub license_file: Option<Utf8PathBuf>,
     pub readme: Option<Utf8PathBuf>,
     pub repository: Option<String>,
+    pub include: Option<Vec<String>>,
```
mkaput, line 78: this field should be a separate PR

wawel37: I just thought that adding this single field isn't that much of a big change to make it as a separate PR, but I see your point, should i convert this to stacked PR?

mkaput: I think so

scarb/tests/package.rs
```diff
@@ -1525,3 +1526,105 @@ fn package_with_publish_disabled() {
...
+
+#[test]
+fn package_with_package_script() {
+    let t = TempDir::new().unwrap();
+
+    #[cfg(not(windows))]
+    let script_code = indoc! { r#"cargo build --release && mkdir -p ../scarb/cairo-plugin && cp target/release/libfoo.so ../scarb/cairo-plugin"#};
+
+    #[cfg(windows)]
+    let script_code = indoc! { r#"cargo build --release && mkdir -p ../scarb/cairo-plugin && cp target/release/libfoo.dll ../scarb/cairo-plugin"#};
```
DelevoXDG, line 1538: I think the users are not supposed to use relatives paths like `../scarb/cairo-plugin` or `../scarb/cairo-plugin`, and binaries should not be copied somewhere outside of the cwd tree.

The binaries imo should either:
1. Be required to be placed in `target/scarb/cairo-plugin/`
2. Be automatically located regardless of their path and then placed in the correct directory

maciektr:
Replacing with https://github.com/software-mansion/scarb/pull/1836

PR #1791, opened 2024-11-29
Prepare release `2.9.1` 

maciektr:


Cargo.toml
```diff
@@ -21,7 +21,7 @@ members = [
 "resolver" = "2"
 
 [workspace.package]
-version = "2.8.5"
```
DelevoXDG, line 24: Why did we skip over 2.9.0 that included `d453e4197414990ea803190dc2af8375471cbb1f` though? 🤔

DelevoXDG (approved):

Cargo.toml
DelevoXDG: Discussed offline

PR #1792, opened 2024-11-30
fix: bump bytes to 1.9.0

chenrui333:
as bytes 1.6.0 got yanked, bump it to 1.9.0

relates to https://github.com/Homebrew/homebrew-core/pull/199546


maciektr (approved):
Thank you so much! ❤️

PR #1793, opened 2024-11-30
Add optional argument to dump scoped Sierra statement weights

m-kus:
This PR introduces an optional cli argument that enables profiling and tells runner to dump scoped Sierra statement weights to a file. This is useful for profiling standalone large Cairo programs which also operate on large input values (read from file).

Scoped Sierra statement weights are aggregated by respective call stack (with multiple recursive calls collapsed) — that way we get a clearer picture of the program execution.

The format of the output file is FlameGraph compatible, so you can do `flamegraph.pl trace.txt > graph.svg` and get a nice chart.

If you'd like I can convert to pprof compatible format instead.


maciektr:
Hi @m-kus !

The profiler from `cairo-lang-runner` crate is meant to be mostly used as a helper for internal work on compiler debugging / development. It may not be stable / supported in future. For this reason, we do not want to make it part of our public interface.  

We suggest using https://github.com/software-mansion/cairo-profiler instead.

PR #1803, opened 2024-12-03
Fix label displaying in case 2 part linakge in scarb doc

wawel37:


maciektr (approved):

extensions/scarb-doc/src/docs_generation/markdown/context.rs
```diff
@@ -62,9 +62,9 @@ impl<'a> MarkdownGenerationContext<'a> {
                     }
                     _ => path_to_file_link(&resolved_item.full_path()),
                 },
-                None => link.to_string(),
+                None => link.path.clone().unwrap_or(link.label.clone()),
```
DelevoXDG, line 65: Should we be defaulting to treating label as link? 🤔

wawel37: If it's a common one part `[Item]` link, then label is also a path in this case. 
- Put the path as the link. (2 part)
- If the path is None (it implies that the link is a one part), we use the label, which is also a path

DelevoXDG (approved):

PR #1804, opened 2024-12-03
Use new ls package

wawel37:


maciektr (approved):

Cargo.lock
```diff
@@ -581,17 +594,42 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
- "cairo-lang-filesystem",
- "cairo-lang-lowering",
- "cairo-lang-parser",
- "cairo-lang-project",
- "cairo-lang-semantic",
- "cairo-lang-sierra",
- "cairo-lang-sierra-generator",
- "cairo-lang-syntax",
- "cairo-lang-utils",
+ "cairo-lang-defs 2.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
```
mkaput, line 597: you're now shipping two copies of the compiler that are potentially incompatible, that's bad

I'm working on a proper solution for this but it will take me some more hours to ship

mkaput (changes requested):

PR #1811, opened 2024-12-04
Refactor: split proc macro tests into multiple files

maciektr:
commit-id:b0254722


scarb/tests/proc_macro_build.rs
```diff
@@ -0,0 +1,436 @@
...
+        .arg("build")
+        // Disable output from Cargo.
+        .env("CARGO_TERM_QUIET", "true")
+        .current_dir(&project)
+        .assert()
+        .failure()
+        .stdout_matches(indoc! {r#"
+        [..]Compiling some v1.0.0 ([..]Scarb.toml)
+        [..]Compiling hello v1.0.0 ([..]Scarb.toml)
+        error: Plugin diagnostic: Unsupported attribute.
```
DelevoXDG, line 429: Seems like a `proc_macro_diagnostics.rs` test? 🤔

scarb/tests/proc_macro_expand.rs
DelevoXDG, line 1: Maybe we should keep diagnostic tests in this file?
The current division makes sense, but similar tests being in different files seems a bit confusing to me.

scarb/tests/proc_macro_build.rs
maciektr: This diagnostic is not raised by our procedural macros system, but rather by attribute validations in the compiler itself. So I would not move it there, as it's not a diagnostic that we control.

scarb/tests/proc_macro_expand.rs
maciektr: I do agree they are similar, but in a sense all of them are. My main pain point with the current setup is, that the test file grew too big to navigate comfortably. Merging *expand and *diagnostics, we already go over 1000 lines, even though we might want to add more diagnostics in the future. So personally, I would separate them even if it feels a bit superficial right now. 
Also, I see a logical division between those two - in *expand we check if the Cairo code is expanded by proc macros correctly, while *diagnostics is more concerned with plugin raised diagnostics (returned through the result field), or code mappings / location translations. 
Let me know if any of these arguments resonate with you or not?

DelevoXDG: > Let me know if any of these arguments resonate with you or not?

They do, but it seems to me that we already have some files with over 1000 lines (`build`, `package`). While it does introduce a slight complication, most IDEs support folding code including tests anyway. Additionally, keeping similar tests together in the same file can make it easier to maintain them and to find and reference them when adding new tests, which seems more important to me.

I don't have a super strong say on this though, if my arguments are not convincing enough, the current way is ok as well 😄

DelevoXDG (approved):

PR #1836, opened 2024-12-10
AOT package script

### LINKED ISSUE #1771: Add `package` magic script
Add a new “magic” script (magic in that sense that Scarb has some logic that makes use of this script, like the test one being also used by `scarb test` command) called `package`. While running `scarb package` it will look for this script and, if present, it will execute it after preparing the tarball directory and before making an archive. The script will be run with `PWD` set to tarball preparation directory (unlike package directory, this is important and is exceptional!).

This script will be used to automate prebuilding macros (as an alternative to forcing devs to do this manually, see #1770). The rough procedure should look like this:

1. Create directory `target/scarb/cairo-plugin` in tarball preparation directory. 
2. For each target triplet (see platform support for full list):
    1. call `rustup component add $TARGET` (avoid doing that if we can cheaply detect if it’s already installed)
    2. call `cargo build --target $TARGET` to build the plugin (note: ensure cargo shows a clear message if any build tools are missing)
    3. copy generated `*.so/*.dll` + debug info files to `target/scarb/cairo-plugin` directory, following this naming pattern:
        `${PACKAGE_NAME}_v${PACKAGE_VERSION}_${TARGET}.${so|dylib|dll} `
3. Ensure no other garbage is left in tarball directory. We do not want to publish unwanted files!

## Example: usage by Starknet Foundry

Starknet Foundry will use the script mechanism in their CI for automatic packaging. It will implement the packaging procedure as a bash script that will be committed to the repository and hooked as a package script.
We will ask Dojo people to copy Foundry’s approach.
So, Foundry plugin will change its [Scarb.toml](https://github.com/foundry-rs/starknet-foundry/blob/867c6457cd4bbf31a932de1821f60200a8ab823c/crates/snforge-scarb-plugin/Scarb.toml) to look as follows:

```toml
[package]
name = "snforge_scarb_plugin"
version = "0.33.1"
edition = "2024_07"

[scripts]
package = "prebuild.sh"
```

> [!NOTE]
> No such task was created in Foundry repo.

maciektr:
fixes #1771
commit-id:40843cae

---

**Stack**:
- #1846
- #1836 ⬅

### DIFF
--- a/scarb/src/ops/package.rs
+++ b/scarb/src/ops/package.rs
@@ -37,6 +37,8 @@ const RESERVED_FILES: &[&str] = &[
     VCS_INFO_FILE_NAME,
 ];
 
+const SCARB_PREPACKAGE_SCRIPT_NAME: &str = "package";
+
 #[derive(Clone)]
 pub struct PackageOpts {
     pub allow_dirty: bool,
@@ -166,6 +168,8 @@ fn package_one_impl(
         check_metadata(pkg, ws.config())?;
     }
 
+    run_prepackage_script(pkg, ws)?;
+
     let recipe = prepare_archive_recipe(pkg, opts, ws)?;
     let num_files = recipe.len();
 
@@ -216,6 +220,24 @@ fn package_one_impl(
     ));
 
     Ok(dst)
+}
+
+fn run_prepackage_script(package: &Package, ws: &Workspace<'_>) -> Result<()> {
+    if let Some(script_definition) = package.manifest.scripts.get(SCARB_PREPACKAGE_SCRIPT_NAME) {
+        // Ensure no two instance of Scarb will run `package` script at the same time.
+        let _guard = ws.target_dir().child("scarb").advisory_lock(
+            ".scarb-package.lock",
+            "`package` script",
+            ws.config(),
+        );
+        ws.config().ui().print(Status::new(
+            "Running",
+            &format!("`package` script for `{}`", package.id.name),
+        ));
+        ops::execute_script(script_definition, &[], ws, package.root(), None)
+    } else {
+        Ok(())
+    }
 }
 
 fn list_one_impl(
--- a/scarb/tests/package.rs
+++ b/scarb/tests/package.rs
@@ -10,6 +10,7 @@ use assert_fs::prelude::*;
 use assert_fs::TempDir;
 use indoc::{formatdoc, indoc};
 use itertools::Itertools;
+use libloading::library_filename;
 use scarb::DEFAULT_TARGET_DIR_NAME;
 use scarb_build_metadata::CAIRO_VERSION;
 use scarb_test_support::cairo_plugin_project_builder::CairoPluginProjectBuilder;
@@ -22,7 +23,7 @@ use scarb_test_support::workspace_builder::WorkspaceBuilder;
 use test_case::test_case;
 
 struct PackageChecker {
-    actual_files: HashMap<PathBuf, String>,
+    actual_files: HashMap<PathBuf, Vec<u8>>,
     base_name: PathBuf,
 }
 
@@ -39,7 +40,7 @@ impl PackageChecker {
     fn assert(path: &Path) -> Self {
         let mut archive = Self::open(path);
 
-        let actual_files: HashMap<PathBuf, String> = archive
+        let actual_files: HashMap<PathBuf, Vec<u8>> = archive
             .entries()
             .expect("failed to get archive entries")
             .map(|entry| {
@@ -48,10 +49,11 @@ impl PackageChecker {
                     .path()
                     .expect("failed to get archive entry path")
                     .into_owned();
-                let mut contents = String::new();
+                let mut contents: Vec<u8> = Vec::new();
                 entry
-                    .read_to_string(&mut contents)
+                    .read_to_end(&mut contents)
                     .expect("failed to read archive entry contents");
+
                 (name, contents)
             })
             .collect();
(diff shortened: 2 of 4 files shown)


scarb/src/ops/package.rs
```diff
@@ -166,14 +173,44 @@ fn package_one_impl(
...
+        ws.config().ui().print(Status::new(
+            "Running package script with package",
+            &pkg_id.to_string(),
+        ));
+
+        if let Some(includes) = pkg.manifest.metadata.include.clone() {
+            copy_items_to_target_dir(includes, target_dir_path)?;
+        }
+
+        execute_script(script_definition, &[], ws, target_dir_path, None)?;
```
maciektr, line 189: Why are we executing the script in `target/package`?

It seems to me it should be either executed in the root dir, or if we want avoid modifying any unnecessary files there, in `target/package/<name>-<version>`?

Originally posted in: https://github.com/software-mansion/scarb/pull/1789#discussion_r1877804646

scarb/tests/package.rs
```diff
@@ -1525,3 +1526,105 @@ fn package_with_publish_disabled() {
...
+    #[cfg(windows)]
+    let script_code = indoc! { r#"cargo build --release && mkdir -p ../scarb/cairo-plugin && cp target/release/libfoo.dll ../scarb/cairo-plugin"#};
+
+    CairoPluginProjectBuilder::start()
+        .name("foo")
+        .scarb_project(|b| {
+            b.name("foo")
+                .version("1.0.0")
+                .manifest_package_extra(formatdoc! {r#"
+      include = ["Cargo.lock", "Cargo.toml", "src"]
```
maciektr, line 1546: I'm not sure it's supposed to work that way 🤔 

These files are included in the tarball by default, shouldn't we use all the files that are included in the tarball by default when running a package script?

Originally posted in: https://github.com/software-mansion/scarb/pull/1789#discussion_r1877863937

scarb/tests/package.rs
```diff
@@ -1525,3 +1526,105 @@ fn package_with_publish_disabled() {
...
+
+#[test]
+fn package_with_package_script() {
+    let t = TempDir::new().unwrap();
+
+    #[cfg(not(windows))]
+    let script_code = indoc! { r#"cargo build --release && mkdir -p ../scarb/cairo-plugin && cp target/release/libfoo.so ../scarb/cairo-plugin"#};
+
+    #[cfg(windows)]
+    let script_code = indoc! { r#"cargo build --release && mkdir -p ../scarb/cairo-plugin && cp target/release/libfoo.dll ../scarb/cairo-plugin"#};
```
maciektr, line 1538: I think the users are not supposed to use relatives paths like `../scarb/cairo-plugin` or `../scarb/cairo-plugin`, and binaries should not be copied somewhere outside of the cwd tree.

The binaries imo should either:
1. Be required to be placed in `target/scarb/cairo-plugin/`
2. Be automatically located regardless of their path and then placed in the correct directory

Originally posted in: https://github.com/software-mansion/scarb/pull/1789#discussion_r1877886709

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -197,6 +199,7 @@ pub struct TomlPackage {
     pub license_file: Option<MaybeWorkspaceField<Utf8PathBuf>>,
     pub readme: Option<MaybeWorkspaceField<PathOrBool>>,
     pub repository: Option<MaybeWorkspaceField<String>>,
+    pub include: Option<MaybeWorkspaceField<Vec<String>>>,
```
maciektr, line 202: this field should be a separate PR

Originally posted in: https://github.com/software-mansion/scarb/pull/1789#discussion_r1872775817

mkaput (approved):

scarb/tests/package.rs
```diff
@@ -1636,3 +1640,90 @@ fn files_outside_package_cannot_be_included() {
...
         "#});
 }
+
+#[test]
+fn package_script_is_run() {
+    let t = TempDir::new().unwrap();
+    simple_project()
+        .manifest_package_extra(indoc! {r#"
+            [scripts]
+            package = "echo 'Hello!'"
```
DelevoXDG, line 1650: In the original issue, it was mentioned that the `package` field specifies the path to a package script, not the script’s content itself. This still seems like the preferred approach to me.

A script is unlikely to be a one-liner as in these tests. Generating binaries for multiple platforms and naming them accordingly will likely require a few lines of code 😄. I’m not sure we want to keep all of this within the manifest. Additionally, having the script as a separate file would make it easier to dry-run the script. I don’t think we want to force users to run `scarb package` just for that.

scarb/tests/package.rs
```diff
@@ -102,11 +104,13 @@ impl PackageChecker {
         self
     }
 
-    fn read_file(&self, path: impl AsRef<Path>) -> &str {
+    fn read_file(&self, path: impl AsRef<Path>) -> String {
```
DelevoXDG, line 107: Why this change? 🤔

scarb/tests/package.rs
mkaput: A path to a script is also a valid script (+/- the file is `chmod +x`ed which we can expect people to have anyway). This works as I expected.

image

scarb/tests/package.rs
maciektr: I've changed `actual_files` from `HashMap<PathBuf, String>,` to `HashMap<PathBuf, Vec<u8>>` so it can store files that are not valid utf-8 Strings - i.e. the dll binaries. It has to be stored in memory, as it's unpacked from a tarball without extracting to disk. 

Then I just converted it back to utf-8 string with `to_string_loosly`, which was the first thing that came to my mind, which requires allocations (thus String instead of &str). 
Yes, it can still be `str` when used with `std::str::from_utf8` - changed it back. It does not really matter here (it's just one allocation, just tests), but you are right it can be avoided :D .

scarb/tests/package.rs
maciektr: It's handled by our scripts system, so there is no need to test it here, as the running of scripts is handled extensively in `tests/scripts.rs`

DelevoXDG (approved):

PR #1839, opened 2024-12-11
feat: executable plugin

FroyaTheHen:
commit-id:eefec831

---

**Stack**:
- #1840
- #1839 ⬅


scarb/src/lib.rs
```diff
@@ -42,3 +42,4 @@ pub const TEST_ASSERTS_PLUGIN_NAME: &str = "assert_macros";
 pub const CAIRO_RUN_PLUGIN_NAME: &str = "cairo_run";
 pub const CARGO_MANIFEST_FILE_NAME: &str = "Cargo.toml";
 pub const CARGO_LOCK_FILE_NAME: &str = "Cargo.lock";
+pub const EXECUTABLE_PLUGIN_NAME: &str = "executable";
```
maciektr, line 45: Would you mind renaming the plugin to `cairo_execute`? I feel like it would be more idiomatic to our plugin names.

maciektr (approved):

scarb/src/lib.rs
FroyaTheHen: not at all. just to be clear, you mean the value not the const
```suggestion
pub const EXECUTABLE_PLUGIN_NAME: &str = "cairo_execute";
```
??

maciektr: Yes, here and everywhere where it is needed (i.e. in scarblib templates).

scarb/scarblib/executable/Scarb.toml
maciektr, line 1: Pls change directory name as well :D

maciektr (approved):

utils/scarb-test-support/src/project_builder.rs
```diff
@@ -114,6 +114,10 @@ impl ProjectBuilder {
         self.dep_builtin("starknet")
     }
 
+    pub fn dep_executable(self) -> Self {
```
Draggu, line 117:
```suggestion
    pub fn dep_cairo_execute(self) -> Self {
```

maciektr (approved):

PR #1840, opened 2024-12-11
feat: compile executable

FroyaTheHen:
commit-id:4510e5a4

### DIFF
--- a/scarb/src/compiler/compilers/executable.rs
+++ b/scarb/src/compiler/compilers/executable.rs
@@ -1,0 +1,47 @@
+use crate::compiler::helpers::write_json;
+use crate::compiler::helpers::{build_compiler_config, collect_main_crate_ids};
+use crate::compiler::{CairoCompilationUnit, CompilationUnitAttributes, Compiler};
+use crate::core::{TargetKind, Workspace};
+use anyhow::Result;
+use cairo_lang_compiler::db::RootDatabase;
+use cairo_lang_executable::executable::Executable;
+use tracing::trace_span;
+
+pub struct ExecutableCompiler;
+
+impl Compiler for ExecutableCompiler {
+    fn target_kind(&self) -> TargetKind {
+        TargetKind::EXECUTABLE.clone()
+    }
+
+    fn compile(
+        &self,
+        unit: CairoCompilationUnit,
+        db: &mut RootDatabase,
+        ws: &Workspace<'_>,
+    ) -> Result<()> {
+        let target_dir = unit.target_dir(ws);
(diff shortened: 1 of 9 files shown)


scarb/src/compiler/compilers/executable.rs
```diff
@@ -0,0 +1,44 @@
...
+    fn compile(
+        &self,
+        unit: CairoCompilationUnit,
+        db: &mut RootDatabase,
+        ws: &Workspace<'_>,
+    ) -> Result<()> {
+        let target_dir = unit.target_dir(ws);
+        let main_crate_ids = collect_main_crate_ids(&unit, db);
+        let compiler_config = build_compiler_config(db, &unit, &main_crate_ids, ws);
+        let _ = trace_span!("compile_executable").enter();
```
maciektr, line 26:
```suggestion
        let _guard = trace_span!("compile_executable").enter();
```
Would you mind changing it here and in other places? TIL `let _ =` might drop right away :D

maciektr: Unresolving, as it's still not changed :D `let _ = trace_span`

FroyaTheHen: sorry, my bad 🙈 
`let _guard = trace_span!("compile_executable").enter();` still raises `temporary value dropped while borrowed [E0716]`. 
How about 👇 
```
        let span = trace_span!("compile_executable");
        let executable = {
            let _guard = span.enter();
            Executable::new(
                cairo_lang_executable::compile::compile_executable_in_prepared_db(
                    db,
                    None,
                    main_crate_ids,
                    compiler_config.diagnostics_reporter,
                )?,
            )
        };
```
??
Ofc, I will change other places as well, just want to consult it first with you :)

Draggu: https://docs.rs/tracing/latest/tracing/struct.Span.html#method.in_scope

maciektr: ^ Both are fine imho. `in_scope` is only unavoidable when working with asyncs.

scarb/src/compiler/compilers/lib.rs
```diff
@@ -114,8 +116,9 @@ impl Compiler for LibCompiler {
                 .context("failed calculating Sierra variables")?
             };
 
+            let span = trace_span!("compile_casm");
             let cairo_program = {
-                let _ = trace_span!("compile_casm").enter();
+                let _ = span.enter();
```
maciektr, line 121:
```suggestion
                let _guard = span.enter();
```

maciektr (approved):

maciektr (approved):

PR #1842, opened 2024-12-11
Add include field to the manifest

### LINKED ISSUE #1770: Ensure procmacros can be manually aot-compiled
## Summary

With a naive, no-effort approach, we will have a chicken-and-egg problem here that will make manual building impossible. All because the tarball preparation directory is temporary to the scarb package command execution. To overcome this we have to make sure that it is possible to allow users to include the target/scarb/cairo-plugin directory into the tarball. Cargo has [include/exclude fields,](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) which seem like the most plausible solution we can port.

## What determines whether package contains prebuilt binaries?

Any Scarb package that contains a `target/scarb/cairo-plugin directory` in tarball is considered as one containing prebuilt binaries. This name was chosen because it nicely hides inside Scarb’s/Cargo’s standard build file directory, which is very likely to already be avoided in source repositories and thus shouldn’t conflict with package sources.
Within this directory, binaries for all platforms will follow the following naming pattern: `${PACKAGE_NAME}_v${PACKAGE_VERSION}_${TARGET}.${so|dylib|dll}`

## Platform support

Due to licensing and technical constraints (macOS SDK can only be run on macOS), packaging with AOT compilation will be legally possible only on the following platform:

* `aarch64-apple-darwin`

maciektr:
fixes #1770

---

**Stack**:
- #1846
- #1836
- #1842 ⬅


mkaput (approved):

scarb/tests/package.rs
```diff
@@ -1525,3 +1525,114 @@ fn package_with_publish_disabled() {
...
+        .arg("--no-metadata")
+        .current_dir(&pkg)
+        .assert()
+        .failure()
+        .stdout_matches(indoc! {r#"
+        [..] Packaging foo v1.0.0 [..]
+        error: file `[..]file.txt` is not part of `foo`
+
+        Caused by:
+            prefix not found
```
DelevoXDG, line 1636: Nit:
can we remove this? 🤔 It's more confusing than helpful

website/docs/registries/publishing.md
```diff
@@ -44,17 +46,35 @@ publish = false
...
+directory for being uploaded, where both `name` and `version` correspond to the values in `Scarb.toml`.
+
+### Files included in the package
+
+All files in the package directory are included in the resulting tarball, except for the following:
+
+- Files excluded with rules defined in any `.scarbignore`, `.gitignore` or `.ignore` files.
+- The `.git` directory.
+- Any subdirectories containing `Scarb.toml` file.
+- The `<package root>/target` directory.
```
DelevoXDG, line 75: nit:
`<package root>/target` is likely to be causing more problems so we can reorder these to simplify troubleshooting
```suggestion
- The `<package root>/target` directory.
- Any subdirectories containing `Scarb.toml` file.
- The `.git` directory.
```

scarb/src/core/publishing/manifest_normalization.rs
```diff
@@ -73,6 +73,7 @@ fn generate_package(pkg: &Package) -> Box<TomlPackage> {
             .clone()
             .map(|_| MaybeWorkspace::Defined((Utf8PathBuf::from(DEFAULT_README_FILE_NAME)).into())),
         repository: metadata.repository.clone().map(MaybeWorkspace::Defined),
+        include: None,
```
DelevoXDG, line 76: Why is it empty?

Shouldn’t packaging with the normalized manifest produce the same result as packaging with the original one, including the same additional files?


It also is normalized in Cargo: https://github.com/rust-lang/cargo/blob/351b421bed006071ecd7f79cc81335d879f2ef45/src/cargo/util/toml/mod.rs#L1558

DelevoXDG (changes requested):

scarb/tests/package.rs
```diff
@@ -1525,3 +1525,143 @@ fn package_with_publish_disabled() {
...
+            "other/some/dir/file.txt",
+            "other/file.txt",
+            "target/some/file.txt",
+            "target/file.txt",
+        ])
+        .file_eq("VERSION", "1")
+        .file_eq_path("Scarb.orig.toml", t.child("Scarb.toml"))
+        .file_eq_path("src/lib.cairo", t.child("src/lib.cairo"))
+        .file_eq_path("src/foo.cairo", t.child("src/foo.cairo"))
+        .file_matches_nl(
```
DelevoXDG, line 1589: I think we should check the contents of some (at least one) of the included files.

DelevoXDG (approved):
Left one note re. tests, looks good otherwise

scarb/tests/package.rs
maciektr: It does check contents: 
```
        snapbox::assert_eq_path(expected_path, self.read_file(path));
```
https://docs.rs/snapbox/0.4.17/snapbox/fn.assert_eq_path.html

PR #1846, opened 2024-12-13
Add allowed prebuilts list

maciektr:
commit-id:66715aa1

### DIFF
--- a/scarb/src/core/manifest/toml_manifest.rs
+++ b/scarb/src/core/manifest/toml_manifest.rs
@@ -357,6 +357,12 @@ impl DefaultForProfile for TomlProfile {
     }
 }
 
+#[derive(Debug, Default, Deserialize, Serialize, Clone)]
+#[serde(rename_all = "kebab-case")]
+pub struct TomlToolScarbMetadata {
+    pub allow_prebuilt_plugins: Option<Vec<String>>,
+}
+
 impl TomlManifest {
     pub fn read_from_path(path: &Utf8Path) -> Result<Self> {
         let contents = fs::read_to_string(path)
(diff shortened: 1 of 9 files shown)


mkaput (approved):

scarb/src/core/resolver.rs
```diff
@@ -72,6 +72,67 @@ impl Resolve {
...
+    }
+}
+
+#[derive(Debug, Default)]
+struct SubTreeFilter<T: Sized + Eq + Hash>(HashSet<T>);
+
+impl<T: Sized + Eq + Hash> SubTreeFilter<T> {
+    fn new(allowed: Vec<T>) -> Result<Self> {
+        Ok(Self(allowed.into_iter().collect()))
+    }
```
DelevoXDG, line 127: Why not return `Self` directly? 🤔 

Should also simplify `filter_subtrees`

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -357,6 +357,12 @@ impl DefaultForProfile for TomlProfile {
     }
 }
 
+#[derive(Debug, Default, Deserialize, Serialize, Clone)]
+#[serde(rename_all = "kebab-case")]
+pub struct TomlToolScarbMetadata {
+    pub allow_prebuilt_macros: Option<Vec<String>>,
```
DelevoXDG, line 363: Nit:

Proc macros use `[cairo_plugin]` target in Scarb manifest, so just for consistency it might be better to name this
```suggestion
    pub allow_prebuilt_plugins: Option<Vec<String>>,
```

scarb/src/core/resolver.rs
maciektr: Ah, right. Just a leftover. Thanks!

PR #1855, opened 2024-12-19
Add contract size limits warnings

### LINKED ISSUE #1774: Add contract size limits warnings
### Problem

Emit a warning when either of the following limits breaks: 
- max number of sierra program felts > 80*1024
- max number of casm program felts > 80*1024
- size of class > 4mb
- size of compiled class > 4mb

Source of current limits: https://docs.starknet.io/tools/limits-and-triggers/

### Proposed Solution

_No response_

### Notes

_No response_

DelevoXDG:
Closes #1774

### DIFF
--- a/scarb/src/compiler/compilers/starknet_contract/validations.rs
+++ b/scarb/src/compiler/compilers/starknet_contract/validations.rs
@@ -20,10 +20,6 @@ use std::iter::zip;
 use tracing::debug;
 
 const AUTO_WITHDRAW_GAS_FLAG: &str = "add_withdraw_gas";
-const MAX_SIERRA_PROGRAM_FELTS: usize = 81290;
-const MAX_CASM_PROGRAM_FELTS: usize = 81290;
-const MAX_CONTRACT_CLASS_BYTES: usize = 4089446;
-const MAX_COMPILED_CONTRACT_CLASS_BYTES: usize = 4089446;
 
 pub fn ensure_gas_enabled(db: &mut RootDatabase) -> anyhow::Result<()> {
     let flag = FlagId::new(db.as_files_group_mut(), AUTO_WITHDRAW_GAS_FLAG);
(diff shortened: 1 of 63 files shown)


scarb/src/compiler/compilers/starknet_contract/validations.rs
```diff
@@ -117,3 +122,43 @@ pub fn check_allowed_libfuncs(
 
     Ok(())
 }
+
+pub fn check_sierra_size_limits(classes: &[ContractClass], ws: &Workspace<'_>) {
+    for class in classes {
+        let sierra_felts = class.sierra_program.to_vec().len();
```
maciektr, line 128:
```suggestion
        let sierra_felts = class.sierra_program.len();
```
Isn't this a vec already? 🤔

scarb/src/compiler/compilers/starknet_contract/validations.rs
```diff
@@ -117,3 +122,43 @@ pub fn check_allowed_libfuncs(
...
+        if sierra_felts > MAX_SIERRA_PROGRAM_FELTS {
+            ws.config().ui().warn(formatdoc! {r#"
+                Sierra program exceeds maximum byte-code size on Starknet:
+                {MAX_SIERRA_PROGRAM_FELTS} felts allowed. Actual size: {sierra_felts} felts.
+            "#});
+        }
+
+        let class_size = serde_json::to_vec(class).unwrap().len();
```
maciektr, line 136: This is logically correct, but serializing and allocating this just to check size seems wastefull, as we need to serialize it again to write into the file. 
Maybe we can add some logic to artifacts writer (https://github.com/software-mansion/scarb/blob/main/scarb/src/compiler/compilers/starknet_contract/artifacts_writer.rs) and count it there? 
For instance, we could wrap the `BufWriter` used to write serialized struct to the file (https://github.com/software-mansion/scarb/blob/9f5bb47f52b8bbccb896b7a75482cd6a9493a567/scarb/src/compiler/helpers.rs#L109) with some struct that implements `std::io::Write` https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write, to act as a proxy -> save count of written bytes in some internal counter, otherwise just call `.write()` or `.flush()` of the `BufWriter` used over the file. Just thought about this, let me know if there are better approaches.

maciektr (changes requested):

maciektr (approved):

maciektr:
👏

maciektr (approved):

PR #1869, opened 2025-01-07
Use less partitions in Scarb CI

maciektr:
Spend more time in a single runner, but less time waiting for runners. 

With only 20 runners available in the org, few open Scarb PRs will quickly use all of them. While a single jobs will finish quickly, jobs from multiple PRs will start taking runners simultaneously, causing them to spend a lot of time waiting for available runners, while each PR has few jobs finished and few not yet started.


mkaput (approved):
Don't forget to change branch rulesets.

PR #1872, opened 2025-01-08
feat: executable command

FroyaTheHen:


extensions/scarb-cairo-execute/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+    let file_path = create_incremental_file_in_dir("target/cairo-execute", Some(".zip"))?;
+    runner.get_cairo_pie()?.write_zip_file(&file_path)?;
+
+    Ok(())
+}
+
+fn load_prebuilt_executable(
+    path: &Utf8PathBuf,
+    filename: String,
+) -> Result<Executable, anyhow::Error> {
```
maciektr, line 153:
```suggestion
) -> anyhow::Result<Executable> {
```

Please import `anyhow::Result` too, instead of adding prefix.

extensions/scarb-cairo-execute/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+        file_path.exists(),
+        formatdoc! {r#"
+            package has not been compiled, file does not exist: {filename}
+            help: run `scarb build` to compile the package
+        "#}
+    );
+    let file = std::fs::File::open(file_path)
+        .map_err(|e| anyhow::anyhow!("Failed to open file: {}", e))?;
+    serde_json::from_reader(file)
+        .map_err(|e| anyhow::anyhow!("Failed parsing prebuilt executable: {}", e))
```
maciektr, line 165:
```suggestion
    let file = std::fs::File::open(file_path)
        .with_context("failed to open executable program: {file_path}")?;
    serde_json::from_reader(file)
        .with_context("failed to deserialize executable program: {file_path}")
```

extensions/scarb-cairo-execute/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+
+        let air_private_file_path = create_incremental_file_in_dir(
+            &base_path,
+            Some(AIR_PRIVATE_FILE.1),
+            AIR_PRIVATE_FILE.0,
+        )?;
+        let absolute = |path_buf: PathBuf| {
+            path_buf
+                .as_path()
+                .canonicalize()
```
FroyaTheHen, line 239: `cargo clippy` warns agains using `canonicalize()`, should i use `fsx::canonicalize_utf8` does it apply in this case? please advise

maciektr: Yes, you should. `fsx` is basically our wrapper over path manipulation functions that handles some edge-cases that we have run into in the past. The clippy lint you see has been added by us here https://github.com/software-mansion/scarb/blob/f1e8abb65ee21efb2c813ff95b8d2ecd440d940c/clippy.toml#L1

maciektr:
Replaced with https://github.com/software-mansion/scarb/pull/1885

PR #1873, opened 2025-01-08
Moved cli frontend of the cairo linter to scarb command

wawel37:


scarb/src/ops/lint.rs
```diff
@@ -0,0 +1,246 @@
...
+                    .ui()
+                    .print(Status::new("Skipping proc macro", &compilation_unit.name())),
+                CompilationUnit::Cairo(compilation_unit) => {
+                    ws.config()
+                        .ui()
+                        .print(Status::new("Checking", &compilation_unit.name()));
+
+                    let additional_plugins = if opts.test {
+                        vec![
+                            test_plugin_suite(),
```
maciektr, line 100: This plugin conflicts with Starknet Foundry one. It is also explicitly declared by the user in deps. I think it should not be injected here.

scarb/src/ops/lint.rs
```diff
@@ -0,0 +1,246 @@
...
+                            cairo_lint_plugin_suite(),
+                            starknet_plugin_suite(),
```
maciektr, line 102: It's declared by the user in project deps.

Cargo.toml
```diff
@@ -71,6 +72,7 @@ cairo-lang-test-plugin = "*"
 cairo-lang-test-runner = "*"
 cairo-lang-utils = { version = "*", features = ["env_logger"] }
 cairo-language-server = "*"
+cairo-lint-core = {git = "https://github.com/software-mansion/cairo-lint", rev = "3cd07f36a6d509c8844284dd88a17535b6b1e265"}
```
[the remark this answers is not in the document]
mkaput: same story as CairoLS

scarb/src/bin/scarb/args.rs
```diff
@@ -531,6 +533,27 @@ pub struct PublishArgs {
     pub ignore_cairo_version: bool,
 }
 
+#[derive(Parser, Clone, Debug)]
+pub struct LintArgs {
+    /// Name of the package.
+    #[command(flatten)]
+    pub packages_filter: PackagesFilter,
+    /// Path to the file or project to analyze
+    pub path: Option<String>,
```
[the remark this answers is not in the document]
wawel37: We will support it in the future, but u are right. Will remove it for now

wawel37:
FYI, will change the `cairo-toolchain-xtasks` version after we merge [this](https://github.com/software-mansion-labs/cairo-toolchain-xtasks/pull/9)

Cargo.toml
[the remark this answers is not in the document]
wawel37: Made it upgradable with the use of the xtask upgrade

scarb/src/ops/lint.rs
```diff
@@ -0,0 +1,237 @@
...
+                        compilation_unit.components().iter().any(|component| {
+                            component.target_kind() == TargetKind::TEST
+                                && component.package.id == package.id
+                        });
+                    let is_integration_test_package =
+                        compilation_unit.main_package_id().name.to_string()
+                            == format!("{}_integrationtest", package.id.name)
+                            && compilation_unit.main_package_id().version == package.id.version;
+                    is_main_component || has_test_components || is_integration_test_package
+                })
```
maciektr, line 76: I think this should work in a slightly different way. 

1. Compilation units should be deduplicated per package, as there is no need to lint both `lib` and `starknet-contract` targets and each of them will get own CU. 
2. To handle tests: 

For unit tests (i.e. those defined in module tree under `src` dir), whether tests will be checked or not is decided purely by cfg attrs. If you want to run checks of those, choose package CU with test target kind. When not, choose lib/starknet-contract/any available. 

For integration tests (i.e. those in `tests` dir), to find integration tests CU of the package you should: 
- Take package you want to lint
- Find target of this package with target kind `test` and `test-type` param set to `integration`
- Take name of this target from previous step 
- Use this name and package id to construct new package id with https://github.com/software-mansion/scarb/blob/fefeab6f9ec075ad6e9c8a568ebf39af7d6d299a/scarb/src/core/package/id.rs#L37
- Find CU with main package id equal to the one from previous step 

We cannot rely on `format!("{}_integrationtest"` though, while this happens to be the default name we add in auto-detected test targets, it can also be changed in Scarb manifest file (by defining the test target manually). 

3. Note, that if package relies on procedural macros, we need to compile them before linting the package. This means we need to take compilation units of those proc macro packages and run `ops::compile_unit` on them. See how `scarb check` does this: https://github.com/software-mansion/scarb/blob/fefeab6f9ec075ad6e9c8a568ebf39af7d6d299a/scarb/src/ops/compile.rs#L242-L274

scarb/src/ops/lint.rs
```diff
@@ -0,0 +1,237 @@
...
+                .collect::<Vec<_>>()
+        } else {
+            vec![compilation_units
+                .iter()
+                .find(|compilation_unit| compilation_unit.main_package_id() == package.id)
```
maciektr, line 81: Note this would also check package units tests - see my previous comment.

scarb/src/ops/lint.rs
```diff
@@ -0,0 +1,237 @@
...
+                .unwrap()]
+        };
+
+        for compilation_unit in package_compilation_units {
+            match compilation_unit {
+                // We skip proc macros as we don't want to check anything related to rust code.
+                CompilationUnit::ProcMacro(_) => ws
+                    .config()
+                    .ui()
+                    .print(Status::new("Skipping proc macro", &compilation_unit.name())),
```
maciektr, line 91:
```suggestion
                    .print(Status::new("Skipping Cairo plugin package", &compilation_unit.name())),
```

scarb/src/ops/lint.rs
```diff
@@ -0,0 +1,237 @@
...
+
+                    let additional_plugins = vec![cairo_lint_plugin_suite()];
+                    let ScarbDatabase { db, .. } =
+                        build_scarb_root_database(compilation_unit, ws, additional_plugins)?;
+
+                    let main_component = compilation_unit
+                        .components
+                        .iter()
+                        .find(|component| component.package.id == compilation_unit.main_package_id)
+                        .expect("main component is guaranteed to exist in compilation unit");
```
maciektr, line 105: Please use instead
https://github.com/software-mansion/scarb/blob/fefeab6f9ec075ad6e9c8a568ebf39af7d6d299a/scarb/src/compiler/compilation_unit.rs#L141

scarb/src/ops/lint.rs
```diff
@@ -0,0 +1,237 @@
...
+                        discriminator: main_component.id.to_discriminator(),
+                    });
+                    let mut diags = Vec::new();
+
+                    for module_id in &*db.crate_modules(crate_id) {
+                        if let Maybe::Ok(module_diags) = db.module_semantic_diagnostics(*module_id)
+                        {
+                            diags.push(module_diags);
+                        }
+                    }
```
maciektr, line 118:
```suggestion    
                    let diags = db
                        .crate_modules(crate_id)
                        .iter()
                        .flat_map(|module_id| db.module_semantic_diagnostics(*module_id).ok())
                        .collect_vec();
```

scarb/src/ops/lint.rs
```diff
@@ -0,0 +1,237 @@
...
+
+                    let should_lint_panics = package
+                        .tool_metadata("cairo-lint")
+                        .and_then(|config| config["nopanic"].as_bool())
+                        .unwrap_or(false);
```
maciektr, line 123: Please add a struct describing the config and deserialize into it.

scarb/src/ops/compile.rs
```diff
@@ -219,7 +219,7 @@ fn compile_unit_inner(unit: CompilationUnit, ws: &Workspace<'_>) -> Result<()> {
             let ScarbDatabase {
                 mut db,
                 proc_macro_host,
-            } = build_scarb_root_database(&unit, ws)?;
+            } = build_scarb_root_database(&unit, ws, vec![])?;
```
mkaput, line 222:
```suggestion
            } = build_scarb_root_database(&unit, ws, Default::default())?;
```

I know your code was shorter, but using `default` does better when refactoring

scarb/src/ops/expand.rs
```diff
@@ -168,7 +168,7 @@ fn do_expand(
     opts: ExpandOpts,
     ws: &Workspace<'_>,
 ) -> Result<()> {
-    let ScarbDatabase { db, .. } = build_scarb_root_database(compilation_unit, ws)?;
+    let ScarbDatabase { db, .. } = build_scarb_root_database(compilation_unit, ws, vec![])?;
```
mkaput, line 171:
```suggestion
    let ScarbDatabase { db, .. } = build_scarb_root_database(compilation_unit, ws, Default::default())?;
```

scarb/src/ops/lint.rs
mkaput: ah yeah, makes sense. didn't catch that

scarb/tests/lint.rs
```diff
@@ -0,0 +1,227 @@
...
+        .unwrap();
+
+    let snapbox = Scarb::quick_snapbox()
+        .arg("lint")
+        .arg("-t")
+        .current_dir(&t)
+        .assert()
+        .success();
+
+    // Current expected values include ANSI color codes because lint has custom renderer.
```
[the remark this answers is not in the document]
wawel37: Haven't noticed this feature, actually really helpful. As @mkaput said, we will cover it in a separate PR. https://github.com/software-mansion/cairo-lint/issues/156

scarb/src/bin/scarb/args.rs
```diff
@@ -531,6 +533,25 @@ pub struct PublishArgs {
...
 
+#[derive(Parser, Clone, Debug)]
+pub struct LintArgs {
+    /// Name of the package.
+    #[command(flatten)]
+    pub packages_filter: PackagesFilter,
+
+    /// Should lint the tests.
+    #[arg(short, long, default_value_t = false)]
+    pub test: bool,
```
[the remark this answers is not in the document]
wawel37: Hmmm, not so sure about that tbh. I checked and `cargo clippy` has a `--tests` args, and it inherently uses the flag from the `cargo check`, which _Check all targets that have `test = true` set_. https://github.com/rust-lang/cargo/blob/eb69b444a3528663db73de6fcbae2aa8146b01ac/src/bin/cargo/commands/check.rs#L25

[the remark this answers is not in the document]
mkaput: ok

wawel37:
Moved the fixing logic back to cairo-lint-core. https://github.com/software-mansion/cairo-lint/pull/179. As soon as it's merged, will upgrade the deps here in Scarb

maciektr:
@wawel37 Maybe we could calculate the change in `cairo-lint-core` and write to file here in Scarb? Wdyt?

wawel37:
> @wawel37 Maybe we could calculate the change in `cairo-lint-core` and write to file here in Scarb? Wdyt?

TBH I've thought about it. I think both of the solutions are good. I mean, if we put the logic of mutating the files back in the Scarb, the responsibility of fixing will be separated to 2 different places but on the other hand everything that mutates files will stay within Scarb logic. 

tl;dr:
I don't have a strong opinion on that

mkaput:
decoupling fix computation from application will be very handy for CairoLS integration, because we need to send fix patches to the langage server as text messages there

scarb/Cargo.toml
```diff
@@ -116,6 +117,7 @@ similar-asserts.workspace = true
 snapbox.workspace = true
 test-case.workspace = true
 test-for-each-example = { path = "../utils/test-for-each-example" }
+cairo-lint-core = { workspace = true, features = ["testing-colors"]}
```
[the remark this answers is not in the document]
wawel37: The version with feature is already underneath the `dev-dependencies`

mkaput (approved):

mkaput (approved):

scarb/src/ops/lint.rs
wawel37: Filtered out CU that has `"target": "test"` in cfgset 👍

scarb/src/ops/lint.rs
wawel37: Done 👍

maciektr (approved):
💪

PR #1888, opened 2025-01-16
Scarb execute name change, arguments file, error msgs

maciektr:
- **Rename scarb-cairo-execute to scarb-execute**
- **Support reading arguments from file in scarb execute**
- **Unify error messages in scarb execute**


extensions/scarb-execute/src/main.rs
```diff
@@ -203,9 +228,9 @@ fn main_inner(args: Args, ui: Ui) -> Result<(), anyhow::Error> {
...
 
     if args.run.print_program_output {
-        let mut output_buffer = "Program Output:\n".to_string();
+        let mut output_buffer = "Program output:\n".to_string();
         runner.vm.write_output(&mut output_buffer)?;
-        print!("{output_buffer}");
+        ui.print(output_buffer.trim_end());
     }
 
     let output_dir = scarb_target_dir.join("scarb-execute");
```
DelevoXDG, line 236: Should we keep the output in `scarb-execute`, not just `execute`? 🤔 

We typically don't add `scarb-` prefix to target outputs e.g. for builds or packages?

extensions/scarb-execute/src/main.rs
```diff
@@ -275,14 +300,14 @@ fn load_prebuilt_executable(path: &Utf8Path, filename: String) -> Result<Executa
     ensure!(
         file_path.exists(),
         formatdoc! {r#"
-            package has not been compiled, file does not exist: {filename}
+            package has not been compiled, file does not exist: `{filename}`
```
DelevoXDG, line 303: We should also change other uses e.g. https://github.com/software-mansion/scarb/blob/5e2567bc49f16afc9ea95cd2b625780167604a7c/extensions/scarb-cairo-run/src/main.rs#L114

Although I'm not sure this change is really necessary.

extensions/scarb-execute/src/main.rs
```diff
@@ -275,14 +300,14 @@ fn load_prebuilt_executable(path: &Utf8Path, filename: String) -> Result<Executa
...
             help: run `scarb build` to compile the package
         "#}
     );
     let file = fs::File::open(&file_path)
-        .with_context(|| format!("failed to open executable program: {file_path}"))?;
+        .with_context(|| format!("failed to open executable program: `{file_path}`"))?;
```
DelevoXDG, line 308: Same as above, I think we should decide on one approach use it across the codebase.

extensions/scarb-execute/src/main.rs
maciektr: I'd argue we do use ` way more often that we do not, and cairo-run is rather an outlier than a norm. Imho we should use it.

DelevoXDG: Actually, it's hard for me to find us using backticks for *paths* specifically. The opposite cases are quite easy to find:
- https://github.com/software-mansion/scarb/blob/5e2567bc49f16afc9ea95cd2b625780167604a7c/scarb/src/manifest_editor/mod.rs#L48
- https://github.com/software-mansion/scarb/blob/5e2567bc49f16afc9ea95cd2b625780167604a7c/scarb/src/ops/package.rs#L461
- https://github.com/software-mansion/scarb/blob/5e2567bc49f16afc9ea95cd2b625780167604a7c/scarb/src/ops/workspace.rs#L90
- https://github.com/software-mansion/scarb/blob/5e2567bc49f16afc9ea95cd2b625780167604a7c/scarb/src/flock.rs#L476

Not saying this is something to be addressed as a part of this PR, but since the intention was to unify error messages, I think the same should be done in all cases.

DelevoXDG (approved):

extensions/scarb-execute/src/main.rs
maciektr: Interesting observation, tbh I never attached too much weight to this. What I meant by unify here, was just so they start with lowercase and do not end with comma. We might do a more widespread "unification" in Scarb generally, defining some guidelines first.

PR #1890, opened 2025-01-16
Add plugin discovery utilities to scarb-metadata

integraledelebesgue:
**Stack**:
- #1921
- #1913
- #1970
- #1920
- #1919
- #1890 ⬅

### DIFF
--- a/scarb-metadata/src/lib.rs
+++ b/scarb-metadata/src/lib.rs
@@ -537,6 +537,21 @@ impl Metadata {
         self.packages.iter().find(|p| p.id == *id)
     }
 
+    /// Checks if the plugin associated with [`CompilationUnitCairoPluginMetadata`] is built-in into Scarb.
+    pub fn is_builtin_plugin(&self, plugin: &CompilationUnitCairoPluginMetadata) -> Option<bool> {
+        self.get_package(&plugin.package)?
+            .targets
+            .iter()
+            .find(|&target| target.kind == "cairo-plugin")
+            .map(|target| {
+                target
+                    .params
+                    .get("builtin")
+                    .and_then(|value| value.as_bool())
+                    .unwrap_or_default()
+            })
+    }
+
     /// Returns reference to [`CompilationUnitMetadata`] corresponding to the [`CompilationUnitId`].
     pub fn get_compilation_unit(&self, id: &CompilationUnitId) -> Option<&CompilationUnitMetadata> {
         self.compilation_units.iter().find(|p| p.id == *id)


scarb-metadata/src/lib.rs
```diff
@@ -532,6 +532,68 @@ impl Metadata {
         self.packages.iter().find(|p| p.id == *id)
     }
 
+    /// Returns reference to the [`CompilationUnitCairoPluginMetadata`] of all plugins that the package depends on.
+    pub fn get_package_plugin_dependencies(
+        &self,
+        id: &PackageId,
+    ) -> Option<Vec<&CompilationUnitCairoPluginMetadata>> {
```
piotmag769, line 539: You are mixing package dependencies with component dependencies here. Why would you need this function?

scarb-metadata/src/lib.rs
```diff
@@ -532,6 +532,68 @@ impl Metadata {
...
+        self.compilation_units
+            .iter()
+            .flat_map(|unit| unit.cairo_plugins.iter())
+            .find(|plugin| {
+                plugin
+                    .discriminator
+                    .as_ref()
+                    .is_some_and(|disc| disc == discriminator)
+            })
+    }
```
piotmag769, line 585: This is useless imo, API user will look the plugin metadata within the CU. Also there is no guarantee that the plugin id is unique in all metadata, only in a specific CU

piotmag769 (changes requested):

scarb-metadata/src/lib.rs
integraledelebesgue: It's handy for LS, I use it [here](https://github.com/software-mansion/cairols/blob/67150c98621d1988bc57c76cd8f0dbb6cd7282e6/src/project/scarb.rs#L187). However, I can rework or delete this function.

integraledelebesgue: Now I think it's actually useless ;v I will collect the plugin metadata in LS

piotmag769 (approved):

mkaput (approved):

scarb-metadata/src/lib.rs
```diff
@@ -534,6 +534,21 @@ impl Metadata {
         self.packages.iter().find(|p| p.id == *id)
     }
 
+    /// Checks if the plugin associated with [`CompilationUnitCairoPluginMetadata`] is built-in into the compiler.
+    pub fn is_builtin(&self, plugin: &CompilationUnitCairoPluginMetadata) -> Option<bool> {
```
maciektr, line 538:
```suggestion
    pub fn is_builtin_plugin(&self, plugin: &CompilationUnitCairoPluginMetadata) -> Option<bool> {
```
nit: `Metadata::is_builtin` would suggest to me, that metadata itself can be builtin or not

scarb-metadata/src/lib.rs
```diff
@@ -534,6 +534,21 @@ impl Metadata {
         self.packages.iter().find(|p| p.id == *id)
     }
 
+    /// Checks if the plugin associated with [`CompilationUnitCairoPluginMetadata`] is built-in into the compiler.
```
maciektr, line 537:
```suggestion
    /// Checks if the plugin associated with [`CompilationUnitCairoPluginMetadata`] is built-in into Scarb.
```
nit

maciektr (approved):

maciektr (approved):

maciektr (approved):

PR #1904, opened 2025-01-20
Disable gas cfg when `enable-gas=false`

### LINKED ISSUE #1901: Disable gas cfg in compilation units when `enable-gas=false`
`enable-gas=false` is not sufficient, we need project compiled with the gas disabled cfg, without the user specifying anything else

DelevoXDG:
Closes #1901


piotmag769 (approved):

piotmag769 (commented):
Add docs pls

DelevoXDG:
> Add docs pls

#1905

PR #1916, opened 2025-01-24
[test] scarb-prove win

DelevoXDG:


DelevoXDG:
Determined that win error is on the stwo prover side

PR #1913, opened 2025-01-22
Introduce per-crate plugins to the compilation model

integraledelebesgue:
**Stack**:
- #1921
- #1913 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #1919, opened 2025-01-27
Introduce procedural macro repository

integraledelebesgue:
**Stack**:
- #1921
- #1913
- #1970
- #1920
- #1919 ⬅

### DIFF
--- a/scarb/src/compiler/plugin/proc_macro/host.rs
+++ b/scarb/src/compiler/plugin/proc_macro/host.rs
@@ -3,7 +3,7 @@ use crate::compiler::plugin::proc_macro::{
     Expansion, ExpansionKind, FromSyntaxNode, ProcMacroInstance,
 };
 use crate::core::{Config, Package, PackageId};
-use anyhow::{ensure, Context, Result};
+use anyhow::{bail, ensure, Context, Result};
 use cairo_lang_defs::ids::{ModuleItemId, TopLevelLanguageElementId};
 use cairo_lang_defs::patcher::{PatchBuilder, RewriteNode};
 use cairo_lang_defs::plugin::{
@@ -1170,5 +1170,43 @@ impl ProcMacroHost {
 
     pub fn macros(&self) -> &[Arc<ProcMacroInstance>] {
         &self.macros
+    }
+}
+
+/// A global storage for dynamically-loaded procedural macros.
+/// Loads dynamic shared libraries and hides them beside [`ProcMacroInstance`].
+/// Guarantees that every library is loaded exactly once,
+/// but does not prevent loading multiple versions of the same library.
+#[derive(Default)]
+pub struct ProcMacroRepository {
+    /// A mapping between the [`PackageId`] of the package which defines the plugin
+    /// and the [`ProcMacroInstance`] holding the underlying shared library.
+    macros: RwLock<HashMap<PackageId, Arc<ProcMacroInstance>>>,
+}
+
+impl ProcMacroRepository {
+    /// Returns the [`ProcMacroInstance`] representing the procedural macros defined in the [`Package`].
+    /// Loads the underlying shared library if it has not been loaded yet.
+    pub fn get_or_load(&self, package: Package, config: &Config) -> Result<Arc<ProcMacroInstance>> {
+        let Ok(macros) = self.macros.read() else {
+            bail!("could not get a read access to the ProcMacroRepository");
+        };
+
+        if let Some(instance) = macros.get(&package.id) {
+            return Ok(instance.clone());
+        }
+
+        let Ok(mut macros) = self.macros.write() else {
+            bail!("could not get a write access to the ProcMacroRepository");
+        };
+
+        let lib_path = package
+            .shared_lib_path(config)
+            .context("could not resolve shared library path")?;
+
+        let instance = Arc::new(ProcMacroInstance::try_new(package.id, lib_path)?);
+        macros.insert(package.id, instance.clone());
+
+        Ok(instance)
     }
 }


scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -1172,3 +1173,66 @@ impl ProcMacroHost {
...
+        instance: Arc<ProcMacroInstance>,
+    ) -> Result<()> {
+        let plugin = Arc::new(ProcMacroHostPlugin::try_new(vec![instance])?);
+        self.macros.insert(package_id, plugin);
+        Ok(())
+    }
+
+    /// Returns the [`ProcMacroHostPlugin`] representing the procedural macros defined in the [`Package`].
+    /// Loads the underlying shared library if it has not been loaded yet.
+    pub fn get_or_load(
```
integraledelebesgue, line 1203: I will probably change this to something like `try_load` or `load_if_not_present`. The return value is not used anywhere.

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -1172,3 +1173,66 @@ impl ProcMacroHost {
...
+
+    /// Drops the [`ProcMacroRepository`], invoking the necessary
+    /// [`ProcMacroHostPlugin::post_process`] on each stored proc macro.
+    pub fn safe_drop(self, db: &dyn SemanticGroup) -> Result<()> {
+        for plugin in self.macros.values() {
+            plugin.post_process(db)?;
+        }
+
+        Ok(())
+    }
```
piotmag769, line 1237: Don't we want this done via `Drop`?

piotmag769 (commented):
@maciektr could you TAL?

scarb/src/compiler/plugin/proc_macro/host.rs
mkaput: I don't think we can, because this result has to be handled (error printed to the user and execution halted).

Imho the name of this function isn't the best, could be just `post_process`

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -1172,3 +1173,66 @@ impl ProcMacroHost {
         &self.macros
     }
 }
+
+/// A global storage for dynamically-loaded procedural macros.
+/// Loads dynamic shared libraries and hides them beside [`ProcMacroHostPlugin`].
+/// Guarantees that every library is loaded exactly once,
+/// but does not prevent loading multiple versions of the same library.
+#[derive(Default)]
+pub struct ProcMacroRepository {
```
mkaput, line 1182: where is this used?

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -1172,3 +1173,66 @@ impl ProcMacroHost {
...
+
+impl ProcMacroRepository {
+    /// Inserts the pre-loaded [`ProcMacroInstance`] to the storage, without checking
+    /// whether some plugin for the [`PackageId`] already exists.
+    pub fn insert_prebuilt_unchecked(
+        &mut self,
+        package_id: PackageId,
+        instance: Arc<ProcMacroInstance>,
+    ) -> Result<()> {
+        let plugin = Arc::new(ProcMacroHostPlugin::try_new(vec![instance])?);
```
maciektr, line 1196: We will need some validations on proc macro expansion duplicates, right now it's handled in the ProcMacroHostPlugin with assumption that all proc macros are added to a single plugin. See https://github.com/software-mansion/scarb/blob/b3558737984e987c746fe1794a08ce0dd3ac9172/scarb/src/compiler/plugin/proc_macro/host.rs#L138

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -1172,3 +1173,66 @@ impl ProcMacroHost {
...
+
+/// A global storage for dynamically-loaded procedural macros.
+/// Loads dynamic shared libraries and hides them beside [`ProcMacroHostPlugin`].
+/// Guarantees that every library is loaded exactly once,
+/// but does not prevent loading multiple versions of the same library.
+#[derive(Default)]
+pub struct ProcMacroRepository {
+    /// A mapping between the [`PackageId`] of the package which defines the plugin
+    /// and the [`ProcMacroHostPlugin`] holding the underlying shared library.
+    macros: HashMap<PackageId, Arc<ProcMacroHostPlugin>>,
```
maciektr, line 1185: It's not extremely obvious for me, whether we should have 
```
proc macro expansion - many to 1 -> proc macro instance - 1 to 1 -> proc macro host plugin - many to 1 -> single package cairo plugin dependencies
```
or 
``` 
proc macro expansion - many to 1 -> proc macro instance - many to 1 -> proc macro host plugin - 1 to 1 -> single package cairo plugin dependencies
```
(or any other configuration)
Can you elaborate on the decision made here?

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -1172,3 +1173,66 @@ impl ProcMacroHost {
...
+        config: &Config,
+    ) -> Result<Arc<ProcMacroHostPlugin>> {
+        match self.macros.entry(package.id) {
+            Entry::Occupied(slot) => Ok(slot.get().clone()),
+            Entry::Vacant(slot) => {
+                let lib_path = package
+                    .shared_lib_path(config)
+                    .context("could not resolve shared library path")?;
+
+                let instance = Arc::new(ProcMacroInstance::try_new(package.id, lib_path)?);
```
maciektr, line 1215: Note: `ProcMacroInstance::try_new` is responsible for issuing `dlopen` call, so this will result in loading the same (compiled) proc macro dll multiple times.

scarb/src/compiler/plugin/proc_macro/host.rs
integraledelebesgue: I use it in the [plugin collection algorithm in the next PR](https://github.com/software-mansion/scarb/blob/fdd6dd02e8073e2ed39276a17ab77d7d9265d43b/scarb/src/compiler/db.rs#L59)

mkaput (approved):
nothing more to add from my side on top of current open discussions

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -1172,3 +1173,37 @@ impl ProcMacroHost {
...
+
+/// A global storage for dynamically-loaded procedural macros.
+/// Loads dynamic shared libraries and hides them beside [`ProcMacroInstance`].
+/// Guarantees that every library is loaded exactly once,
+/// but does not prevent loading multiple versions of the same library.
+#[derive(Default)]
+pub struct ProcMacroRepository {
+    /// A mapping between the [`PackageId`] of the package which defines the plugin
+    /// and the [`ProcMacroInstance`] holding the underlying shared library.
+    macros: OnceMap<PackageId, Arc<ProcMacroInstance>>,
```
maciektr, line 1185: Why not just `RwLock<HashMap<>>`? 🤔

integraledelebesgue: Done, reworked to `RwLock`. I haven't heard about it before 😅

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -1172,3 +1172,41 @@ impl ProcMacroHost {
...
+    /// and the [`ProcMacroInstance`] holding the underlying shared library.
+    macros: RwLock<HashMap<PackageId, Arc<ProcMacroInstance>>>,
+}
+
+impl ProcMacroRepository {
+    /// Returns the [`ProcMacroInstance`] representing the procedural macros defined in the [`Package`].
+    /// Loads the underlying shared library if it has not been loaded yet.
+    pub fn get_or_load(&self, package: Package, config: &Config) -> Result<Arc<ProcMacroInstance>> {
+        let Ok(macros) = self.macros.read() else {
+            bail!("Could get a read access to the ProcMacroRepository");
```
piotmag769, line 1192:
```suggestion
            bail!("Could not get a read access to the ProcMacroRepository");
```

piotmag769 (approved):

scarb/src/compiler/plugin/proc_macro/host.rs
maciektr: Also, start with lowercase please

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -1172,3 +1172,41 @@ impl ProcMacroHost {
...
+        };
+
+        if let Some(instance) = macros.get(&package.id) {
+            return Ok(instance.clone());
+        }
+
+        let Ok(mut macros) = self.macros.write() else {
+            bail!("Could not get a write access to the ProcMacroRepository");
```
maciektr, line 1200:
```suggestion
            bail!("could not get a write access to the ProcMacroRepository");
```
vide supra

maciektr (approved):

scarb/src/compiler/plugin/proc_macro/host.rs
integraledelebesgue: Sure, done

PR #1920, opened 2025-01-27
Implement per-crate plugin collection

integraledelebesgue:
**Stack**:
- #1921
- #1913
- #1970
- #1920 ⬅

### DIFF
--- a/scarb/src/compiler/plugin/collection.rs
+++ b/scarb/src/compiler/plugin/collection.rs
@@ -1,0 +1,185 @@
+use std::{collections::HashMap, sync::Arc};
+
+use anyhow::Result;
+use cairo_lang_semantic::{inline_macros::get_default_plugin_suite, plugin::PluginSuite};
+use itertools::Itertools;
+
+use crate::{
+    compiler::{CairoCompilationUnit, CompilationUnitComponentId, CompilationUnitDependency},
+    core::Workspace,
+};
+
+#[cfg(doc)]
+use crate::core::PackageId;
+
+use super::proc_macro::{ProcMacroHostPlugin, ProcMacroInstance};
+
+pub struct PluginsForComponents {
+    pub plugins: HashMap<CompilationUnitComponentId, PluginSuite>,
+    pub proc_macros: HashMap<CompilationUnitComponentId, Arc<ProcMacroHostPlugin>>,
+}
+
+impl PluginsForComponents {
+    /// Collects plugins, either built-in or procedural macros, for all components of the [`CairoCompilationUnit`].
+    pub fn collect(workspace: &Workspace<'_>, unit: &CairoCompilationUnit) -> Result<Self> {
+        let mut plugins = collect_builtin_plugins(workspace, unit)?;
+
+        let proc_macros = collect_proc_macros(workspace, unit)?
+            .into_iter()
+            .map(|(component_id, instances)| {
+                let plugin = Arc::new(ProcMacroHostPlugin::try_new(instances)?);
+                Ok((component_id, plugin))
+            })
+            .collect::<Result<HashMap<_, _>>>()?;
+
+        for (component_id, suite) in plugins.iter_mut() {
+            if let Some(proc_macro) = proc_macros.get(component_id) {
+                suite.add(ProcMacroHostPlugin::build_plugin_suite(proc_macro.clone()));
+            }
+        }
+
+        Ok(Self {
+            plugins,
+            proc_macros,
+        })
+    }
+}
+
+// NOTE: Since this structure is used to handle JsonRPC requests, its keys have to be serialized to strings.
+//
+/// A container for Proc Macro Server to manage macros present in the analyzed workspace.
+///
+/// # Invariant
+/// Correct usage of this struct during proc macro server <-> LS communication
+/// relies on the implicit contract that keys of `macros_for_packages` are of form
+/// `PackageId.to_serialized_string()` which is always equal to
+/// `scarb_metadata::CompilationUnitComponentId.repr`.
+pub struct WorkspaceProcMacros {
+    /// A mapping of the form: `package (as a serialized [`PackageId`]) -> plugin`.
+    /// Contains IDs of all packages from the workspace, each mapped to a [`ProcMacroHostPlugin`] which contains
+    /// **all proc macro dependencies of the package** collected from **all compilation units it appears in**.
+    pub macros_for_packages: HashMap<String, Arc<ProcMacroHostPlugin>>,
+}
+
+impl WorkspaceProcMacros {
+    /// Collects and groups procedural macros for all packages in the workspace.
+    pub fn collect(
+        workspace: &Workspace<'_>,
+        compilation_units: &[&CairoCompilationUnit],
+    ) -> Result<Self> {
+        let mut macros_for_components = HashMap::<_, Vec<_>>::new();
+
+        for &unit in compilation_units {
+            for (component_id, mut macro_instances) in collect_proc_macros(workspace, unit)? {
+                // Here we assume that CompilationUnitComponentId.repr == PackageId.to_serialized_string().
+                let key = component_id.package_id.to_serialized_string();
+                macros_for_components
+                    .entry(key)
+                    .or_default()
+                    .append(&mut macro_instances);
+            }
+        }
+
+        let macros_for_packages = macros_for_components
+            .into_iter()
+            .map(|(package_id, macro_instances)| {
+                let deduplicated_instances = macro_instances
+                    .into_iter()
+                    .unique_by(|instance| instance.package_id())
+                    .collect();
+
+                let plugin = Arc::new(ProcMacroHostPlugin::try_new(deduplicated_instances)?);
+
+                Ok((package_id, plugin))
+            })
+            .collect::<Result<HashMap<_, _>>>()?;
+
+        Ok(Self {
+            macros_for_packages,
+        })
+    }
+
+    /// Returns a [`ProcMacroHostPlugin`] assigned to the package with `package_id`.
+    pub fn get(&self, package_id: &str) -> Option<Arc<ProcMacroHostPlugin>> {
+        self.macros_for_packages.get(package_id).cloned()
+    }
+}
+
+/// Builds [`PluginSuite`]s for each component of the [`CairoCompilationUnit`],
+/// according to the dependencies on builtin macros.
+fn collect_builtin_plugins(
+    workspace: &Workspace<'_>,
+    unit: &CairoCompilationUnit,
+) -> Result<HashMap<CompilationUnitComponentId, PluginSuite>> {
+    let mut plugin_suites = HashMap::new();
+
+    for component in unit.components.iter() {
+        let mut component_suite = get_default_plugin_suite();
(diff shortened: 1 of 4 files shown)


scarb/src/compiler/db.rs
```diff
@@ -48,6 +53,98 @@ pub(crate) fn build_scarb_root_database(
...
+/// loads their shared libraries and stores them in the [`ProcMacroRepository`].
+#[allow(dead_code)] // Will be used in the next PR.
+fn collect_proc_macros(
+    unit: &CairoCompilationUnit,
+    ws: &Workspace<'_>,
+) -> Result<ProcMacroRepository> {
+    // TODO: Discuss
+    // Is it possible the same plugin is allowed
+    // to be pre-built in one manifest and not allowed in the other?
+    // Should we bother inserting the pre-builts without checking?
```
piotmag769, line 66: Good question. Btw @maciektr what happens if i.e. for the user's package A it is NOT allowed to be pre-built but for the dependency B of the package A it is allowed?

mkaput: Only prebuilt manifests of the main component should matter.

maciektr: It's possible that out of two workspace members, one allow using a prebuilt macro and one does not. Right now, in such case we would act accordingly, loading prebuilt for one, and building from sources for the second one. See https://github.com/software-mansion/scarb/blob/78619feba14523670ab1943a4feea8dc6568d0cb/scarb/src/ops/resolve.rs#L277

piotmag769 (commented):
Will review when @maciektr's comments are addressed

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,173 @@
...
+            macros_for_compilation_units,
+        })
+    }
+
+    /// Returns a `ProcMacroHostPlugin` assigned to the component with `cu_component_id`
+    /// in a compilation unit which member package has `cu_member_id`.
+    pub fn get(
+        &self,
+        compilation_unit_member_id: &str,
+        compilation_unit_component_id: &str,
```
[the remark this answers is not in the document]
integraledelebesgue: This function is used by the proc macro server, based on requests containing discriminator strings. Constructing e.g. `PackageId`s from them would be tricky

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,173 @@
...
+                                .to_discriminator()
+                                .unwrap_or_else(|| component_id.to_crate_identifier().into()) // Relevant only for corelib
+                                .to_string(),
+                            component_plugins,
+                        )
+                    })
+                    .collect();
+                Ok((unit_id, plugins))
+            })
+            .collect::<anyhow::Result<HashMap<_, _>>>()?;
```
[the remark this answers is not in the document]
integraledelebesgue: Done

scarb/src/compiler/db.rs
integraledelebesgue: In the current implementation, prebuilts take precedence over the source ones, please see https://github.com/software-mansion/scarb/blob/45d7a9e35cdbff3eb1edb34181f6cb59ca5aee7c/scarb/src/compiler/plugin/collection.rs#L153-L155
That's exactly the behaviour we agreed to stick with (@piotmag769, @maciektr).

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,167 @@
...
+    unit: &CairoCompilationUnit,
+) -> Result<HashMap<CompilationUnitComponentId, PluginSuite>> {
+    let mut plugin_suites = HashMap::new();
+
+    for component in unit.components.iter() {
+        let mut component_suite = get_default_plugin_suite();
+
+        for dependency in component.dependencies.iter() {
+            if matches!(dependency, CompilationUnitDependency::Library(_)) {
+                continue;
```
piotmag769, line 96: [nit] imo it's better to check if it doesn't match plugin dependency (in case we ever add more variants to the enum)

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            let Some(plugin) = unit
+                .cairo_plugins
+                .iter()
+                .find(|plugin| &plugin.component_dependency_id == dependency.component_id())
+            else {
+                continue;
```
piotmag769, line 104: In what case we can fall here? I don't think we ever should

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,167 @@
...
+    let proc_macro_repository = workspace.config().proc_macro_repository();
+    let mut proc_macros_for_components = HashMap::new();
+
+    for component in unit.components.iter() {
+        let mut component_proc_macro_instances = Vec::new();
+
+        for dependency in component.dependencies.iter() {
+            if matches!(dependency, CompilationUnitDependency::Library(_)) {
+                continue;
+            }
```
piotmag769, line 139: Same nit as above

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,167 @@
...
+
+            let Some(plugin) = unit
+                .cairo_plugins
+                .iter()
+                .find(|plugin| &plugin.component_dependency_id == dependency.component_id())
+            else {
+                continue;
```
piotmag769, line 146: Same question as above

piotmag769 (commented):
Looks good in general

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        Ok(Self {
+            plugins,
+            proc_macros,
+        })
+    }
+}
+
+// NOTE: Since this structure is used to handle JsonRPC requests, its keys have to be serialized to strings.
```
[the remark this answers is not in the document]
integraledelebesgue: I placed it here because it is closely related to the analogous collection logic used in the compilation and reuses some of its parts, namely `collect_proc_macros`. I can locate it in the PMS-specific module in the other PR if you think that's more appropriate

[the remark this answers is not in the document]
piotmag769: I haven't paid attention that it reuses this logic tbf, I guess it's fine then

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,185 @@
...
+                match macros_for_components.entry(key) {
+                    Entry::Vacant(entry) => {
+                        entry.insert(macro_instances);
+                    }
+                    Entry::Occupied(mut entry) => {
+                        entry.get_mut().append(&mut macro_instances);
+                    }
+                };
+            }
+        }
```
piotmag769, line 77:
```suggestion
macros_for_components.entry(key).or_default().append(&mut macro_instances);
```

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        let macros_for_components = macros_for_components
+            .into_iter()
+            .map(|(component_id, macro_instances)| {
+                let deduplicated_instances = macro_instances.into_iter().unique().collect();
```
piotmag769, line 82: Good catch!

piotmag769 (approved):

scarb/src/compiler/plugin/collection.rs
integraledelebesgue: Agree, fixed

scarb/src/compiler/plugin/collection.rs
integraledelebesgue: We cannot actually. I put an `expect` there

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// A mapping of the form: serialized CompilationUnitComponentId -> plugin.
+    pub macros_for_components: HashMap<String, Arc<ProcMacroHostPlugin>>,
+}
+
+impl WorkspaceProcMacros {
+    /// Collects and groups procedural macros for all components of all compilation units in the workspace.
+    pub fn collect(
+        workspace: &Workspace<'_>,
+        compilation_units: &[&CairoCompilationUnit],
+    ) -> Result<Self> {
```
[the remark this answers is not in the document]
integraledelebesgue: Done

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// NOTE: Since this structure is used to handle JsonRPC requests, its keys have to be serialized to strings.
+// NOTE: It relies on the implicit contract: CompilationUnitComponentId.repr == PackageId.to_serialized_string().
+//
+/// A container for Proc Macro Server to manage macros present in the analyzed workspace.
+pub struct WorkspaceProcMacros {
+    /// A mapping of the form: `package (as a serialized [`PackageId`]) -> plugin`.
+    /// Contains IDs of all packages from the workspace, each mapped to a [`ProcMacroHostPlugin`] which contains
+    /// **all macros that the package may depend on in all compilation units it appears in**.
```
piotmag769, line 55:
```suggestion
    /// **all proc macro dependencies of the package** collected from **all compilation units it appears in**.
```

scarb/src/compiler/plugin/collection.rs
```diff
@@ -0,0 +1,175 @@
...
+
+        Ok(Self {
+            plugins,
+            proc_macros,
+        })
+    }
+}
+
+// NOTE: Since this structure is used to handle JsonRPC requests, its keys have to be serialized to strings.
+// NOTE: It relies on the implicit contract: CompilationUnitComponentId.repr == PackageId.to_serialized_string().
```
piotmag769, line 49:
```suggestion
// INVARIANT: Correct usage of this struct during proc macro server <-> LS communication relies on the implicit contract that keys of `macros_for_packages` are of form `PackageId.to_serialized_string()` which is always equal to `scarb_metadata::CompilationUnitComponentId.repr`.
```

Could be in `///` doc too since it's rly important.

piotmag769 (approved):

scarb/src/compiler/plugin/collection.rs
integraledelebesgue: Done

scarb/src/compiler/plugin/collection.rs
integraledelebesgue: Done

scarb/src/compiler/plugin/proc_macro/ffi.rs
```diff
@@ -353,6 +354,20 @@ struct Plugin {
     vtable: VTableV0,
 }
 
+impl std::cmp::PartialEq for Plugin {
```
maciektr, line 357: Discussed offline, will be removed

integraledelebesgue: Done

maciektr (approved):

piotmag769 (approved):

PR #1930, opened 2025-01-29
Fix quote! parsing for puncts not followed by ident & emit a whitespace preceding a quote var 

maciektr:


plugins/cairo-lang-quote/src/lib.rs
```diff
@@ -1,12 +1,12 @@
 use std::iter::Peekable;
 
-use proc_macro::{Delimiter, TokenStream as RustTokenStream, TokenTree as RustTokenTree};
-use proc_macro2::{Ident, Span};
+use proc_macro2::{Delimiter, Ident, Span, TokenTree};
```
maciektr, line 3: We use `quote!` anyway, which works on `proc_macro2`, using it throughout the code allows us for easier testing, as `proc_macro2` can be used wherever, while `proc_macro` only in a procedural macro code.

plugins/cairo-lang-quote/src/lib.rs
```diff
@@ -108,8 +119,84 @@ pub fn quote(input: RustTokenStream) -> RustTokenStream {
...
+                QuoteToken::Content("(".to_string()),
+                QuoteToken::Content(")".to_string()),
+                QuoteToken::Content("{".to_string()),
+                QuoteToken::Content("}".to_string()),
+            ]
+        );
+    }
+
+    #[test]
+    fn quote_var_whitespace() {
```
maciektr, line 163: Before this fix, `mod #name {` got translated to ` modname {` 🙃

wawel37: That's on me, my bad

wawel37 (approved):
LGTM 💪

PR #1940, opened 2025-01-30
Scarb execute: Support printing resource usage, exit with non-zero code on panic

maciektr:


extensions/scarb-execute/src/output.rs
```diff
@@ -0,0 +1,196 @@
...
+#[derive(Serialize)]
+pub struct ExecutionResources {
+    n_steps: usize,
+    n_memory_holes: usize,
+    builtin_instance_counter: Vec<(String, usize)>,
+    syscalls: Vec<(String, usize)>,
+}
+
+impl ExecutionResources {
+    pub fn try_new(runner: &CairoRunner, hint_processor: CairoHintProcessor) -> Result<Self> {
```
DelevoXDG, line 73: Why not take `CairoHintProcessor` by reference?

extensions/scarb-execute/src/output.rs
```diff
@@ -0,0 +1,196 @@
...
+
+impl From<PanicReason> for Result<()> {
+    fn from(panic_reason: PanicReason) -> Self {
+        if let Some(reason) = panic_reason.0 {
+            Err(anyhow::anyhow!(reason))
+        } else {
+            Ok(())
+        }
+    }
+}
```
DelevoXDG, line 196: I get the idea, but IMO it’s extremely confusing that an `Ok` result actually indicates a panic. It seems counterintuitive to have success and error inverted, and converting panic reason into a result.

Instead, I wonder if we could instead define something like:
```rs
pub enum ExecutionOutcome {
    Success,
    Panic { reason: String },
}

impl ExecutionOutcome {
    pub fn try_new(runner: &CairoRunner, hint_processor: &CairoHintProcessor) -> Result<Self> {
        if let [.., start_marker, end_marker] = &hint_processor.markers[..] {
            ...
            Ok(ExecutionOutcome::Panic { reason })
        } else {
            Ok(ExecutionOutcome::Success)
        }
    }
}
```

I feel like this would make handling of panic reason more straightforward.

extensions/scarb-execute/src/output.rs
maciektr: We could, but we would need to clone the StarknetExecutionResources for this: https://github.com/software-mansion/scarb/blob/382168ba86efb70306831c21394a8fbf164324d6/extensions/scarb-execute/src/output.rs#L79 and we don't need the hint processor later anyway, so we can avoid this clone by consuming it right away.

extensions/scarb-execute/src/lib.rs
```diff
@@ -128,22 +130,19 @@ pub fn execute(
...
-                format!("panic data markers mismatch: start={start_marker}, end={end_marker}")
-            })?;
-            let panic_data = runner
-                .vm
-                .get_integer_range(*start_marker, size)
-                .with_context(|| "failed reading panic data")?;
-            ui.print(format_for_panic(panic_data.into_iter().map(|value| *value)));
-        }
-    }
+    let panic_reason = PanicReason::try_new(&runner, &hint_processor);
```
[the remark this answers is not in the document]
maciektr: Yes, I don't see a reason to not do this 🤔 Do you?

extensions/scarb-execute/src/output.rs
maciektr: > an Ok result actually indicates a panic

Hmm, does it? If there is a panic, this function will convert to an anyhow::Error 🤔 

> something like

Yep, enum is better than an option! Thanks, I'll change that!

extensions/scarb-execute/src/output.rs
```diff
@@ -0,0 +1,201 @@
...
+
+impl From<ExecutionStatus> for Result<()> {
+    fn from(panic_reason: ExecutionStatus) -> Self {
+        if let ExecutionStatus::Panic { reason } = panic_reason {
+            Err(anyhow::anyhow!(reason))
+        } else {
+            Ok(())
+        }
+    }
+}
```
DelevoXDG, line 201:
```suggestion
impl From<ExecutionStatus> for Result<()> {
    fn from(status: ExecutionStatus) -> Self {
        if let ExecutionStatus::Panic { reason } = status {
            Err(anyhow::anyhow!(reason))
        } else {
            Ok(())
        }
    }
}
```

extensions/scarb-execute/src/lib.rs
```diff
@@ -195,6 +194,8 @@ pub fn execute(
         fs::write(air_private_input_path, output_value)?;
     }
 
+    execution_status?.into_result()?;
```
DelevoXDG, line 197: Nit:

I would still prefer checking directly
```suggestion
    if let ExecutionStatus::Panic { reason } = execution_status? {
        return Err(anyhow::anyhow!(reason));
    }
```

maciektr: If it's more readable to you, sure

PR #1941, opened 2025-01-30
Executable target: allow syscalls flag

maciektr:
Needs https://github.com/starkware-libs/cairo/pull/7189


scarb/tests/metadata.rs
DelevoXDG, line 1: Nit:
Should we maybe add test that checks that default behaviour (`syscalls_allowed = false`) leads to error upon attempt to execute?

DelevoXDG (approved):

scarb/tests/metadata.rs
maciektr: We could, but I am not sure if there is a reliable syscall we can test against - all I've seen (including ones from the PR) just change the validation error to a panic later on

PR #1951, opened 2025-02-04
Allow choosing executable 

maciektr:


extensions/scarb-execute/src/lib.rs
```diff
@@ -205,6 +211,54 @@ pub fn execute(
...
+fn find_build_target<'a>(
+    package: &'a PackageMetadata,
+    executable_name: Option<&str>,
+    executable_function: Option<&str>,
+) -> Result<&'a TargetMetadata> {
+    package
+        .targets
+        .iter()
+        .filter(|target| target.kind.as_str() == "executable")
+        .find(|target| {
```
DelevoXDG, line 223: This will return the first matched target though, even if multiple matching targets are present. 

I'm not sure that's desired behaviour 🤔

DelevoXDG: Would *probably* be solved if `--executable-name` and `--executable--function` were mutually exclusive

extensions/scarb-execute/tests/build.rs
DelevoXDG, line 1: Should we check the output files as in the rest of the tests? 🤔

extensions/scarb-execute/src/lib.rs
```diff
@@ -205,6 +211,54 @@ pub fn execute(
...
+                .unwrap_or_default();
+
+            executable_name_match || executable_function_match
+        })
+        .map(Ok)
+        .unwrap_or_else(|| {
+            let mut all = package
+                .targets
+                .iter()
+                .filter(|target| target.kind.as_str() == "executable");
```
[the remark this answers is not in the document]
maciektr: Split into more vars instead of chain, let me know if it helps.

scarb/src/compiler/compilers/executable.rs
```diff
@@ -71,3 +82,114 @@ impl Compiler for ExecutableCompiler {
...
+        -> {scarb_toml}
+        {manifest}
+        "#,
+        executable_names
+    }
+}
+
+/// Search crates identified by `main_crate_ids` for functions annotated with `#[executable]` attribute.
+/// If `executable_path` is provided, only functions with exactly the same path will be returned.
+fn find_executable_functions(
```
maciektr, line 159: Can be simplified after LS cairo bump.

extensions/scarb-execute/src/lib.rs
```diff
@@ -205,6 +211,54 @@ pub fn execute(
...
+            let executable_function_match =
+                if let (Some(left), Some(right)) = (build_target_function, executable_function) {
+                    left == right
+                } else {
+                    false
+                };
+            let executable_name_match = executable_name
+                .as_ref()
+                .map(|name| &target.name == name)
+                .unwrap_or_default();
```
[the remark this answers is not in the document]
DelevoXDG: > We still need as_ref

Can you pls elaborate why?

scarb/src/compiler/compilers/executable.rs
```diff
@@ -71,3 +82,114 @@ impl Compiler for ExecutableCompiler {
...
+                .iter()
+                .map(|executable| originating_function_path(db, *executable))
+                .sorted()
+                .collect_vec();
+            let scarb_toml = unit
+                .main_component()
+                .package
+                .manifest_path()
+                .workspace_relative(ws);
+            anyhow::bail!(ambiguous_executable_error_message(
```
DelevoXDG, line 115: Nit:
why not import `bail!`?

scarb/src/compiler/compilers/executable.rs
```diff
@@ -71,3 +82,114 @@ impl Compiler for ExecutableCompiler {
...
+                executable_names,
+                scarb_toml
+            ));
+        }
+    };
+
+    compile_executable_function_in_prepared_db(db, executable, diagnostics_reporter, config)
+}
+
+fn ambiguous_executable_error_message(executables: Vec<String>, scarb_toml: &Utf8Path) -> String {
```
DelevoXDG, line 125: I feel like current name may be misleading 

```suggestion
fn multiple_executables_error_message(executables: Vec<String>, scarb_toml: &Utf8Path) -> String {
```

scarb/src/compiler/compilers/executable.rs
```diff
@@ -71,3 +82,114 @@ impl Compiler for ExecutableCompiler {
...
+    db: &RootDatabase,
+    main_crate_ids: Vec<CrateId>,
+    executable_path: Option<&str>,
+) -> Vec<ConcreteFunctionWithBodyId> {
+    let mut executables: Vec<_> = find_executable_function_ids(db, main_crate_ids)
+        .into_iter()
+        .filter_map(|(id, labels)| {
+            labels
+                .into_iter()
+                .any(|l| l == EXECUTABLE_RAW_ATTR)
```
DelevoXDG, line 169: Nit:
```suggestion
                .any(|label| label == EXECUTABLE_RAW_ATTR)
```

scarb/src/compiler/compilers/executable.rs
```diff
@@ -71,3 +82,114 @@ impl Compiler for ExecutableCompiler {
...
+/// Returns the path to the function that the executable is wrapping.
+///
+/// If the executable is not wrapping a function, returns the full path of the executable.
+fn originating_function_path(db: &RootDatabase, wrapper: ConcreteFunctionWithBodyId) -> String {
+    let semantic = wrapper.base_semantic_function(db);
+    let wrapper_name = semantic.name(db);
+    let wrapper_full_path = semantic.full_path(db.upcast());
+    let Some(wrapped_name) = wrapper_name.strip_prefix(EXECUTABLE_PREFIX) else {
+        return wrapper_full_path;
+    };
```
DelevoXDG, line 190:
```suggestion
    let Some(function_name) = wrapper_name.strip_prefix(EXECUTABLE_PREFIX) else {
        return wrapper_full_path;
    };
```

It's really confusing when we have both `wrapper_name` and `wrapped_name` nearby.

scarb/src/compiler/compilers/executable.rs
```diff
@@ -71,3 +82,114 @@ impl Compiler for ExecutableCompiler {
...
+    let Some(wrapper_path_to_module) = wrapper_full_path.strip_suffix(wrapper_name.as_str()) else {
+        return wrapper_full_path;
+    };
```
DelevoXDG, line 193: It would be helpful to have some comments for these cases

extensions/scarb-execute/src/lib.rs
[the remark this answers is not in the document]
maciektr: Ah, you are right, we do not!

extensions/scarb-execute/src/args.rs
```diff
@@ -30,6 +30,14 @@ pub struct ExecutionArgs {
...
     pub no_build: bool,
 
+    /// Choose build target to run by target name.
+    #[arg(long)]
+    pub executable_name: Option<String>,
+
+    /// Choose build target to run by function path.
+    #[arg(long, conflicts_with = "executable_name")]
+    pub executable_function: Option<String>,
+
```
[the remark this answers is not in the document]
maciektr: What exactly do you want to achieve? 🤔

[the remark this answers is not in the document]
DelevoXDG: To organize it same way as `ProgramArguments`, and pass this struct to `find_build_target` instead of individual args.

But as I said, it's a nitpick

scarb/src/compiler/compilers/executable.rs
maciektr: This function is copied from the compiler as is and will be replaced with import as soon as cairo-ls is bumped to compiler main.

extensions/scarb-execute/src/args.rs
DelevoXDG, line 1: Also, we should probably update scarb-prove to support the new args.

maciektr: Hmm, doesn't it work ootb, since it imports the args parser? 🤔 
https://github.com/software-mansion/scarb/blob/3910e927d53aabc1b626868053cca6c276c25462/extensions/scarb-prove/src/main.rs#L51

extensions/scarb-execute/src/lib.rs
```diff
@@ -205,6 +213,87 @@ pub fn execute(
...
+    if let Some(function) = executable_function {
+        bail!(
+            "no executable target with executable function `{function}` found for package `{}`",
+            package.name
+        )
+    }
+    let mut executable_targets = package
+        .targets
+        .iter()
+        .filter(|target| target.kind.as_str() == "executable");
```
DelevoXDG, line 259: Why not collect to vec once?

Can avoid double-filtering, remove some duplications, and simplify check we perform at the end of this funcion.

extensions/scarb-execute/src/lib.rs
```diff
@@ -205,6 +213,87 @@ pub fn execute(
...
+    let Some(first_target) = executable_targets.next() else {
+        bail!(missing_executable_target_error(metadata, package))
+    };
+    if executable_targets.next().is_some() {
+        let msg = formatdoc! {r#"
+                    more than one executable target found for package `{}`
+                    help: specify the target with `--executable-name` or `--executable-function`
+                    "#, package.name};
+        bail!(msg);
+    }
```
DelevoXDG, line 270: Consider using `ensure!`

extensions/scarb-execute/src/args.rs
DelevoXDG: Sure, but we also need to update the conflicts: https://github.com/software-mansion/scarb/blob/3910e927d53aabc1b626868053cca6c276c25462/extensions/scarb-prove/src/main.rs#L29

maciektr: Ah, makes sense, thanks!

extensions/scarb-execute/src/lib.rs
maciektr: Is this really worth rewriting? 🤔  This is logically equivalent, has roughly 50 loc and all it does is iterate a list and emit warnings.

scarb/src/compiler/compilers/executable.rs
```diff
@@ -71,3 +85,138 @@ impl Compiler for ExecutableCompiler {
...
+                scarb_toml
+            ));
+        }
+    };
+
+    compile_executable_function_in_prepared_db(db, executable, diagnostics_reporter, config)
+}
+
+fn multiple_executables_error_message(executables: Vec<String>, scarb_toml: &Utf8Path) -> String {
+    let executable_names = executables.clone().join("\n    ");
```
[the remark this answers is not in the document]
maciektr: Hmm, why do you care? 🤔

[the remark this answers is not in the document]
maciektr: Ok!

extensions/scarb-execute/src/lib.rs
DelevoXDG: Just wondered if there's a specific reason you didn't want to include it when I initially suggested it: https://github.com/software-mansion/scarb/pull/1951#discussion_r1942627767 

> Is this really worth rewriting? 

I think the upsides I mentioned are valid, but I suppose it isn't critical 🤔

extensions/scarb-execute/src/lib.rs
```diff
@@ -205,6 +208,89 @@ pub fn execute(
...
+) -> Result<&'a TargetMetadata> {
+    let executable_targets = package
+        .targets
+        .iter()
+        .filter(|target| target.kind.as_str() == "executable")
+        .collect::<Vec<_>>();
+
+    if executable_targets.is_empty() {
+        bail!(missing_executable_target_error(metadata, package));
+    }
```
DelevoXDG, line 224: Nit:

Consider using `ensure!`

DelevoXDG (approved):
LGTM

PR #1957, opened 2025-02-07
Pubgrub lockfile lookup fix

### LINKED ISSUE #1883: Pubgrub resolve locked versions bug

maciektr:
fixes #1883

When resolving package dependencies, the resolver accepts workspace lockfile (Scarb.lock) as one of the inputs. If manifest dependencies can be locked to concrete revisions defined in the lockfile (https://github.com/software-mansion/scarb/blob/52c5afe0e26c49e81b93d1b8929dc500a61495e5/scarb/src/core/lockfile.rs#L98) (i.e. unless the dependency conflicts with the locked version, for instance because the rev in dependency has been changed by the user between runs), it should be used by the resolver instead of the most recent one. 

This PR attempts to solve a bug in mechanism of applying locked revisions in the resolver. When querying the registry for dependency package summaries (https://github.com/software-mansion/scarb/blob/47c7c7d8dda31fccc4f63836027977e4ea0f28f4/scarb/src/resolver/algorithm/provider.rs#L181) we try to lock the manifest dependency to the version specified in the lockfile. You can see, that we actually request two different dependencies. If our package A depends on some package B via git dependency, but B depends on another package C via path dependency, we attempt to rewrite the dependency between B and C from path to git. This allows unification if A (or other package in the dep tree) also depends on C via git dependency. However, we cannot know for sure if this attempt will be successful, as C may be under B in the directory tree, thus being shadowed when access via git dependency (we traverse the dir tree to look for packages, but we do not enter subdirectories of any directory with Scarb.toml file). Instead, we send to request and use the rewritten git one if it succeeds, or fallback to path one if not. 

This code did not attempt to lock a transitive dependency that has been rewritten from path to git according to the lockfile, which it should.

Changes:
- **Revert "Temporarily default to primitive resolver (#1892)"**
- **Update error message**
- **Lookup dependency in lockfile after rewriting nested path to git source when requesting dependency metadata**

### DIFF
--- a/scarb/src/resolver/algorithm/provider.rs
+++ b/scarb/src/resolver/algorithm/provider.rs
@@ -179,21 +179,16 @@ impl PubGrubDependencyProvider {
     }
 
     fn request_dependencies(&self, summary: &Summary) -> Result<(), DependencyProviderError> {
-        for dep in summary.dependencies.iter() {
-            let locked_package_id = self.lockfile.packages_matching(dep.clone());
-            let dep = if let Some(locked_package_id) = locked_package_id {
-                rewrite_locked_dependency(dep.clone(), locked_package_id?)
-            } else {
-                dep.clone()
-            };
-
+        for dependency in summary.dependencies.iter() {
+            let dep = lock_dependency(&self.lockfile, dependency.clone())?;
             if self.state.index.packages().register(dep.clone()) {
                 self.request_sink
-                    .blocking_send(Request::Package(dep.clone()))
+                    .blocking_send(Request::Package(dep))
                     .unwrap();
             }
 
-            let dep = rewrite_path_dependency_source_id(summary.package_id, &dep)?;
+            let dep = rewrite_path_dependency_source_id(summary.package_id, dependency);
+            let dep = lock_dependency(&self.lockfile, dep)?;
             if self.state.index.packages().register(dep.clone()) {
                 self.request_sink
                     .blocking_send(Request::Package(dep))
@@ -318,20 +313,14 @@ impl DependencyProvider for PubGrubDependencyProvider {
             .filtered_full_dependencies(dep_filter)
             .cloned()
             .map(|dependency| {
-                let original_dep = dependency.clone();
-                let dependency =
-                    rewrite_path_dependency_source_id(summary.package_id, &dependency)?;
-                let locked_package_id = self.lockfile.packages_matching(dependency.clone());
-                let dependency = if let Some(locked_package_id) = locked_package_id {
-                    rewrite_locked_dependency(dependency.clone(), locked_package_id?)
-                } else {
-                    dependency
-                };
+                let original_dependency = dependency.clone();
+                let dependency = rewrite_path_dependency_source_id(summary.package_id, &dependency);
+                let dependency = lock_dependency(&self.lockfile, dependency)?;
 
                 let dep_name = dependency.name.clone().to_string();
                 let summaries = self.wait_for_summaries(dependency.clone())?;
                 let summaries = if summaries.is_empty() {
-                    self.wait_for_summaries(original_dep.clone())?
+                    self.wait_for_summaries(original_dependency.clone())?
                 } else {
                     summaries
                 };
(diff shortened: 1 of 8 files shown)


scarb/src/resolver/algorithm/provider.rs
```diff
@@ -179,21 +179,16 @@ impl PubGrubDependencyProvider {
...
-        for dep in summary.dependencies.iter() {
-            let locked_package_id = self.lockfile.packages_matching(dep.clone());
-            let dep = if let Some(locked_package_id) = locked_package_id {
-                rewrite_locked_dependency(dep.clone(), locked_package_id?)
-            } else {
-                dep.clone()
-            };
-
+        for dependency in summary.dependencies.iter() {
+            let dep = lock_dependency(&self.lockfile, dependency.clone())?;
```
DelevoXDG, line 183: Nit: 
Why not overwrite `dep`?

It seems mildly confusing to have two things in scope that refer to the same thing

scarb/src/resolver/algorithm/provider.rs
```diff
@@ -179,21 +179,16 @@ impl PubGrubDependencyProvider {
...
             if self.state.index.packages().register(dep.clone()) {
                 self.request_sink
-                    .blocking_send(Request::Package(dep.clone()))
+                    .blocking_send(Request::Package(dep))
                     .unwrap();
             }
 
-            let dep = rewrite_path_dependency_source_id(summary.package_id, &dep)?;
+            let dep = rewrite_path_dependency_source_id(summary.package_id, dependency);
+            let dep = lock_dependency(&self.lockfile, dep)?;
```
DelevoXDG, line 191: Can you please clarify why do we need to lock the dep twice? 🤔

scarb/src/resolver/algorithm/provider.rs
```diff
@@ -366,6 +355,18 @@ impl From<DependencyVersionReq> for SemverPubgrub {
...
+fn lock_dependency(
+    lockfile: &Lockfile,
+    dep: ManifestDependency,
+) -> Result<ManifestDependency, DependencyProviderError> {
+    lockfile
+        .packages_matching(dep.clone())
+        .map(|locked_package_id| Ok(rewrite_locked_dependency(dep.clone(), locked_package_id?)))
+        .unwrap_or(Ok(dep))
+}
+
```
DelevoXDG, line 369: Why not `rewrite_path_dependency_source_id` as a part of this?

scarb/tests/git_source.rs
```diff
@@ -463,6 +463,45 @@ fn transitive_path_dep() {
...
+        .arg("fetch")
+        .current_dir(&t)
+        .assert()
+        .success();
+
+    Scarb::quick_snapbox()
+        .arg("fetch")
+        .current_dir(&t)
+        .assert()
+        .success();
```
DelevoXDG, line 502: Should we also check lockfile contents as a part of this test?

scarb/src/resolver/algorithm/provider.rs
maciektr: It's possible, that we wouldn't be able to lock dependency before rewriting (e.g. if it's a path dep) but we can after rewriting (e.g. for path rewritten to git).

scarb/tests/git_source.rs
maciektr: The dep did not change, so there is nothing else that could be written to lock I think - but we can check if the lock has been used, added assert on that.

scarb/src/resolver/algorithm/provider.rs
DelevoXDG: Sorry if I'm missing something, but why not attempt rewriting path dep to git one, and then attempt to lock?

scarb/src/resolver/mod.rs
```diff
@@ -38,8 +38,7 @@ pub async fn resolve(
             let s = var.as_str();
             s == "true" || s == "1"
         })
-        // Defaults to primitive!
-        .unwrap_or(true);
+        .unwrap_or(false);
```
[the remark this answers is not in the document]
maciektr: The flag did not change, only the value used when it's not set

DelevoXDG (approved):

PR #1979, opened 2025-02-14
feat: ui print doc dir

FroyaTheHen:


extensions/scarb-doc/src/main.rs
```diff
@@ -87,7 +88,7 @@ fn main_inner(args: Args, ui: Ui) -> Result<()> {
             }
         }
     }
-
+    ui.print(Status::new("Saved to directory:", output_dir.as_str()));
```
FroyaTheHen, line 91: imo, path `.../target/doc`  should be informative enough but let me know if you disagree (why) and think that the entire path `.../target/doc/my_project/output.json` file or `.../target/doc` respectively for user chosen output-format should be printed out.

extensions/scarb-doc/tests/core.rs
```diff
@@ -21,3 +23,25 @@ fn can_doc_corelib() {
...
+        .current_dir(&t)
+        .stdout_json::<Metadata>();
+
+    Scarb::quick_snapbox()
+        .arg("doc")
+        .current_dir(&t)
+        .assert()
+        .success()
+        .stdout_matches(formatdoc! {r#"
+                Saved to directory: {dir}/doc
```
maciektr, line 43: Instead of 
```
Saved to directory: /private/var/folders/2t/7dt08_ss585ct08q57mvw_q40000gn/T/.tmpBguKBZ/target/doc
```

Please make it 

```
Saving output to: target/doc/hello_world/
```
And for `--output-format=json`
```
Saving output to: target/doc/output.json
```

You can take a look at scarb-execute/main.rs fn display_path for the prefix removal. I suggest "Saving" because we do "Compiling", "Running", etc for statuses - not sure if it's grammatically correct though 🤔 

Btw. `output.json` can be changed to `hello_world.json`, but that a separate problem :D

maciektr (changes requested):

extensions/scarb-doc/tests/core.rs
FroyaTheHen: Well, for sure it is grammatically correct, just might not exactly be the truth if the info is printed to the output after the files already exists. However, it makes perfect sense to change to "Saving" for the sake of consistency. Will do!

maciektr (approved):
👍

PR #1984, opened 2025-02-17
Split nightly crates tests job to build and run steps

maciektr:
This way we can build stable and nightly in parallel, instead of first waiting for stable to build and then building nightly. I suggest using nextest, even if we do not use partitioning, because it's well suited for archiving build and running from the archive.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,11 +89,37 @@ jobs:
       - name: run tests
         run: cargo test -p scarb-metadata
 
-  test-nightly-crates:
-    name: test nightly crates ${{ matrix.platform.name }}
+  build-nightly-test:
+    name: build nightly test ${{ matrix.platform.name }}
+    runs-on: ${{ matrix.platform.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - name: linux x86-64
+            os: ubuntu-latest
+          - name: windows x86-64
+            os: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_NIGHTLY_TOOLCHAIN }}
+      - uses: taiki-e/install-action@nextest
+      - uses: Swatinem/rust-cache@v2
+      - name: nextest archive
+        run: cargo nextest archive --all-features --cargo-profile ci --archive-file 'nextest-nightly-archive-${{ matrix.platform.os }}.tar.zst' --package scarb-prove --package scarb-verify
+      - uses: actions/upload-artifact@v4
+        with:
+          name: nextest-nightly-archive-${{ matrix.platform.os }}
+          path: nextest-nightly-archive-${{ matrix.platform.os }}.tar.zst
+
+  test-nightly:
+    name: test nightly ${{ matrix.platform.name }}
     runs-on: ${{ matrix.platform.os }}
     needs:
       - build-test
+      - build-nightly-test
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -89,11 +89,37 @@ jobs:
...
+    name: build nightly test ${{ matrix.platform.name }}
+    runs-on: ${{ matrix.platform.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - name: linux x86-64
+            os: ubuntu-latest
+          - name: windows x86-64
+            os: windows-latest
```
DelevoXDG, line 102: Maybe this should be commented out as well, as we're only testing for linux? 🤔

.github/workflows/ci.yml
```diff
@@ -108,20 +134,34 @@ jobs:
...
+      # Do not run tests on windows, until stwo supports it.
+      # - name: Unpack build stable rust artifacts from nextest archive on windows
+      #   if: matrix.platform.os == 'windows-latest'
+      #   run: |
+      #     zstd -d nextest-archive-windows-latest.tar.zst
+      #     tar -xf nextest-archive-windows-latest.tar
+      #     ren target stable-crates-target
+      #     echo "NEXTEST_BIN_EXE_scarb=$env:GITHUB_WORKSPACE/stable-crates-target/ci/scarb" >> $env:GITHUB_ENV
+      #     Add-Content $env:GITHUB_PATH "$env:GITHUB_WORKSPACE/stable-crates-target/ci"
+      #     Add-Content $env:GITHUB_PATH "$env:GITHUB_WORKSPACE/target/ci"
```
DelevoXDG, line 159: Nit: I would rather remove it altogether, as afaik it's unknown when and whether swto will get windows support, and we probably don't know for sure that this works.

.github/workflows/ci.yml
```diff
@@ -108,20 +134,34 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ env.RUST_NIGHTLY_TOOLCHAIN }}
+      - uses: taiki-e/install-action@nextest
       - uses: Swatinem/rust-cache@v2
       - uses: actions/download-artifact@v4
```
DelevoXDG, line 139: Nit:
```suggestion
      - name: Download stable build artifacts  
         uses: actions/download-artifact@v4
```

.github/workflows/ci.yml
```diff
@@ -108,20 +134,34 @@ jobs:
...
+      # - name: Unpack build stable rust artifacts from nextest archive on windows
+      #   if: matrix.platform.os == 'windows-latest'
+      #   run: |
+      #     zstd -d nextest-archive-windows-latest.tar.zst
+      #     tar -xf nextest-archive-windows-latest.tar
+      #     ren target stable-crates-target
+      #     echo "NEXTEST_BIN_EXE_scarb=$env:GITHUB_WORKSPACE/stable-crates-target/ci/scarb" >> $env:GITHUB_ENV
+      #     Add-Content $env:GITHUB_PATH "$env:GITHUB_WORKSPACE/stable-crates-target/ci"
+      #     Add-Content $env:GITHUB_PATH "$env:GITHUB_WORKSPACE/target/ci"
+      - uses: actions/download-artifact@v4
```
DelevoXDG, line 160: Nit:
```suggestion
      - name: Download nightly build artifacts  
         uses: actions/download-artifact@v4
```

.github/workflows/ci.yml
maciektr: We can 👍

DelevoXDG (approved):
I would still add one-line comment explaining why windows is removed from build and tests mentioning stwo, otherwise 👍🏻

PR #1994, opened 2025-02-18
Propagate ui verbosity when using scarbcommand in extensions

### LINKED ISSUE #1923: Verbosity settings are not propagated to Scarb calls in subcommands
`cairo-run`, `cairo-test`, `execute` rely on `ScarbCommand` to make calls to Scarb, e.g.:
https://github.com/software-mansion/scarb/blob/18f1ca120c4ae72664e9e192e1555b1d6fc905e5/extensions/scarb-execute/src/main.rs#L149

However, verbosity settings are not propagated to these Scarb calls:
```
scarb execute --quiet
   Compiling executable_example v0.1.0 ([..]/scarb/examples/executable_example/Scarb.toml)
    Finished `dev` profile target(s) in 2 seconds
```

We should propagate these either via flag or env.

maciektr:
fixes #1923 
- **Use scarb-ui in cairo-test**
- **Propagate verbosity level when calling ScarbCommand**

### DIFF
--- a/extensions/scarb-cairo-run/src/main.rs
+++ b/extensions/scarb-cairo-run/src/main.rs
@@ -101,11 +101,7 @@ fn main_inner(ui: &Ui, args: Args) -> Result<()> {
         ScarbCommand::new()
             .arg("build")
             .env("SCARB_PACKAGES_FILTER", filter.to_env())
-            .env_if(
-                "SCARB_UI_VERBOSITY",
-                ui.verbosity().to_string(),
-                !ui.verbosity().is_default(),
-            )
+            .env("SCARB_UI_VERBOSITY", ui.verbosity().to_string())
             .run()?;
     }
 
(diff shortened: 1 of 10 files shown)


extensions/scarb-cairo-run/src/main.rs
```diff
@@ -101,6 +101,11 @@ fn main_inner(ui: &Ui, args: Args) -> Result<()> {
         ScarbCommand::new()
             .arg("build")
             .env("SCARB_PACKAGES_FILTER", filter.to_env())
+            .env_if(
+                "SCARB_UI_VERBOSITY",
+                ui.verbosity().to_string(),
+                !ui.verbosity().is_default(),
+            )
```
DelevoXDG, line 108: Do we really need this? 🤔 

If verbosity is default, why not just set it to default instead of skipping it? 🤔

maciektr: I am sure there was some reason why I did it like this! But I don't remember and can't catch any problems in tests, so removed.

DelevoXDG (approved):

Issue #1996, opened 2025-02-19
Add tests that cover plugin-per-crate changes


mkaput:
As far as I know there are no tests that cover plugin-per-crate mechanism so techincally this thing is untested.

say, there is not code that checks whether two conflicting macros can be using by 2 crates that belong to the same CU

PR #2000, opened 2025-02-21
Cairo plugin reexports

### LINKED ISSUE #2004: Support re-exporting proc macros from a Cairo lib package
### Problem

Create a manifest syntax from proc macro reexports 

### Proposed Solution

_No response_

### Notes

_No response_

maciektr:
fixes #2004

- **Fix duplicated component dependency on itself**
- **Cairo plugin reexports**

### DIFF
--- a/scarb/src/ops/resolve.rs
+++ b/scarb/src/ops/resolve.rs
@@ -59,13 +59,50 @@ impl WorkspaceResolve {
         &self,
         package_id: PackageId,
         target_kind: &TargetKind,
-    ) -> Vec<Package> {
+    ) -> Result<Vec<Package>> {
         assert!(self.packages.contains_key(&package_id));
-        self.resolve
+        let dependencies = self
+            .resolve
             .package_dependencies_for_target_kind(package_id, target_kind)
             .iter()
             .map(|id| self.packages[id].clone())
-            .collect_vec()
+            .collect_vec();
+
+        let re_exported = dependencies
+            .iter()
+            .flat_map(|dependency| {
+                let package_dependencies = self
+                    .resolve
+                    .package_dependencies_for_target_kind(dependency.id, target_kind);
+                dependency
+                    .manifest
+                    .summary
+                    .re_export_cairo_plugins
+                    .iter()
+                    .map(move |plugin_name| {
+                        package_dependencies
+                            .iter()
+                            .find(|id| id.name == *plugin_name)
+                            .map(|id| {
+                                self.packages.get(id).expect("workspace resolve packages must include all dependency graph nodes").clone()
+                            })
+                            .filter(|package| package.is_cairo_plugin())
+                            .map(Ok)
+                            .unwrap_or_else(|| {
+                                bail!(
+                                "package `{}` cannot re-export cairo plugin `{plugin_name}` which is not a dependency of `{}`",
+                                dependency.id.name, dependency.id.name)
+                            })
+                    })
+
+            })
+            .collect_vec();
+
+        dependencies
+            .into_iter()
+            .map(Ok)
+            .chain(re_exported)
+            .collect::<Result<Vec<Package>>>()
     }
 
     /// Get all dependencies with allowed prebuilt macros for a given package.
@@ -517,7 +554,7 @@ fn cairo_compilation_unit_for_target(
         .iter()
         .find(|component| component.package.id == member.id)
         .unwrap();
-    let mut test_package_deps = solution.component_dependencies(member_component, &components);
+    let mut test_package_deps = solution.component_dependencies(member_component, &components)?;
     if is_integration_test {
         test_package_deps.push(CompilationUnitDependency::Library(
             member_component.id.clone(),
@@ -527,13 +564,13 @@ fn cairo_compilation_unit_for_target(
     let dependencies_for_components: Vec<_> = components
         .iter()
         .map(|component| {
-            if component.package.id == test_package_id {
+            Ok(if component.package.id == test_package_id {
                 test_package_deps.clone()
             } else {
-                solution.component_dependencies(component, &components)
-            }
+                solution.component_dependencies(component, &components)?
+            })
         })
-        .collect();
+        .collect::<Result<Vec<_>>>()?;
 
     for (component, dependencies) in zip(&mut components, dependencies_for_components) {
         component.dependencies = dependencies;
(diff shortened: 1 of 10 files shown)


scarb/tests/build_cairo_plugin.rs
```diff
@@ -1559,3 +1559,69 @@ fn can_expand_impl_inner_func_attrr() {
...
+            }
+        "##})
+        .build(&t);
+    let dep = temp.child("dep");
+    ProjectBuilder::start()
+        .name("dep")
+        .version("1.0.0")
+        .dep("some", &t)
+        .manifest_package_extra(indoc! {r#"
+            cairo-plugin-reexports = ["some"]
```
DelevoXDG, line 1588: Nit:

`re-export-cairo-plugins` seems more readable to me.

It's more typical to see "re-export" with dash, hence my mild confusion.

scarb/tests/cairo_plugin_reexports.rs
```diff
@@ -0,0 +1,244 @@
...
+    assert_eq!(
+        beautiful_deps_packages.len(),
+        beautiful_deps.len(),
+        "all dependencies should be found in the packages list"
+    );
+    assert_eq!(
+        beautiful_deps_packages,
+        vec!["beautiful", "core", "world"],
+        "beautiful component invalid dependencies"
+    );
```
DelevoXDG, line 98: Why check the `beautiful` comp? 

Aren't checks on `hello` comp not sufficient to test whether re-exports work? 🤔

scarb/tests/metadata.rs
```diff
@@ -1538,3 +1538,122 @@ fn executable_target_can_allow_syscalls() {
         "syscalls not allowed"
     );
 }
+
+#[test]
+fn cairo_plugins_added_as_component_dependencies() {
```
DelevoXDG, line 1543: Correct me if I'm wrong, but this test is basically the same as `cairo_plugin_reexport_simple`, but without re-exports?

Wouldn't it be reasonable to name this test accordingly, and move it to the same file.

scarb/src/ops/resolve.rs
```diff
@@ -517,21 +554,23 @@ fn cairo_compilation_unit_for_target(
...
-    let mut test_package_deps = solution.component_dependencies(member_component, &components);
-    test_package_deps.push(CompilationUnitDependency::Library(
-        member_component.id.clone(),
-    ));
+    let mut test_package_deps = solution.component_dependencies(member_component, &components)?;
+    if is_integration_test {
+        test_package_deps.push(CompilationUnitDependency::Library(
+            member_component.id.clone(),
+        ));
+    }
```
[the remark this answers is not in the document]
maciektr: https://github.com/software-mansion/scarb/blob/15cb8e10d2b6f3cc921fa0a390b8e82a1b357cae/scarb/tests/metadata.rs#L1543
There was a bug introduced with per-crate plugins, which would duplicate the dep in some cases

scarb/src/ops/resolve.rs
```diff
@@ -59,13 +59,50 @@ impl WorkspaceResolve {
...
+            .resolve
             .package_dependencies_for_target_kind(package_id, target_kind)
             .iter()
             .map(|id| self.packages[id].clone())
-            .collect_vec()
+            .collect_vec();
+
+        let mut reexported = Vec::new();
+        for dependency in &dependencies {
+            let package_dependencies = self
```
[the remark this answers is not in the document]
maciektr: We would need to collect the result into a separate vector before flat_map (https://stackoverflow.com/questions/57902478/why-do-i-need-to-collect-into-a-vector-when-using-flat-map), so I prefer extending the vec here directly.

scarb/tests/cairo_plugin_reexports.rs
maciektr: Why not? 🤔

DelevoXDG: Why yes? 😄 

Unless there's a specific reason to have it, it seems like it unnecessarily complicates the test.

scarb/src/ops/resolve.rs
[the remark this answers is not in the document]
DelevoXDG: I think this should work
```rs
        let re_exported = dependencies
        .iter()
        .flat_map(|dependency| {
            let package_dependencies = self
                .resolve
                .package_dependencies_for_target_kind(dependency.id, target_kind);

            dependency
                .manifest
                .summary
                .re_export_cairo_plugins
                .iter()
                .map(move |plugin_name| {
                    package_dependencies
                        .iter()
                        .find(|id| id.name == *plugin_name)
                        .map(|id| {
                            self.packages.get(id).expect("workspace resolve packages must include all dependency graph nodes").clone()
                        })
                        .filter(|package| package.is_cairo_plugin())
                        .map(Ok)
                        .unwrap_or_else(|| {
                            bail!(
                                "package `{}` cannot re-export cairo plugin `{plugin_name}` which is not a dependency of `{}`",
                                dependency.id.name, dependency.id.name
                            )
                        })
                })
        })
        .collect_vec();
```

Not sure if that's what you've been referring to, but it seems we don't need `collect_vec` ***before*** `flat_map`. It doesn't seem like singular `collect_vec` is worse than extending a mutable vector 🤔

scarb/tests/metadata.rs
maciektr: > Similar tests that relate to the same logic should be kept in the same file.
> It might not test re-exports directly, but this test does aim to check that re-exports are not default behavior

But it's not the same logic. Cases in the reexports file test a very specific feature of plugin reexports. This test case ensures that plugin dependencies are assigned by the resolver correctly. We need this check wether we implement reexports or not. 

>  since it wasn't part of metadata tests before?

#1996 

> Since there a lot of duplicate logic, it's easier to include changes across the board if necessary.

At the end of the day, this logic is reading a value of field in json.

scarb/tests/cairo_plugin_reexports.rs
maciektr: I don't see what specific bug could cause a package to stop depending on a plugin it reexports, but if we test dependency manipulation around reexports, IMHO It's a valid assertion to make.

DelevoXDG: Ah, I get your logic here now

> I don't see what specific bug could cause a package to stop depending on a plugin it reexports

Me too, but I suppose it's theoretically possible package is somehow manipulated and loses a dependency. Let's keep it 👍🏻

scarb/tests/metadata.rs
DelevoXDG: > But it's not the same logic. Cases in the reexports file test a very specific feature of plugin reexports. This test case ensures that plugin dependencies are assigned by the resolver correctly. We need this check wether we implement reexports or not.

Ok, makes sense

> #1996

> say, there is not code that checks whether two conflicting macros can be using by 2 crates that belong to the same CU

Not sure that's what is being tested there?

scarb/src/ops/resolve.rs
```diff
@@ -59,13 +59,50 @@ impl WorkspaceResolve {
...
-        self.resolve
+        let dependencies = self
+            .resolve
             .package_dependencies_for_target_kind(package_id, target_kind)
             .iter()
             .map(|id| self.packages[id].clone())
-            .collect_vec()
+            .collect_vec();
+
+        let reexported = dependencies
```
DelevoXDG, line 71:
```suggestion
        let re_exported = dependencies
```
Let's update it everywhere across the codebase, including the docs, just so all related elements are easily searchable.

DelevoXDG (approved):

PR #2006, opened 2025-02-21
Update rust edition to 2024

maciektr:
- **Add rustfmt file**
- **Reformat with 2024 style edition**
- **Update Rust edition**
- **Wrap set_env calls in tests in unsafe block**
- **disambiguate into_future call**
- **Add missing unsafe blocks**


plugins/cairo-lang-macro-attributes/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "cairo-lang-macro-attributes"
 version = "0.1.0"
-edition.workspace = true
+edition = "2021"
```
DelevoXDG, line 4: Are we planning to update plugins to `2024` separately?

.rustfmt.toml
```diff
@@ -0,0 +1 @@
+style_edition = "2024"
```
DelevoXDG, line 1: Nit:

image

plugins/cairo-lang-macro-attributes/Cargo.toml
maciektr: This crates are compiled on user hardware, so updating the edition would enforce all of them to update to Rust 1.85, which we cannot do, so it will need to stay on 2021 for a long time

DelevoXDG (approved):

PR #2007, opened 2025-02-21
Do not compile unused proc macros when called with packages filter

### LINKED ISSUE #2008: Do not build proc macros that are not used in compilation due to packages filters

maciektr:
fixes #2008

### DIFF
--- a/scarb/src/ops/compile.rs
+++ b/scarb/src/ops/compile.rs
@@ -117,7 +117,7 @@ fn process<F>(
     operation_type: Option<&str>,
 ) -> Result<()>
 where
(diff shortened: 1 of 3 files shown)


FroyaTheHen (commented):
Do I understand correctly that this PR has purely an optimisation purpose?

maciektr:
@FroyaTheHen Other reason would be that sometimes you may want to work on a workspace with plugins even if you don't have Cargo installed. Also, it's just more correct - if you build (only) a package that does not depend on some macro, you shouldn't need to build that macro as well.

FroyaTheHen (approved):

PR #2025, opened 2025-02-25
Update test ProjectBuilder default Cairo edition to 2024_07

maciektr:
commit-id:7e728476

Fix test fixtures after edition change

---

**Stack**:
- #2028
- #2027
- #2025 ⬅


extensions/scarb-doc/tests/diagnostics.rs
```diff
@@ -100,6 +100,7 @@ fn test_diagnostics_error() {
 
     ProjectBuilder::start()
         .name("hello_world")
+        .edition("2023_01")
```
DelevoXDG, line 103: Can you please provide some context?

maciektr: Right now all tests run on `2023_01`, so they may rely on some behaviour that has been changed in a newer release (like not having the concept of visibility, or different corelib prelude). Using the older edition allows us to avoid rewriting all fixtures.

DelevoXDG (approved):

PR #2029, opened 2025-02-25
Ignore `debug_info` when emitting file size warnings for contracts compilation

### LINKED ISSUE #2016: Ignore debug_info when emitting file size warnings for contracts compilation
### Problem

tbd

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text

```

DelevoXDG:
Closes #2016

### DIFF
--- a/scarb/src/compiler/compilers/starknet_contract/artifacts_writer.rs
+++ b/scarb/src/compiler/compilers/starknet_contract/artifacts_writer.rs
@@ -6,8 +6,10 @@ use crate::flock::Filesystem;
 use cairo_lang_compiler::db::RootDatabase;
 use cairo_lang_defs::ids::NamedLanguageElementId;
 use cairo_lang_starknet::contract::ContractDeclaration;
+use cairo_lang_starknet_classes::abi::Contract;
 use cairo_lang_starknet_classes::casm_contract_class::CasmContractClass;
-use cairo_lang_starknet_classes::contract_class::ContractClass;
+use cairo_lang_starknet_classes::contract_class::{ContractClass, ContractEntryPoints};
+use cairo_lang_utils::bigint::BigUintAsHex;
 use cairo_lang_utils::UpcastMut;
 use indoc::formatdoc;
 use itertools::{izip, Itertools};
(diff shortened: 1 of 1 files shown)


scarb/src/compiler/compilers/starknet_contract/artifacts_writer.rs
```diff
@@ -20,6 +22,27 @@ const MAX_CASM_PROGRAM_FELTS: usize = 81290;
...
 
+// Represents a contract in the Starknet network as defined in Starknet JSON-RPC spec:
+// https://github.com/starkware-libs/starknet-specs/blob/2030a650be4e40cfa34d5051a0334f375384a421/api/starknet_api_openrpc.json#L3030
+#[derive(Clone, Debug, Serialize)]
+struct RpcContractClass<'a> {
+    sierra_program: &'a Vec<BigUintAsHex>,
+    contract_class_version: &'a String,
+    entry_points_by_type: &'a ContractEntryPoints,
+    abi: &'a Option<Contract>,
+}
```
maciektr, line 33: Can we just clone the `ContractClass` and override debug info with `None` instead? I don't really want to re-define types from `cairo_lang_starknet_classes` or mention RPC here 🤔

maciektr (commented):
Apart from previous comment, it works 🦾

scarb/src/compiler/compilers/starknet_contract/artifacts_writer.rs
DelevoXDG: Yeah, we sure can; that was my initial idea as well, but I wanted to avoid excessive cloning where we can.

If you don't think that's significant, I can make the change.

On the side note, it is kind of reasonable to have such class. If there are ever changes to `ContractClass` (new fields that are not present in RPC version), we would probably want to ignore them when emitting warnings as well. 
While it's good to keep Scarb untied from RPC versions, in this case (where we're making a check against network standards), I suppose it's one place where an exception to this rule would make sense.

WDYT?

maciektr: > avoid excessive cloning

Ok, so keep it, but please rename so it does not mention RPC.

DelevoXDG:
```suggestion
// Represents a contract in the Starknet network as defined in Starknet JSON-RPC spec:
// https://github.com/starkware-libs/starknet-specs/blob/2030a650be4e40cfa34d5051a0334f375384a421/api/starknet_api_openrpc.json#L3030
#[derive(Clone, Debug, Serialize)]
struct StarknetContractClass<'a> {
    sierra_program: &'a Vec<BigUintAsHex>,
    contract_class_version: &'a String,
    entry_points_by_type: &'a ContractEntryPoints,
    abi: &'a Option<Contract>,
}
```

WDYT?

maciektr: I want to avoid leading someone reading the code to believe that we have defined some types that can be trusted to be up to date with rpc spec or whatever the network is doing. So I would even suggest `ContractClassNoDebug` or something similar.

maciektr (approved):

Issue #2035, opened 2025-03-03
Fix: typo, factual error


FroyaTheHen:
> Typo on artifacts and avoid should decrease code size not increase (and indeed may result in slower execution)

![Image](https://github.com/user-attachments/assets/a526b5bb-2306-4dfa-af8c-ae5d9daf3e38)

Issue #1950, opened 2025-02-04
Add docs for Scarb execute & prove & verify


maciektr:
- Link STWO 
- Add a warning that the prover should not be trusted yet

TobieTom:
Can I take over this issue? I just recently learned Rust and Cairo so I am taking up a lot of challenges to add to my experience.

PR #2036, opened 2025-03-03
Add docs for execute, prove and verify, update docs

### LINKED ISSUE #1949: Add docs for executable target

maciektr:
fixes #1949 #1950 #1894 #2035

### DIFF
--- a/website/docs/extensions/execute.md
+++ b/website/docs/extensions/execute.md
@@ -58,14 +58,13 @@ They can be passed to the `scarb execute` command via either `--arguments` or `-
 The expected input with `--arguments` is a comma-separated list of integers.
 This list should correspond to the Cairo’s Serde of main’s arguments, for example:
 
-| main’s signature               | valid arguments example | valid arguments file contents example |
-| :----------------------------- | :---------------------- | :------------------------------------ |
-| `fn main(num: u8)`             | 1                       | ["0x1"]                               |
-| `fn main(num1: u8, num2: u16)` | 1,27                    | ["0x1", "0x1b"]                       |
-
-| `fn main(num1: u8, tuple: (u16, u16))` | 1,2,27 | ["0x1", "0x2", "0x1b"] |
-| `fn main(num1: u8, num2: u256)` | 1,2,27 | ["0x1", "0x2", "0x1b"] |
-| `fn main(num1: u8, arr: Array<u8>)` | 1,2,1,2 | ["0x1", "0x2", "0x1", "0x2"] |
+| main’s signature                       | valid arguments example | valid arguments file contents example |
+| :------------------------------------- | :---------------------- | :------------------------------------ |
+| `fn main(num: u8)`                     | 1                       | ["0x1"]                               |
+| `fn main(num1: u8, num2: u16)`         | 1,27                    | ["0x1", "0x1b"]                       |
+| `fn main(num1: u8, tuple: (u16, u16))` | 1,2,27                  | ["0x1", "0x2", "0x1b"]                |
+| `fn main(num1: u8, num2: u256)`        | 1,2,27                  | ["0x1", "0x2", "0x1b"]                |
+| `fn main(num1: u8, arr: Array<u8>)`    | 1,2,1,2                 | ["0x1", "0x2", "0x1", "0x2"]          |
 
 Note that when using `--arguments-file`, the expected input is an array of felts represented as hex string.
 See the [documentation](https://docs.starknet.io/architecture-and-concepts/smart-contracts/serialization-of-cairo-types/) for more information about Cairo’s Serde.
--- a/website/docs/extensions/execute.md
+++ b/website/docs/extensions/execute.md
@@ -1,0 +1,73 @@
+<script setup>
+import { data as rel } from "../../github.data";
+</script>
+
+# Scarb execute
+
+The `scarb execute` command executes a function from a local package.
+Only packages defining the [executable target](../reference/targets#executable-target) can be executed.
+It does automatically compile the cairo code within the package so using `scarb build` beforehand is not necessary.
+This automatically called build can be skipped with the `--no-build` flag.
+
+## Choosing a function to run
+
+If your package defines multiple main functions (see [choosing the main function](../reference/targets#choosing-the-main-function)),
+you need to specify which executable target should be run.
+
+This can be achieved through one of two flags:
+
+- `--executable-name` to choose the target by its name.
+- `--executable-function` to choose the target by the main function it defines.
+  Those flags are mutually exclusive.
+
+## Saving the execution information
+
+The execution will be carried out for one of two execution targets: `standalone` or `bootloader`.
+You can choose the target with the `--target` flag.
+Standalone means that the program will be executed as-is, and intended to be proven directly with `scarb prove`.
+hen we run with the bootloader target, the program’s execution is expected to be wrapped by the
+[bootloader’s](https://github.com/Moonsong-Labs/cairo-bootloader?tab=readme-ov-file#cairo-bootloader) execution,
+which itself will be proven via Stwo.
+
+After the execution, information about it will be saved to the target directory of the package.
+
+For `standalone` target, the output will be saved as trace files (`air_public_input.json`, `air_private_input.json`,
+`memory.bin`, and `trace.bin`), which can be used for creating a proof with `scarb prove`.
+For `bootloader` target, the output will be saved as a CairoPie format (Position Indenpendent Execution),
+which is not yet supported by the `prove` command.
+
+See more on [proving and verifying execution](./prove-and-verify.md) page.
+
+## Resource usage and program output
+
+To print the Cairo program output, you can use the `--print-program-output` flag.
+Otherwise, the output will be discarded.
+
+To print detailed execution resources usage, you can use the `--print-resource-usage` flag.
+This will show information about:
+
+- `n_steps`
+- `n_memory_holes`
+- `builtin_instance_counter`
+- `syscalls`
+
+## Program arguments
+
+The executable function may accept arguments.
+They can be passed to the `scarb execute` command via either `--arguments` or `--arguments-file` flag.
+
+The expected input with `--arguments` is a comma-separated list of integers.
+This list should correspond to the Cairo’s Serde of main’s arguments, for example:
+
+| main’s signature                       | valid arguments example |
+| :------------------------------------- | :---------------------- |
+| `fn main(num: u8)`                     | 1                       |
+| `fn main(num1: u8, num2: u16)`         | 1,2                     |
+| `fn main(num1: u8, tuple: (u16, u16))` | 1,2,3                   |
+| `fn main(num1: u8, num2: u256)`        | 1,2,3                   |
+| `fn main(num1: u8, arr: Array<u8>)`    | 1,2,1,2                 |
+
+See the [documentation](https://docs.starknet.io/architecture-and-concepts/smart-contracts/serialization-of-cairo-types/) for more information about Cairo’s Serde.
+
+Note that when using `--arguments-file`, the expected input is an array of felts represented as hex string.
+For example, `1,2,3` in the above table becomes `["0x1", "0x2", "0x3"]`.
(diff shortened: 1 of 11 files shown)


website/docs/extensions/execute.md
```diff
@@ -0,0 +1,73 @@
...
+hen we run with the bootloader target, the program’s execution is expected to be wrapped by the
+[bootloader’s](https://github.com/Moonsong-Labs/cairo-bootloader?tab=readme-ov-file#cairo-bootloader) execution,
+which itself will be proven via Stwo.
+
+After the execution, information about it will be saved to the target directory of the package.
+
+For `standalone` target, the output will be saved as trace files (`air_public_input.json`, `air_private_input.json`,
+`memory.bin`, and `trace.bin`), which can be used for creating a proof with `scarb prove`.
+For `bootloader` target, the output will be saved as a CairoPie format (Position Indenpendent Execution),
+which is not yet supported by the `prove` command.
```
DelevoXDG, line 37: as above
```suggestion
After the execution, information about it will be saved to the target directory of the package:

- For `standalone` target, the output will be saved as trace files (`air_public_input.json`, `air_private_input.json`,
  `memory.bin`, and `trace.bin`), which can be used for creating a proof with `scarb prove`.
- For `bootloader` target, the output will be saved as a CairoPie format (Position Indenpendent Execution),
  which is not yet supported by the `scarb prove` command.
```

website/docs/extensions/execute.md
```diff
@@ -0,0 +1,73 @@
...
+| `fn main(num: u8)`                     | 1                       |
+| `fn main(num1: u8, num2: u16)`         | 1,2                     |
+| `fn main(num1: u8, tuple: (u16, u16))` | 1,2,3                   |
+| `fn main(num1: u8, num2: u256)`        | 1,2,3                   |
+| `fn main(num1: u8, arr: Array<u8>)`    | 1,2,1,2                 |
+
+See the [documentation](https://docs.starknet.io/architecture-and-concepts/smart-contracts/serialization-of-cairo-types/) for more information about Cairo’s Serde.
+
+Note that when using `--arguments-file`, the expected input is an array of felts represented as hex string.
+For example, `1,2,3` in the above table becomes `["0x1", "0x2", "0x3"]`.
```
DelevoXDG, line 73: Hmmm, maybe we could just update the table above to have:
- valid `--arguments` example
- valid `--arguments-file` example

Would be mildly useful to add one number like `27` == `0x1b`

website/docs/extensions/prove-and-verify.md
```diff
@@ -0,0 +1,46 @@
...
+
+Only packages defining the [executable target](../reference/targets.md/#Executable-target) can be proven.
+To prove the execution, you need to run the `scarb execute` command first, which will save execution information under the `target/execute/<target name>` directory.
+For each execution, a new output directory will be created, with consecutive number as names (e.g. `execution1`, `execution2`, ...).
+To clean the target directory, you can use the `scarb clean` command.
+
+To prove the execution, you can run:
+
+```shell
+scarb prove --execution_id <index of the relevant execution>
```
DelevoXDG, line 26:
```suggestion
scarb prove --execution-id <index of the relevant execution>
```

website/docs/extensions/prove-and-verify.md
```diff
@@ -0,0 +1,46 @@
...
+You can also run `scarb prove` with the `--execute` flag, which will run the `scarb execute` command automatically
+before proving the execution for you.
+
+The proof for the trace files inside the execution folder will be generated, and a `proof.json` file will be placed inside the execution directory.
+
+To verify the proof, you can run:
+
+```shell
+scarb verify <path to proof json file>
+```
```
DelevoXDG, line 38: I think we should put into separate section:
````suggestion
## Verifying Cairo proof

To verify the proof, you can run:

```shell
scarb verify <path to proof json file>
```
````

File: website/docs/extensions/prove-and-verify.md [hunk 5]
```diff
@@ -0,0 +1,46 @@
...
+To clean the target directory, you can use the `scarb clean` command.
+
+To prove the execution, you can run:
+
+```shell
+scarb prove --execution_id <index of the relevant execution>
+```
+
+You can also run `scarb prove` with the `--execute` flag, which will run the `scarb execute` command automatically
+before proving the execution for you.
```
DelevoXDG, line 30: Should we also document the following?
- Prover args (`--track-relations`, `--display-components`)
- Availability of execute args, with link to `execute` docs, when running `scarb prove` with `--execute` flag

website/docs/extensions/prove-and-verify.md
```diff
@@ -0,0 +1,46 @@
...
+```shell
+scarb verify <path to proof json file>
+```
+
+## Performance
+
+The `stwo-cairo` prover can highly benefit from platform specific optimizations, that are not available when Scarb
+is run from a precompiled binary.
+For the best performance, it is recommended to build `scarb-prove` and `scarb-verify` crates from source,
+with following compilation flags: `RUSTFLAGS="-C target-cpu=native -C opt-level=3" --features="std"`.
```
DelevoXDG, line 45: I know this is probably going to be used by experienced users only, maybe we should specify that updating local bin with newly built binaries is necessary?

website/docs/extensions/prove-and-verify.md
```diff
@@ -0,0 +1,46 @@
...
+For production use, it is recommended to use the `stwo-cairo` prover directly.
```
DelevoXDG, line 46: Just to be sure: does that mean manual builds as specified above, or not using `scarb prove` and `scarb verify` altogether? 🤔

website/docs/guides/creating-executable-package.md
```diff
@@ -0,0 +1,48 @@
...
+import {data as constants} from "../../constants.data";
+</script>
+
+## Defining an Executable Package
+
+Start a new Scarb project with `Scarb new <project_name>`, and add the following to your `Scarb.toml` file:
+
+1. Specify that this package should compile to a Cairo executable by adding `[[target.executable]]` to your toml file (note that `lib` or `starknet-contract` targets cannot be executed in this way)
+2. Add the `cairo_execute="{{ rel.stable.starknetPackageVersionReq }}"` plugin to your dependencies
+3. Disable gas usage (gas is only supported for `lib` or `starknet-contract` targets) by adding `enable-gas = false` under the `[cairo]` section in your toml.
```
DelevoXDG, line 12: minor adjustments:
```suggestion
Start a new Scarb project with `scarb new <project_name>`. 
In your `Scarb.toml` file:

1. Set the package to compile to a Cairo executable by adding `[[target.executable]]` (note that `lib` or `starknet-contract` targets cannot be executed in this way).
2. Add the `cairo_execute="{{ rel.stable.starknetPackageVersionReq }}"` plugin to your dependencies.
3. Disable gas usage by adding `enable-gas = false` under the `[cairo]` section (gas is only supported for `lib` or `starknet-contract` targets).
```

website/docs/guides/creating-executable-package.md
```diff
@@ -0,0 +1,48 @@
...
+
+You can now run:
+
+```
+scarb execute -p test_execute --print-program-output --arguments 5
+```
+
+Where `test_execute` is the name of the package with the executable target (as defined in our Scarb.toml manifest)
+
+The above command runs our executable function within the `test-execute` package and prints the program's output segment, which contains a success bit (0 for success) followed by the Cairo Serde of main’s output or the panic reason in case of a panic.
```
DelevoXDG, line 48: 1. I suppose we should also add info that execution output is saved.
2. Maybe we should link smth like "for more information on `scarb execute`" here?
3. Not sure info about panic should be there, and not in `scarb execute` docs.

website/docs/reference/targets.md
```diff
@@ -47,15 +51,113 @@ or textual Sierra, respectively.
...
+
+The executable target defines an artifact that can be executed with `scarb execute` command.
+This target should be used for packages that are meant to be executed rather than used as dependencies.
+See [Scarb execute](../extensions/execute.md) page for more details.
+Execution of the package created by the `scarb execute` command can be proved with `scarb prove` and the proof can be
+verified with `scarb verify`.
+See [proving and verifying execution](../extensions/prove-and-verify.md) page for more details.
+
+> [!TIP]
+> The executable target requires teh Cairo gas calculation to be disabled.
```
DelevoXDG, line 64:
```suggestion
> The executable target requires the Cairo gas calculation to be disabled.
```

DelevoXDG (commented):
Docs look good overall, left several minor typo fixes / adjustments

website/docs/extensions/execute.md
```diff
@@ -0,0 +1,74 @@
+<script setup>
+import { data as rel } from "../../github.data";
+</script>
+
+# Scarb execute
+
+The `scarb execute` command executes a function from a local package.
+Only packages defining the [executable target](../reference/targets#executable-target) can be executed.
+It does automatically compile the Cairo code within the package so using `scarb build` beforehand is not necessary.
+This automatically called build can be skipped with the `--no-build` flag.
```
FroyaTheHen, line 10:
```suggestion
The `scarb execute` command executes a function from a local package.
It does automatically compile the Cairo code within the package so using `scarb build` beforehand is not necessary.
You can skip the automatic build with the `--no-build` flag.

Only packages defining the [executable target](../reference/targets#executable-target) can be executed.
```
Would be a bit clearer imo, but no strong feelings.

website/docs/extensions/execute.md
```diff
@@ -0,0 +1,74 @@
...
+| `fn main(num: u8)`                     | 1                       |
+| `fn main(num1: u8, num2: u16)`         | 1,2                     |
+| `fn main(num1: u8, tuple: (u16, u16))` | 1,2,3                   |
+| `fn main(num1: u8, num2: u256)`        | 1,2,3                   |
+| `fn main(num1: u8, arr: Array<u8>)`    | 1,2,1,2                 |
+
+See the [documentation](https://docs.starknet.io/architecture-and-concepts/smart-contracts/serialization-of-cairo-types/) for more information about Cairo’s Serde.
+
+Note that when using `--arguments-file`, the expected input is an array of felts represented as hex string.
+For example, `1,2,3` in the above table becomes `["0x1", "0x2", "0x3"]`.
```
FroyaTheHen, line 74: I think we could have a separate section for command's flags. Something like for example [pip](https://pip.pypa.io/en/stable/cli/pip_freeze/#options:~:text=insensitive%20sorted%20order.-,Options,-%C2%B6). 
Please consider (not necessarily for this PR but generally).

website/docs/extensions/execute.md
```diff
@@ -0,0 +1,73 @@
+<script setup>
+import { data as rel } from "../../github.data";
+</script>
+
+# Scarb execute
+
+The `scarb execute` command executes a function from a local package.
+Only packages defining the [executable target](../reference/targets#executable-target) can be executed.
+It does automatically compile the cairo code within the package so using `scarb build` beforehand is not necessary.
+This automatically called build can be skipped with the `--no-build` flag.
```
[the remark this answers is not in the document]
maciektr: I don't want to push it too hard, if they are bothered with the build time, they will figure it out 🤔

website/docs/guides/creating-executable-package.md
maciektr: > Not sure info about panic should be there, and not in scarb execute docs.

It's only a guide, not reference, so I think it can combine tips about both 🤔

DelevoXDG: What I rather meant is that it seems like info about panic is more essential to docs of `scarb execute` itself, rather than merely a guide, that is providing brief overview.

website/docs/extensions/execute.md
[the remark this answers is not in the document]
DelevoXDG: I don't insist on pushing it too hard, but it's not explicitly mentioned under which conditions it's meaningful to use `--no-build`. Maybe we could have something like:

```
If `scarb build` or `scarb execute` has been previously used and the package hasn't changed since, this automatic build can be optionally skipped with the `--no-build` flag.
```

wdyt?

website/docs/guides/creating-executable-package.md
maciektr: Changed

website/docs/extensions/prove-and-verify.md
maciektr: > Prover args (--track-relations, --display-components)

I will skip them for now, as it's not clear to me what we would want to write about it here. We can link to some stwo docs if they have them.

website/docs/extensions/execute.md
maciektr: We could do that if we had some way to auto-generate those. Otherwise, I wouldn't really want to burden us with listing all flags and keeping them up to date 🤔

website/docs/extensions/prove-and-verify.md
DelevoXDG: They don't have any docs ATM, so I suppose you're right, these can be skipped

PR #2042, opened 2025-03-04
Represent a package in Proc Macro Server more reliably

integraledelebesgue:
## Changes
* Proc Macro Server now identifies workspace packages using a `CompilationUnitComponent` type which wraps the name and discriminator of the Cairo package
* This new API abstracts the implementation details about the package ids and names and reduces the reliance on implicit contracts

### DIFF
--- a/scarb/src/compiler/plugin/collection.rs
+++ b/scarb/src/compiler/plugin/collection.rs
@@ -3,6 +3,7 @@ use std::{collections::HashMap, sync::Arc};
 use anyhow::Result;
 use cairo_lang_semantic::{inline_macros::get_default_plugin_suite, plugin::PluginSuite};
 use itertools::Itertools;
+use scarb_proc_macro_server_types::scope::CompilationUnitComponent;
 
 use crate::{
     compiler::{CairoCompilationUnit, CompilationUnitComponentId, CompilationUnitDependency},
@@ -48,17 +49,12 @@ impl PluginsForComponents {
 // NOTE: Since this structure is used to handle JsonRPC requests, its keys have to be serialized to strings.
 //
 /// A container for Proc Macro Server to manage macros present in the analyzed workspace.
-///
-/// # Invariant
-/// Correct usage of this struct during proc macro server <-> LS communication
-/// relies on the implicit contract that keys of `macros_for_packages` are of form
-/// `PackageId.to_serialized_string()` which is always equal to
-/// `scarb_metadata::CompilationUnitComponentId.repr`.
 pub struct WorkspaceProcMacros {
-    /// A mapping of the form: `package (as a serialized [`PackageId`]) -> plugin`.
-    /// Contains IDs of all packages from the workspace, each mapped to a [`ProcMacroHostPlugin`] which contains
+    /// A mapping of the form: `cu_component (as a [`CompilationUnitComponent`]) -> plugin`.
+    /// Contains IDs of all components of all compilation units from the workspace,
+    /// each mapped to a [`ProcMacroHostPlugin`] which contains
     /// **all proc macro dependencies of the package** collected from **all compilation units it appears in**.
-    pub macros_for_packages: HashMap<String, Arc<ProcMacroHostPlugin>>,
+    pub macros_for_components: HashMap<CompilationUnitComponent, Arc<ProcMacroHostPlugin>>,
 }
 
 impl WorkspaceProcMacros {
@@ -71,18 +67,23 @@ impl WorkspaceProcMacros {
 
         for &unit in compilation_units {
             for (component_id, mut macro_instances) in collect_proc_macros(workspace, unit)? {
-                // Here we assume that CompilationUnitComponentId.repr == PackageId.to_serialized_string().
-                let key = component_id.package_id.to_serialized_string();
+                let component: CompilationUnitComponent = unit
+                    .components
+                    .iter()
+                    .find(|component| component.id == component_id)
+                    .expect("component should always exist")
+                    .into();
+
                 macros_for_components
-                    .entry(key)
+                    .entry(component)
                     .or_default()
                     .append(&mut macro_instances);
             }
         }
 
-        let macros_for_packages = macros_for_components
+        let macros_for_components = macros_for_components
             .into_iter()
-            .map(|(package_id, macro_instances)| {
+            .map(|(component, macro_instances)| {
                 let deduplicated_instances = macro_instances
                     .into_iter()
                     .unique_by(|instance| instance.package_id())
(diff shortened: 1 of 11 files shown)


utils/scarb-proc-macro-server-types/src/scope.rs
```diff
@@ -1,8 +1,37 @@
...
+}
+
+impl Display for SerializedPackageId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl SerializedPackageId {
+    pub fn name(&self) -> &str {
```
Draggu, line 24: Is there a reason why we keep it in serialized form instead of splitting this into 3 separate fields?

maciektr: Why do you want to do it this way, instead of fetching the package name from metadata via the package id?

maciektr: Shouldn't we send both pid and name in `DefinedMacrosInfo` then? 🤔

utils/scarb-test-support/src/proc_macro_server.rs
```diff
@@ -193,7 +194,7 @@ impl ProcMacroClient {
         // Usually, we can't discover the ID of the mock package used in test, so we extract it from the PMS response.
         let package_id = response
             .keys()
-            .find(|cu_id| cu_id.starts_with(package_name))
+            .find(|cu_id| cu_id.0.starts_with(package_name))
```
Draggu, line 197:
```suggestion
            .find(|cu_id| cu_id.name() == package_name)
```

Shouldn't it be this way now?

utils/scarb-proc-macro-server-types/src/scope.rs
Draggu: >For the same reason, the inner value is kept as it is without splitting into parts

Not convincing at all

piotmag769: >Shouldn't we send both pid and name

We can do that tbh, good idea. But remember that it has to be name from compilation unit (the one that is passed to the compiler), not package name, those may not be equal in the future

piotmag769 (commented):
I agree with breaking the interface for more soundness tbh

Draggu (approved):

piotmag769 (changes requested):

utils/scarb-proc-macro-server-types/src/scope.rs
```diff
@@ -1,8 +1,27 @@
 use serde::{Deserialize, Serialize};
 
+/// Representation of the Scarb package.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub struct Package {
+    /// Name of the Compilation Unit associated with the package.
+    pub name: String,
+    /// A string-serialized `PackageId` from the Scarb workspace
+    pub id: String,
```
piotmag769, line 9: It should be named `discriminator` and be described as this so we can decouple LS from this impl detail

integraledelebesgue: Done

scarb/src/compiler/plugin/collection.rs
```diff
@@ -72,7 +73,7 @@ impl WorkspaceProcMacros {
         for &unit in compilation_units {
             for (component_id, mut macro_instances) in collect_proc_macros(workspace, unit)? {
                 // Here we assume that CompilationUnitComponentId.repr == PackageId.to_serialized_string().
-                let key = component_id.package_id.to_serialized_string();
+                let key: Package = component_id.package_id.into();
```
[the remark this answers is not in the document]
integraledelebesgue: Done, as we discussed offline. I added information about the important invariants in comments and docs.

utils/scarb-proc-macro-server-types/src/scope.rs
```diff
@@ -1,8 +1,27 @@
 use serde::{Deserialize, Serialize};
 
+/// Representation of the Scarb package.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub struct Package {
+    /// Name of the Compilation Unit associated with the package.
```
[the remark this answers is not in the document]
integraledelebesgue: Done

utils/scarb-proc-macro-server-types/src/scope.rs
```diff
@@ -1,8 +1,40 @@
...
+/// # Invariants
+/// `name` and `discriminator` must refer to the same `CompilationUnitComponent` and must be identical to those from `scarb-metadata`.
+/// At the moment, they are obtained using `CompilationUnitComponent::cairo_package_name`
+/// and `CompilationUnitComponentId::to-discriminator`, respectively.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub struct Package {
+    /// Name of the `CompilationUnitComponent` associated with the package.
+    pub name: String,
+    /// A `CompilationUnitComponent` discriminator which represents
+    /// the package uniquely in the workspace.
```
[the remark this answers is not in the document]
integraledelebesgue: I corrected the docs to express that relationship

utils/scarb-proc-macro-server-types/src/scope.rs
```diff
@@ -1,8 +1,40 @@
...
 
+/// Representation of the Scarb package.
+///
+/// # Invariants
+/// 1. (`name`, `discriminator`) pair must represent the package uniquely in the workspace.
+/// 2. `name` and `discriminator` must refer to the same `CompilationUnitComponent` and must be identical to those from `scarb-metadata`.
+///    At the moment, they are obtained using `CompilationUnitComponent::cairo_package_name`
+///    and `CompilationUnitComponentId::to_discriminator`, respectively.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub struct Package {
```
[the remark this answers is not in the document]
integraledelebesgue: Agree, done

maciektr (approved):

utils/scarb-proc-macro-server-types/src/methods/defined_macros.rs
```diff
@@ -1,28 +1,31 @@
...
+/// Response structure containing a listed information
+/// about the macros used by packages from the workspace.
 ///
 /// # Invariant
-/// Correct usage of this struct during proc macro server <-> LS communication
-/// relies on the implicit contract that keys of `macros_by_package_id` are of form
-/// `PackageId.to_serialized_string()` which is always equal to
-/// `scarb_metadata::CompilationUnitComponentId.repr`.
+/// Each [`CompilationUnitComponentMacros`] in `macros_for_packages` should have
+/// an exclusive `package` field which identifies it in the response.
```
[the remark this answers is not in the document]
integraledelebesgue: Done

maciektr (approved):

PR #2071, opened 2025-03-12
Fix slack actions webhooks format

maciektr:


.github/workflows/check.yml
```diff
@@ -28,9 +28,8 @@ jobs:
     steps:
       - name: Notifying about failed tests!
         uses: slackapi/slack-github-action@v2.0.0
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_SCHEDULED_TESTS_FAILURE_WEBHOOK_URL }}
         with:
+          webhook: ${{ secrets.SLACK_SCHEDULED_TESTS_FAILURE_WEBHOOK_URL }}
```
DelevoXDG, line 32: From what I can gather, `webhook-type` also needs to be specified if `webhook` is set:
https://github.com/slackapi/slack-github-action/blob/1ebd7369aaf3a6d504deead6fff697aa960eea9c/src/config.js#L166

PR #2082, opened 2025-03-13
Proc macro: restore v1 macro compiler impl

maciektr:

PR #2083, opened 2025-03-13
Proc macro: read api version from toml file

maciektr:


maciektr:

- #1750
    - #2078
        - #2080
            - #2081
                - #2082
                    - #2083 ◀
                        - #2084
                            - #2085

scarb/src/compiler/plugin/mod.rs
```diff
@@ -25,6 +26,8 @@ pub struct CairoPluginProps {
     /// Mark this macro plugin as builtin.
     /// Builtin plugins are assumed to be available in `CairoPluginRepository` for the whole Scarb execution.
     pub builtin: bool,
+    /// Version of the API used by the plugin.
+    pub api: ProcMacroApiVersion,
```
DelevoXDG, line 30: https://github.com/software-mansion/scarb/pull/2078#discussion_r2001500870

DelevoXDG: Nit:
If we proceed with this approach though, I think:
```
api = "2"
```
(without prefix) would be more in line with how we typically specify versions.

maciektr:
Merged into dev at https://github.com/software-mansion/scarb/pull/1750

PR #2085, opened 2025-03-13
Proc macro: create versioned macro instance struct

maciektr:

PR #2084, opened 2025-03-13
Proc macro: move common struct to top-level proc macro module

maciektr:

PR #2086, opened 2025-03-13
Proc macro: use versioned proc macro host

maciektr:


maciektr:

- #1750
    - #2078
        - #2081
            - #2082
                - #2083
                    - #2084
                        - #2085
                            - #2086 ◀

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -0,0 +1,114 @@
...
+        instances: Vec<Arc<ProcMacroInstance>>,
+        api_version: ProcMacroApiVersion,
+    ) -> Result<Self> {
+        assert!(
+            instances
+                .iter()
+                .map(|instance| instance.api_version())
+                .all_equal(),
+            "all proc macro instances in a single host must use the same API version"
+        );
```
DelevoXDG, line 28: Should we specify also list api versions with corresponding package ids?

scarb/src/compiler/plugin/proc_macro/host.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    pub fn api_version(&self) -> ProcMacroApiVersion {
+        match self {
+            ProcMacroHostPlugin::V1(_) => ProcMacroApiVersion::V1,
+            ProcMacroHostPlugin::V2(_) => ProcMacroApiVersion::V2,
+        }
+    }
+}
+
+pub trait InstancesDeclared {
```
DelevoXDG, line 65: Nit:
```suggestion
pub trait DeclaredProcMacroInstances {
```

scarb/src/compiler/plugin/collection.rs
```diff
@@ -101,7 +102,10 @@ impl WorkspaceProcMacros {
     }
 
     /// Returns a [`ProcMacroHostPlugin`] assigned to the [`CompilationUnitComponent`].
```
DelevoXDG, line 104: This needs to be updated?

Arcticae (approved):

scarb/src/compiler/plugin/collection.rs
maciektr: Changed

scarb/src/compiler/plugin/proc_macro/host.rs
DelevoXDG: Isn't the same true for `"procedural macro `{}` using v1 api used in a context expecting v2 api"` errors? 🤔

maciektr: Yes, those could be panicking as well.

DelevoXDG: I'd rather provide minimal details package ids in both cases - just enough to simplify potential repro.

maciektr: Quite frankly, I feel like this is besides the point. It's more important to ensure we do maintain the invariant (e.g. when running tests), rather than use this for debugging. Esp. as that's run directly after `chunk_by` on the api version (https://github.com/software-mansion/scarb/blob/bc05b2502be42e6618b04da4c27f447c825b0a8b/scarb/src/compiler/plugin/collection.rs#L242)

DelevoXDG (approved):

maciektr:
Merged into dev at https://github.com/software-mansion/scarb/pull/1750

Issue #2053, opened 2025-03-10
Include `Cargo.lock` lockfile for procedural macro implementation when packaging


maciektr:
Since macros are compiled into separate binary artifacts and not used as a Rust dependency of another package, it would be beneficial to include the Cargo lockfile to guarantee deterministic builds / deterministic dependency resolution. 

Make sure Scarb will use the lockfile when building macro. Should we call Cargo with  `--locked`/`--frozen`? 

Action point from Foundry post mortem doc https://docs.google.com/document/d/16JkhBjAOX9vSTnrG5BNJmlCJLnwSE58mv472rpLc6D8/edit?usp=sharing
CC @cptartur

PR #2092, opened 2025-03-17
Include `Cargo.lock` when packaging proc macros 

DelevoXDG:
Closes [#2053](https://github.com/software-mansion/scarb/issues/2053)


scarb/tests/package.rs
DelevoXDG, line 1: The original issue mentioned:

> Make sure Scarb will use the lockfile when building macro. Should we call Cargo with --locked/--frozen?

Locally, we just use `cargo build`, which will use Cargo.lock by default.

We can write a more complex test that will check that locked version is used when building via `cargo build`, but I wasn't sure it's worth it to test what is essentially Cargo functionality.

We can try doing that via [cargo-test-support](https://github.com/rust-lang/cargo/blob/master/crates/cargo-test-support/src/registry.rs) local registry, or something like [cargo-local-registry](https://crates.io/crates/cargo-local-registry) crate, but I'm not sure it's worth the effort.

Thoughts?

scarb/src/ops/package.rs
DelevoXDG, line 1: I'm also wondering:

Should we add check on packaging that emits warning / error that prompts/forces user to specify exact `"=x.y.z"` requirement for all dependencies in Cargo manifest?

maciektr: > Should we add check on packaging that emits warning / error that prompts/forces user to specify exact "=x.y.z" requirement for all dependencies in Cargo manifest?

Nope! Not pinning versions is totally fine, Cargo.lock should be used to ensure deterministic builds instead.

scarb/src/ops/package.rs
```diff
@@ -329,6 +329,12 @@ fn prepare_archive_recipe(
             path: ORIGINAL_CARGO_MANIFEST_FILE_NAME.into(),
             contents: ArchiveFileContents::OnDisk(pkg.root().join(CARGO_MANIFEST_FILE_NAME)),
         });
+
+        // Add original Cargo.lock file.
+        recipe.push(ArchiveFile {
+            path: CARGO_LOCK_FILE_NAME.into(),
+            contents: ArchiveFileContents::OnDisk(pkg.root().join(CARGO_LOCK_FILE_NAME)),
```
maciektr, line 336: What if macro package is part of a Cargo workspace? 🤔

scarb/src/ops/package.rs
maciektr: > Cargo.lock will be used for deterministic builds, but isn't it better for package publisher to also be knowledgeable about specific version that will end up being used? 🤔

He is, through the lockfile. We don't pin dep versions in Cargo.toml in Scarb, do we?

maciektr: If you use a crate as dep, you will not use the lockfile of the dependency - you resolve versions to use in your project. (Analog: using serde as dep in your Cargo package)
We need the lockfiles when the package is compiled as is, not as part of a larger build. (Analog: cargo install mdbook. Mdbook is not your dep, but you need to compile it.)

scarb/src/ops/package.rs
DelevoXDG: Great catch, fixed it to extract it from `.crate` created by `cargo package` instead

scarb/tests/package.rs
DelevoXDG: > I just want to make sure that the lockfile is in a valid location so it's used when we compile macro in our cache.

In packaging tests we test that all necessary files are present inside the tarball (now and prior to my changes). 
It's assumed the entirety of the tarball will be extracted and loaded in cache, since it's the same mechanism for extracting all the other files we previously packaged.
```rs
impl RegistrySource<'_> {
    /// Turn the downloaded `.tar.zst` tarball into a [`Package`].
    ///
    /// This method extracts the tarball into cache directory, and then loads it using
    /// suitably configured [`PathSource`].
    async fn load_package(
        &self,
        id: PackageId,
        archive: FileLockGuard,
        checksum: Checksum,
    ) -> Result<Package> {
        self.config
            .ui()
            .verbose(Status::new("Unpacking", &[id.to](http://id.to/)_string()));
        // `extract` drops the archive internally and thus handles file unlocking.
        let path = self.package_sources.extract(id, archive).await?;
        let path_source = PathSource::recursive_at(&path, self.source_id, self.config);
        let mut package = path_source.download(id).await?;
        package.manifest_mut().summary.set_checksum(checksum);
        Ok(package)
    }
}
```
Not sure we need anything beyond that 🤔

(FYI, I checked full process locally just to be sure. Just not sure we want to replicate the same end-to-end setup that would require setting up some local registry in our tests)

scarb/src/ops/package.rs
```diff
@@ -329,6 +329,19 @@ fn prepare_archive_recipe(
...
+
+        // Add generated Cargo.lock file.
+        recipe.push(ArchiveFile {
+            path: CARGO_LOCKFILE_FILE_NAME.into(),
+            contents: ArchiveFileContents::OnDisk(
+                pkg.target_path(ws.config())
+                    .into_child("package")
+                    .into_child(&crate_archive_basename)
+                    .into_child(CARGO_LOCKFILE_FILE_NAME)
+                    .path_unchecked()
```
maciektr, line 341: From https://doc.rust-lang.org/cargo/commands/cargo-package.html#description 
```
Cargo.lock is automatically included if the package contains an executable binary or example target. [cargo-install(1)](https://doc.rust-lang.org/cargo/commands/cargo-install.html) will use the packaged lock file if the --locked flag is used.
```
So will it be excluded automatically for `cdylib` or not? 🤔

DelevoXDG: It's included (we have a test to check it 😄)

scarb/src/compiler/plugin/proc_macro/compilation.rs
```diff
@@ -282,6 +289,9 @@ impl<'c> From<CargoCommand<'c>> for Command {
             CargoAction::Check => cmd.arg("check"),
             CargoAction::Package(_) => cmd.arg("package"),
         };
+        if args.locked {
+            cmd.arg("--locked");
+        }
```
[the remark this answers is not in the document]
DelevoXDG: You're right, my change stemmed from the misconception that `--locked` prevents updating the lockfile, if such need arises for whatever the reason . 

Upon further research, it seems as though under no circumstances will `--locked` force Cargo to skip updates gracefully, the only thing it will do is error out due to inability to update the lockfile. 

So I suppose allowing Cargo to update the lockfile if necessary - as a fallback - is better.

Reverted.

maciektr (approved):

scarb/src/ops/package.rs
maciektr: For future reference: https://github.com/rust-lang/cargo/issues/13447

Issue #1535, opened 2024-08-23
When defining a package as a internal dependency within workspace, dev-dependencies from both get included when compiling


cptartur:
### Problem

When defining a package as a internal dependency within workspace, dev-dependencies from both get included when compiling. This causes problems when using both cairo-test and starknet foundry in a single workspace.

### Version

2.7.0

ShantelPeters:
Hi @cptartur please can I be assigned to this issue?

ShantelPeters:
Okay sir @maciektr

PR #2099, opened 2025-03-18
Fix dev-deps propagation of compiler plugins for ws members

### LINKED ISSUE #2118: Fix dev-deps propagation of compiler plugins for ws members

maciektr:
fixes #2118
This problem is analog to the https://github.com/software-mansion/scarb/issues/1535 (https://github.com/software-mansion/scarb/pull/1542). 
If a package (A) that is a ws member depends on another ws member (B), and (B) depends on some cairo plugin (C) as dev dependency, test CompilationUnit of (A) will include (C) as dependency of (B), although dev dependencies of (B) should not be propagated to the test compilation of (A) (both Cairo and Plugin dependencies).   

- **Add failing test for propagation of dev dep plugins from members of ws**
- **Do not propagate plugins that are dev dependencies of ws members to packages that depend on these ws members**

### DIFF
--- a/scarb/tests/metadata.rs
+++ b/scarb/tests/metadata.rs
@@ -46,6 +46,27 @@ fn units_and_components(meta: Metadata) -> BTreeMap<String, Vec<String>> {
         .collect::<BTreeMap<_, _>>()
 }
 
+fn units_and_plugins(meta: Metadata) -> BTreeMap<String, Vec<String>> {
+    meta.compilation_units
+        .iter()
+        .map(|cu| {
+            (
+                cu.target.name.clone(),
+                cu.cairo_plugins
+                    .iter()
+                    .map(|c| {
+                        meta.packages
+                            .iter()
+                            .find(|p| p.id == c.package)
+                            .map(|p| p.name.clone())
+                            .unwrap()
+                    })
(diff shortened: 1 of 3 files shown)


scarb/tests/metadata.rs
```diff
@@ -398,6 +419,96 @@ fn dev_deps_are_not_propagated_for_ws_members() {
...
+            ("core".to_string(), vec!["core".to_string()]),
+            (
+                "dep2".to_string(),
+                vec![
+                    "core".to_string(),
+                    "dep2".to_string(),
+                    // Note that `cairo_test` is indeed a dev dependency of `dep2`,
+                    // but it's not propagated to the unit tests of `x`.
+                    // Only dev dependencies of the main component should be enabled.
+                    // "cairo_test".to_string()
```
DelevoXDG, line 501: This seems good, but I wonder if we should make `dep2` dev-depend on some **non-builtin** plugin, just so there's no confusion about builtins being handled differently.

DelevoXDG (approved):
A very minor comment, otherwise looks good

scarb/tests/metadata.rs
maciektr: Added

Issue #2094, opened 2025-03-18
Scarb doc landing page


maciektr:
Docs should have some landing page. 
Top-level docstring of package entrypoint shouldn't be hidden in a module entry (see https://docs.swmansion.com/scarb/corelib/core.html). 
Should be inspired by docs.rs, e.g. https://docs.rs/tokio/latest/tokio/index.html

Issue #1762, opened 2024-11-25
Parse doc comments with correct newline persistance


wawel37:
As for now, scarb doc ignores all newline (beside code blocks). It is a problem because some of the markdown elements require to be parsed with newlines being persisted (for example a list).

AC:
- Make scarb doc parsing comments persist newlines within a markdown elements, that it should (codeblocks, lists and any other).

maciektr:
Related issue: https://github.com/software-mansion/scarb/issues/1795

PR #2121, opened 2025-03-25
feat: doc landing page

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2094

- Due to formatting restrictions in SUMMARY.md the index page is a separate file. 
- However, creating documentation for each item and creating short version of the same items is indeed redundant. I did try to use a HashMap for storing the documentation strings data and retrieve it from the list if exists insead of generating a new one.  After couple of runs compared time of executions of both versions for Cairo _corelib_ the highest difference in execution time was around 1 second (6 seconds for no-hash-map approach, 5 for hash-map approach). I propose the no-hash-map version as the final solution as the time difference is in my opinion not significant enough and no extra memory allocation is needed. Please, let me know if you have different opinion. Keep in mind that measuring the timing could be slightly different at this point as I did the tests quite early developing this code. 
- If you feel like giving it a try at you local environment be my guest. If not, you can see demo below. 
https://github.com/user-attachments/assets/78d52898-1f3c-485f-bd0e-09b9a9c41962


extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -126,49 +158,60 @@ impl MarkdownDocItem for Module {
...
-            header_level + 1,
+            context,
+            false,
         )?;
-        markdown += &generate_markdown_list_for_top_level_subitems(
+        markdown += &generate_markdown_table_summary_for_top_level_subitems(
             &self.extern_functions.iter().collect_vec(),
-            header_level + 1,
+            context,
+            false,
```
FroyaTheHen, line 214: imo, for consistency sake it's good to have all item types in tables instead of modules-in-tables the-rest-in-lists but I do hesitate and don't have any strong feelings on that. 
I guess that quite subjective and subjected to the end-user perspective. Any opinion on that?

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -264,22 +307,40 @@ pub fn mark_duplicated_item_with_relative_path<'a, T: TopLevelMarkdownDocItem +
         .collect::<Vec<_>>()
 }
 
-pub fn generate_markdown_list_for_top_level_subitems<T: TopLevelMarkdownDocItem>(
+pub fn generate_markdown_table_summary_for_top_level_subitems<T: TopLevelMarkdownDocItem>(
     subitems: &[&T],
-    header_level: usize,
+    context: &MarkdownGenerationContext,
+    link_to_index_summary: bool,
```
FroyaTheHen, line 313: the idea is to avoid linking to the all module page `modules.md` from any submodule because that would be misleading.  However, in the current proposition we do link only from the all-module-page back to the very same all-module-page. So maybe it doesn't make any sense to link that section at all ever??

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -59,6 +61,36 @@ pub trait MarkdownDocItem: DocItem {
...
+    fn get_short_documentation(&self, context: &MarkdownGenerationContext) -> String {
+        let mut short_doc_buff = String::new();
+
+        if let Some(tokens) = self.doc() {
+            for token in tokens.iter() {
+                match token {
+                    DocumentationCommentToken::Content(content) => {
+                        if SHORT_DOCUMENTATION_AVOID_PREFIXES
+                            .iter()
+                            .any(|prefix| content.starts_with(prefix))
```
maciektr, line 73: Won't `"\n\n"` be separated into two separate tokens though? 🤔 Taking it from https://github.com/FroyaTheHen/cairo/blob/0b867e8b17690f8963f5722bb3dab143a7e66760/crates/cairo-lang-doc/src/tests/test-data/lists_formatting.txt#L131-L132

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -59,6 +61,36 @@ pub trait MarkdownDocItem: DocItem {
...
+                        }
+                    }
+                    DocumentationCommentToken::Link(link) => {
+                        let file_path = context.resolve_markdown_file_path_from_link(link);
+                        let link_formatted = format!("[{}]({})", link.label.clone(), file_path);
+                        short_doc_buff.push_str(link_formatted.replace("\n", " ").as_str());
+                    }
+                };
+                if short_doc_buff.len() >= SHORT_DOCUMENTATION_LEN {
+                    return short_doc_buff;
```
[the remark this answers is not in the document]
FroyaTheHen: I would rather attach the whole line. This way we can be sure that no word is ever split and the chances that shorted text represents some integrity are higher. `SHORT_DOCUMENTATION_LEN` is a made up number and moving it up by 50 or even 100 characters wouldn't probably make any significant difference, so I think about it more as of a range than a limit itself. 

However, I see the potential of improving this part in such a manner 👇              
```
if short_doc_buff.len() + formatted_token_to_be_pushed.len() >= SHORT_DOCUMENTATION_LEN {
                return short_doc_buff;
            } else {
                short_doc_buff.push_str(formatted_token_to_be_pushed);
            }
```
so we can be sure that if there is an enormous token it wouldn't get into short doc and spoil the visuals of the generated table.

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -264,22 +296,30 @@ pub fn mark_duplicated_item_with_relative_path<'a, T: TopLevelMarkdownDocItem +
...
+        writeln!(&mut markdown, "\n{}\n ---\n| | |\n|:---|:---|", T::HEADER,)?;
 
-        writeln!(&mut markdown, "{header} {}\n", T::HEADER)?;
         let items_with_relative_path = mark_duplicated_item_with_relative_path(subitems);
         for (item, relative_path) in items_with_relative_path {
+            let item_doc = item.get_short_documentation(context);
+            let expand = {
+                match relative_path.clone() {
+                    Some(_) => format!("[ ...](./{})", item.filename()),
+                    None => format!("[...](./{})", item.filename()),
```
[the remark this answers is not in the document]
FroyaTheHen: yeah, I think i mistakenly left it here xD I'll double check 🙌

extensions/scarb-doc/tests/data/hello_world_no_features/src/hello_world-tests.md
```diff
@@ -4,7 +4,9 @@ Tests module
...
 
-## Free functions
-
-- [it_works](./hello_world-tests-it_works.md)
 
+Free functions
+ ---
+| | |
+|:---|:---|
+| [it_works](./hello_world-tests-it_works.md) | Reallyworks.[...](./hello_world-tests-it_works.md) |
```
FroyaTheHen, line 12: just letting you know that I'm aware that a space is missing :) 
👀 https://github.com/software-mansion/scarb/issues/1762#issuecomment-2768361397

extensions/scarb-doc/src/docs_generation/markdown.rs
```diff
@@ -26,6 +26,12 @@ const BASE_HEADER_LEVEL: usize = 1;
 const SOURCE_DIRECTORY: &str = "src";
 const BOOK_TOML_FILENAME: &str = "book.toml";
 pub const SUMMARY_FILENAME: &str = "SUMMARY.md";
+const SHORT_DOCUMENTATION_LEN: usize = 200;
```
FroyaTheHen, line 29: changed from 100 to 200 because it seemed too short when `SHORT_DOCUMENTATION_LEN` became a strict top limit in `fn get_short_documentation`

extensions/scarb-doc/tests/data/hello_world_no_features/src/hello_world-tests.md
FroyaTheHen: I was wrong. The problem is in the changes here. I'll fix it.

maciektr (approved):

maciektr:
Looks great!

PR #2125, opened 2025-03-26
feat: doc sematically sorted modules, interactive path

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2096

- Created nested modules navbar, but i did leave the rest of the elements (free functions, structs, traits etc) as they were. The reason is summary file format restrictions.  Prefix chapters cannot be nested - because of that if we did create a nested structure the elements contained by a module couldn't be distinct by type and we'd end up with a confusing list of mixed types items. 
Imo, the presented way is good enough but there is a possibility of using some custom backend that handles nested prefixed chapters (I assume one exists). So let me know if you'd like to see it instead and I will research our options. 
- added links to `Fully qualified path`


maciektr (approved):
💪

PR #2140, opened 2025-03-31
Add mdbook builder as extension

maciektr:


maciektr:

- #2140 ◀
    - #2141
        - #2142
            - #2143

extensions/scarb-mdbook/src/main.rs
```diff
@@ -0,0 +1,46 @@
...
+use std::process::ExitCode;
+
+#[derive(Parser, Debug)]
+#[command(version, about, long_about = None)]
+struct Args {
+    /// Path to book source directory.
+    #[arg(long)]
+    pub input: Utf8PathBuf,
+    /// Path to book output directory.
+    #[arg(long)]
```
FroyaTheHen, line 16:
```suggestion
    #[arg(long, default_value_t = Utf8PathBuf::from("./book/"))]
```

extensions/scarb-mdbook/src/main.rs
```diff
@@ -0,0 +1,46 @@
...
+        }
+    }
+}
+
+fn main_inner(args: &Args, _ui: Ui) -> Result<()> {
+    let mut book = MDBook::load(args.input.clone())?;
+    book.config.build.build_dir = args.output.clone().into();
+    book.build()?;
+    Ok(())
+}
```
FroyaTheHen, line 46: Ca you please add a simple test? 🙏

extensions/scarb-mdbook/src/main.rs
FroyaTheHen: I see you added the default `/book/` output dir later in commits in `scarb doc --build` so the comment above is not as relevant. I think I'd rather still have a default output dir here but no strong feelings 🤔 so I'll just leave it for you to consider.

maciektr:
Replaced with https://github.com/software-mansion/scarb/pull/2173
(I messed up my stack tracking, sorry 😓 )

PR #2142, opened 2025-03-31
Mark code blocks as Cairo not Rust

maciektr:
Mark code blocks as Cairo not Rust

Add option to build straight from scarb-doc

### DIFF
--- a/extensions/scarb-doc/src/main.rs
+++ b/extensions/scarb-doc/src/main.rs
@@ -1,11 +1,11 @@
-use anyhow::{Context, Result};
+use anyhow::{Context, Result, ensure};
 use clap::Parser;
 use scarb_doc::docs_generation::markdown::MarkdownContent;
 use scarb_doc::errors::MetadataCommandError;
 use scarb_doc::metadata::get_target_dir;
 use std::process::ExitCode;
 
-use scarb_metadata::MetadataCommand;
+use scarb_metadata::{MetadataCommand, ScarbCommand};
 use scarb_ui::args::{PackagesFilter, ToEnvVars, VerbositySpec};
 
 use scarb_doc::generate_packages_information;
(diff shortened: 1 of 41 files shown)


extensions/scarb-doc/src/main.rs
```diff
@@ -100,6 +108,22 @@ fn main_inner(args: Args, ui: Ui) -> Result<()> {
...
+                if args.build {
+                    let build_output_dir = pkg_output_dir.join("book");
+                    ScarbCommand::new()
+                        .arg("mdbook")
+                        .arg("--input")
+                        .arg(pkg_output_dir)
+                        .arg("--output")
+                        .arg(build_output_dir.clone())
+                        .env("SCARB_UI_VERBOSITY", ui.verbosity().to_string())
+                        .run()?;
```
FroyaTheHen, line 120: did you run it without issues in your local env? I am trying to do that. Yet, command exits with `error: no such command: mdbook`. However, scarb-`metadata/src/command/scarb_command.rs` `cmd.status()` is  `Ok(ExitStatus(unix_wait_status(256))) ` so if i change the line` if cmd.status()?.success() {` to `if cmd.status().is_ok() {` it builds the book as expected. Any idea what could be wrong with my local env or if is it not env specific?

maciektr: Hmm, how do you compile it? Sounds like scarb-mdbook has not been compiled. Can you try `cargo build --workspace --exclude scarb-prove --exclude scarb-verify`? Note there is no way in Cargo to declare that scarb-doc depends on scarb-mdbook binary (neither in tests nor in binary build https://github.com/rust-lang/cargo/issues/9096).

FroyaTheHen: Ok, it worked 👌  Setting `$SCARB` env did the job :)

FroyaTheHen (approved):

PR #2147, opened 2025-04-01
Inform about the source packages for proc macros in PMS

integraledelebesgue:
## Changes
* `DefinedMacrosResponse` from `proc-macro-server-types` contains an additional field `debug_info: DebugInfo`.
* `DebugInfo` contains a field `source_packages` which informs about the proc macro packages loaded by the server.

### DIFF
--- a/utils/scarb-proc-macro-server-types/src/methods/defined_macros.rs
+++ b/utils/scarb-proc-macro-server-types/src/methods/defined_macros.rs
@@ -34,6 +34,16 @@ pub struct CompilationUnitComponentMacros {
     pub derives: Vec<String>,
     /// List of executable attributes available.
     pub executables: Vec<String>,
+    /// Additional debug information.
+    pub debug_info: DebugInfo,
+}
+
+/// Stores extra information about the macros managed by the server.
+/// Used for debugging on the LS side.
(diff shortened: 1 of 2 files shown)


maciektr (commented):
Why not just use info from metadata? 🤔

maciektr (approved):

utils/scarb-proc-macro-server-types/src/methods/defined_macros.rs
```diff
@@ -26,6 +26,9 @@ pub struct CompilationUnitComponentMacros {
     /// It should identify [`CompilationUnitComponentMacros`]
     /// uniquely in the [`DefinedMacrosResponse`].
     pub component: CompilationUnitComponent,
+    /// Sources of macros used by the component,
+    /// given as `PackageId::to_serialized_string`.
+    pub macro_packages: Vec<String>,
```
Draggu, line 31: Maybe wrap it into `DebugInfo` struct or something

Draggu (approved):

utils/scarb-proc-macro-server-types/src/methods/defined_macros.rs
integraledelebesgue: Done

piotmag769 (commented):
👍

PR #2148, opened 2025-04-01
Allow warnings for lint cli / respect allow-warnings from manifest

wawel37:


maciektr (commented):
Why not just rely on https://docs.swmansion.com/scarb/docs/reference/manifest.html#allow-warnings ? 🤔

wawel37:
> Why not just rely on https://docs.swmansion.com/scarb/docs/reference/manifest.html#allow-warnings ? 🤔

Do you mean the `--deny-warnings` additional arg? Becuase we do rely on the `allow-warnings` in manifest but the optional arg is added as the forge team requested it and it will be probably helpful during ma'at integration

maciektr (approved):

PR #2151, opened 2025-04-03
Fix casing in derive macro lookup

maciektr:
- **Add test case**
- **Fix casing in derive macro lookup**


scarb/src/compiler/plugin/proc_macro/instance.rs
```diff
@@ -97,7 +105,16 @@ impl ProcMacroInstance {
         self.get_expansions()
             .iter()
             .filter(|e| e.kind == ExpansionKind::Derive)
-            .map(|e| e.name.clone())
+            .map(|e| e.cairo_name.clone())
+            .map(Into::into)
+            .collect()
+    }
+
+    pub fn declared_derives_snake_case(&self) -> Vec<String> {
```
Arcticae, line 113: why do we still need to have something `_snake_case` with those changes? aren't the abstractions enough to avoid this?

Arcticae (approved):

PR #2186, opened 2025-04-16
Run proc macro expansion before executable plugin

maciektr:
Expanding a proc macro attribute on a function creates a virtual file with the expanded implementation and removes the old node. The executable plugin generates a new function - wrapper, which calls the original one. If we run executable plugin before proc macros, using a proc macro attribute on executable function would cause it to be duplicated (and thus fail with error).


maciektr:

- #2186 ◀
    - #2187

DelevoXDG (approved):

PR #2192, opened 2025-04-18
Cli flags for disabling proc macros

maciektr:
- **Allow disabling prebuilt macros with cli arg**
- **Compile prebuilt example only once**
- **Allow disabling all macros with cli flag**

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -100,6 +100,10 @@ pub struct ScarbArgs {
     #[arg(long, env = "SCARB_NO_PROC_MACROS")]
     pub no_proc_macros: bool,
 
+    /// Do not load any prebuilt procedural macros.
+    #[arg(long, env = "SCARB_NO_PREBUILT_PROC_MACROS")]
+    pub no_prebuilt_proc_macros: bool,
+
     /// Specify the profile to use.
     #[command(flatten)]
     pub profile_spec: ProfileSpec,
@@ -183,7 +187,7 @@ pub enum Command {
     ")]
     Package(PackageArgs),
     /// Start the proc macro server.
-    ProcMacroServer(ProcMacroServerArgs),
+    ProcMacroServer,
     /// Upload a package to the registry.
     #[command(after_help = "\
         This command will create distributable, compressed `.tar.zst` archive containing source \
@@ -245,10 +249,6 @@ pub struct BuildArgs {
     /// Do not error on `cairo-version` mismatch.
     #[arg(long, env = "SCARB_IGNORE_CAIRO_VERSION")]
     pub ignore_cairo_version: bool,
-
-    /// Do not load any prebuilt procedural macros.
-    #[arg(long, env = "SCARB_NO_PREBUILT_PROC_MACROS")]
-    pub no_prebuilt_proc_macros: bool,
 }
 
 /// Arguments accepted by the `expand` command.
@@ -280,10 +280,6 @@ pub struct ExpandArgs {
     /// Emit the expanded file to stdout
     #[arg(short, long)]
     pub emit: Option<EmitTarget>,
-
-    /// Do not load any prebuilt procedural macros.
-    #[arg(long, env = "SCARB_NO_PREBUILT_PROC_MACROS")]
-    pub no_prebuilt_proc_macros: bool,
 }
 
 /// Arguments accepted by the `run` command.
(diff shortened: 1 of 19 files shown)


scarb/src/bin/scarb/args.rs
```diff
@@ -96,6 +96,10 @@ pub struct ScarbArgs {
     )]
     pub target_dir: Option<Utf8PathBuf>,
 
+    /// Do not load any procedural macros.
+    #[arg(long, env = "SCARB_NO_PROC_MACROS")]
+    pub no_proc_macros: bool,
```
DelevoXDG, line 101: This is nitpick, but I feel your typical user user might get confused as to *where* the flags should be added, at Scarb level or at command level 
```
scarb --no-proc-macros build --no-prebuilt-proc-macros
```
I wonder if it would be helpful to make both flags global:
```
scarb --no-proc-macros --no-prebuilt-proc-macros build
```

We could then also consider one enum flag, since we have 4 possible combinations of flags for 3 states (`--no-proc-macros` implies `--no-prebuilt-proc-macros`  🤔 
```
--disable-proc-macros <none(default)/prebuilt/all>
```

scarb/src/compiler/plugin/proc_macro/repository.rs
```diff
@@ -1,25 +1,34 @@
...
+        Self {
+            macros: Default::default(),
+            load_proc_macros,
+        }
+    }
+
     /// Returns the [`ProcMacroInstance`] representing the procedural macros defined in the [`Package`].
     /// Loads the underlying shared library if it has not been loaded yet.
     pub fn get_or_load(&self, package: Package, config: &Config) -> Result<Arc<ProcMacroInstance>> {
+        ensure!(self.load_proc_macros, "procedural macros are disallowed");
```
DelevoXDG, line 30: Maybe we should add message that they're disabled *by the user* via a `--no-proc-macros` flag?

Flag is rather hard to miss 😅 , but a leftover env var might be harder to track down based on the current message

scarb/src/ops/package.rs
```diff
@@ -47,6 +47,7 @@ pub struct PackageOpts {
     pub check_metadata: bool,
     pub features: ops::FeaturesOpts,
     pub ignore_cairo_version: bool,
+    pub load_prebuilt_macros: bool,
```
DelevoXDG, line 50: I understand these flags only disable loading proc macros, but I wonder if it's not confusing for the user that the macro is getting packaged with prebuilts even if flag called `--no-prebuilt-proc-macros` is passed 😄

DelevoXDG (commented):
LGTM overall, just some minor UX suggestions

scarb/src/bin/scarb/args.rs
DelevoXDG: Either approach would sort of address https://github.com/software-mansion/scarb/pull/2192#discussion_r2059885157, as the user would not be passing `--no-prebuilt-proc-macros` explicitly as an argument of `package/publish`

scarb/src/compiler/plugin/proc_macro/repository.rs
maciektr: Changed

scarb/src/bin/scarb/args.rs
maciektr: Good point, changed

scarb/src/compiler/plugin/proc_macro/repository.rs
```diff
@@ -1,25 +1,37 @@
...
+            load_proc_macros,
+        }
+    }
+
     /// Returns the [`ProcMacroInstance`] representing the procedural macros defined in the [`Package`].
     /// Loads the underlying shared library if it has not been loaded yet.
     pub fn get_or_load(&self, package: Package, config: &Config) -> Result<Arc<ProcMacroInstance>> {
+        ensure!(
+            self.load_proc_macros,
+            "procedural macros are disallowed with `--no-proc-macros` flag"
```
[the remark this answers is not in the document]
maciektr: I don't want to overdo it 😅 
Printing env vars like this seems more confusing than helpful to me - they are meant to be used in integrations with other tools mostly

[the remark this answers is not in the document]
DelevoXDG: This is minor so I'm not going to block this, but I feel like it could be useful to see *accurate* info on fail on CI as well. The current statement will be inaccurate in case of this being modified by env var rather than the flag. Sure, the user will likely be able to investigate it, but I feel like it's better the full and accurate info, even if it's slightly verbose.
> Printing env vars like this seems more confusing than helpful to me 
the user that has this set with flag probably won't read past the info about the flag

With that said, I leave the final decision to you 👍🏻

DelevoXDG (approved):

scarb/src/compiler/plugin/proc_macro/repository.rs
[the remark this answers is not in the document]
maciektr: I'll merge it then, as I still feel that a) it is correct - env vars are an alternative way of using the interface which we define for a lot of flags that we have b) it's mostly aimed at integrating scarb into other tools.

PR #2208, opened 2025-04-24
Fix code mappings for inner attr expansion

maciektr:
commit-id:f713f23c


scarb/src/compiler/plugin/proc_macro/v2/host/attribute/inner_attribute.rs
```diff
@@ -72,10 +71,14 @@ impl<'a> InnerAttrExpansionContext<'a> {
...
         self,
-        expanded: String,
-        code_mappings: Vec<CodeMapping>,
+        item_builder: PatchBuilder<'_>,
         attr_names: Vec<SmolStr>,
     ) -> PluginResult {
+        let (expanded, mut code_mappings) = item_builder.build();
+        // PatchBuilder::build() adds additional mapping at the end,
+        // which wraps the whole outputted code.
+        // We remove it, so we can properly translate locations spanning multiple token spans.
```
piotmag769, line 80: :o

piotmag769 (approved):

PR #2212, opened 2025-04-28
Introduce --target-names arg for scarb lint

wawel37:
resolves https://github.com/software-mansion/cairo-lint/issues/176

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -539,6 +539,10 @@ pub struct LintArgs {
     #[command(flatten)]
     pub packages_filter: PackagesFilter,
 
+    /// Comma separated list of target names to compile.
+    #[arg(long, value_delimiter = ',')]
+    pub target_names: Vec<String>,
+
     /// Should lint the tests.
     #[arg(short, long, default_value_t = false)]
     pub test: bool,
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -540,7 +540,7 @@ pub struct LintArgs {
     pub packages_filter: PackagesFilter,
 
     /// Comma separated list of target names to compile.
-    #[arg(long, value_delimiter = ',')]
+    #[arg(long, value_delimiter = ',', env = "SCARB_TARGET_NAMES")]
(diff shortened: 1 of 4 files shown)


scarb/src/bin/scarb/args.rs
```diff
@@ -539,6 +539,10 @@ pub struct LintArgs {
     #[command(flatten)]
     pub packages_filter: PackagesFilter,
 
+    /// Comma separated list of target names to compile.
+    #[arg(long, value_delimiter = ',')]
```
maciektr, line 543:
```suggestion
    #[arg(long, value_delimiter = ',', env = "SCARB_TARGET_NAMES")]
```

scarb/tests/lint.rs
```diff
@@ -518,3 +518,65 @@ fn test_missing_feature() {
             note: to use features, you need to define [features] section in Scarb.toml
         "#});
 }
+
+#[test]
+fn lint_selected_targets() {
```
maciektr, line 523: I don't understand why `scarb lint --workspace` does not lint all packages in the workspace anyway? 🤔 This sounds like what should happen without the need to explicitly list target names.

maciektr (changes requested):

scarb/tests/lint.rs
wawel37: Tbh it sounds reasonable to me, that passing `--workspace` alongside with `--target-names` will lint only the targets (including the whole workspace), that were specified in the `--target-names`. Also, I'm not sure, but as I looked roughly through the `build` logic, it acts the same. It first filters out packages based on `pakcage_filter` arg (which also includes `--workspace` inside) https://github.com/software-mansion/scarb/blob/5dc5cceaa26d103fea70f39153ce60ba5f7358e5/scarb/src/bin/scarb/commands/build.rs#L11, and later filters out the targets here: https://github.com/software-mansion/scarb/blob/5dc5cceaa26d103fea70f39153ce60ba5f7358e5/scarb/src/ops/compile.rs#L147.

What do you think?

maciektr: Oh, I think I was confused. It does indeed work correctly. Sorry 🙇

Issue #2161, opened 2025-04-09
Scarb doc duplicated summary list items


FroyaTheHen:
Some items in summary file get duplicated. I am not sure what is the cause, my first call would be to check latest changes in https://github.com/software-mansion/scarb/pull/2125
Afaic, this shouldn't allow the mdbook to build. 

![Image](https://github.com/user-attachments/assets/acef26e6-e69c-47b8-a4be-94c869b47d5d)

PR #2216, opened 2025-04-29
ref: document pub uses

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2161

The idea is to document all reexported items and create pages for all their parents if they don't exist using original path not the reexported one. 
So the expected behaviour is that for example (comes from corelib) for `pub use adapters::PeekableTrait;` gets documented in path` core-iter-adapters-peekable-PeekableTrait` and only there. 
Before the change each _pub use_ of the same item would be documented in `SUMMARY.md` causing the `mdbook build` to fail. 

To reproduce this bug you can pull scarb repo (from before this PR gets merged) generate documentation for corelib or any other package that contains more than one pub use for the same item and run `mdbook build`. 

New interface sample 👇 
image

### DIFF
--- a/extensions/scarb-doc/src/docs_generation/markdown/traits.rs
+++ b/extensions/scarb-doc/src/docs_generation/markdown/traits.rs
@@ -6,8 +6,8 @@ use crate::docs_generation::{DocItem, PrimitiveDocItem, SubPathDocItem, TopLevel
 use crate::location_links::DocLocationLink;
 use crate::types::{
     Constant, Enum, ExternFunction, ExternType, FreeFunction, Impl, ImplAlias, ImplConstant,
-    ImplFunction, ImplType, ItemData, Member, Module, Struct, Trait, TraitConstant, TraitFunction,
-    TraitType, TypeAlias, Variant,
+    ImplFunction, ImplType, ItemData, Member, Module, ModulePubUses, Struct, Trait, TraitConstant,
+    TraitFunction, TraitType, TypeAlias, Variant,
 };
 use anyhow::Result;
 use cairo_lang_doc::parser::{CommentLinkToken, DocumentationCommentToken};
@@ -15,6 +15,8 @@ use itertools::Itertools;
 use std::collections::HashMap;
 use std::fmt::Write;
 use std::option::Option;
+
+const RE_EXPORTS_CHAPTER: &str = "Re-exports";
 
 pub trait TopLevelMarkdownDocItem: MarkdownDocItem + TopLevelDocItem {
     const ITEMS_SUMMARY_FILENAME: &'static str;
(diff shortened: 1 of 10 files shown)


extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -235,6 +237,84 @@ impl MarkdownDocItem for Impl {
...
+    )
+    .unwrap_or("".to_string());
+
+    buff += &generate_markdown_table_summary_for_reeksported_subitems(
+        &module_pubuses.use_free_functions.iter().collect_vec(),
+        context,
+    )
+    .unwrap_or("".to_string());
+
+    generate_markdown_table_summary_for_reeksported_subitems(
```
Draggu, line 258:
```suggestion
    buff += generate_markdown_table_summary_for_reeksported_subitems(
```

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -235,6 +237,84 @@ impl MarkdownDocItem for Impl {
...
+    )
+    .unwrap_or("".to_string());
+
+    buff += &generate_markdown_table_summary_for_reeksported_subitems(
+        &module_pubuses.use_enums.iter().collect_vec(),
+        context,
+    )
+    .unwrap_or("".to_string());
+
+    generate_markdown_table_summary_for_reeksported_subitems(
```
Draggu, line 270:
```suggestion
    buff += generate_markdown_table_summary_for_reeksported_subitems(
```

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -235,6 +237,84 @@ impl MarkdownDocItem for Impl {
...
+        &module_pubuses.use_module_type_aliases.iter().collect_vec(),
+        context,
+    )
+    .unwrap_or("".to_string());
+
+    generate_markdown_table_summary_for_reeksported_subitems(
```
Draggu, line 276:
```suggestion
    buff += generate_markdown_table_summary_for_reeksported_subitems(
```

Draggu (changes requested):

Draggu (approved):

extensions/scarb-doc/src/types.rs
```diff
@@ -50,6 +48,162 @@ impl Crate {
...
+            let parent_module_id = item.id.parent_module(db);
+            let ancestors = get_ancestors_vector(&mut Vec::new(), parent_module_id, db);
+            let pointer = self.ensure_module_structure(db, ancestors);
+            pointer.insert_extern_function(item);
+        }
+        for item in all_pub_ues.use_submodules.into_iter() {
+            let ancestors = get_ancestors_vector(&mut Vec::new(), item.module_id, db);
+            self.ensure_module_structure(db, ancestors);
+            // todo! replace pointer with a new merged module because it cannot be determined at this point if
+            // pointer was a virtual module. If so, not all module items might be documented as expected.
```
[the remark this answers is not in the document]
maciektr: Pls add an issue to reflect this 🙏 You can change this to `TODO(#{issue number})` to point to it in code.

maciektr (approved):

extensions/scarb-doc/src/types.rs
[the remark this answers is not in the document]
maciektr: I see it's fixed in #2217 - you don't need to do anything then, sorry 🙏

PR #2236, opened 2025-05-13
Expose dependency features in scarb-metadata

maciektr:

PR #2237, opened 2025-05-13
Reject source dep overrides when inheriting dep from workspace

maciektr:


maciektr:

- #2234
    - #2235
        - #2236
            - #2237 ◀
                - #2238
                    - #2239
                        - #2240

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -239,9 +239,16 @@ impl From<bool> for PathOrBool {
     }
 }
 
+/// When { workspace = true } you cannot define other keys that configure the source of
+/// the dependency such as `version`, `registry`, `path`, `git`, `branch`, `tag`, `rev`.
+/// You can also not define `default-features`.
+/// Those specifications are rejected through `deny_unknown_fields` attribute.
+/// Only `features` is allowed.
 #[derive(Debug, Default, Clone, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
```
DelevoXDG, line 248: This is one way to do it, but I wonder whether it's not overly aggressive
- Let me know if these are reasons for concern for us, but:
  - Manifests that previously included extra keys alongside `workspace`, and would be normally build with older Scarb versions, will now fail to parse
  - Any keys that could be added in the future Scarb versions would be incompatible with older Scarb versions
  - These are slightly intensified by not the most self-explanatory error message 😄 
- It doesn't seem like we hard-fail this way elsewhere, and Cargo doesn't as well (only [emitting a warning](https://github.com/rust-lang/cargo/blob/47c911e9e6f6461f90ce19142031fe16876a3b95/src/cargo/util/toml/mod.rs#L2658)). I wonder if warning would be sufficient here

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -239,9 +239,146 @@ impl From<bool> for PathOrBool {
...
+                }
+            })
+        }
+    }
+}
+
+/// When { workspace = true } you cannot define other keys that configure the source of
+/// the dependency such as `version`, `registry`, `path`, `git`, `branch`, `tag`, `rev`.
+/// You can also not define `default-features`.
+/// Those specifications are rejected through `deny_unknown_fields` attribute.
```
DelevoXDG, line 376: Let's remove / change this line since we changed the approach used to address this

scarb/tests/workspace.rs
```diff
@@ -126,3 +127,113 @@ fn target_name_duplicate() {
...
+        .failure()
+        .stdout_matches(indoc! {r#"
+        error: failed to parse manifest at: [..]Scarb.toml
+
+        Caused by:
+            TOML parse error at line 7, column 8
+              |
+            7 | some = { workspace = true, default-features = false }
+              |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+            field `default-features` is not allowed when inheriting workspace dependency
```
DelevoXDG, line 207: These are much better 🔥

DelevoXDG (approved):

PR #2238, opened 2025-05-13
Allow overriding features list when inheriting workspace dep

maciektr:


maciektr:

- #2234
    - #2235
        - #2236
            - #2237
                - #2238 ◀
                    - #2239
                        - #2240

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -480,13 +487,31 @@ impl TomlManifest {
...
+
+                // If `TomlWorkspaceDependency` declares `features` list,
+                // rewrite the inherited ws dependency with this list instead of the shared ws one.
+                let dep = match toml_dep {
+                    MaybeWorkspace::Workspace(w) => w
+                        .features
+                        .as_ref()
+                        .map(|features| ws_dep.with_features(features.clone())),
+                    _ => None,
+                };
```
DelevoXDG, line 507: Features specified in workspace-inherited dep are [additive](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace) in Cargo

Correct me if I'm wrong, but shouldn't these be additive here as well, for the same reason we disallow overriding setting `default-features=false` in workspace-inherited dep, as we cannot be disabling features in workspace-inherited dep?

maciektr: Done.

PR #2243, opened 2025-05-14
Refactor features selection

maciektr:

PR #2240, opened 2025-05-13
Allow declaring same dependency for two dep kinds (dependencies + dev-dependencies)

maciektr:

PR #2239, opened 2025-05-13
In resolver, dedpulicate package metadata request by parts of ManifestDependency that define package source / id only

maciektr:

PR #2247, opened 2025-05-15
Deserialize features enabling dependant features in other packages

maciektr:


maciektr:

- #2237
    - #2238
        - #2239
            - #2240
                - #2243
                    - #2244
                        - #2245
                            - #2247 ◀

scarb/src/core/manifest/mod.rs
```diff
@@ -143,27 +143,59 @@ pub fn edition_variant(edition: Edition) -> String {
     edition
 }
 
+#[derive(Clone, Debug)]
+pub struct FeatureEnabled {
```
DelevoXDG, line 147: Nit:
I would change this to `EnabledFeature`, or maybe something like `FeatureId`

scarb/src/core/manifest/mod.rs
```diff
@@ -176,14 +208,24 @@ impl FeaturesDefinition {
...
             enabled_features.iter().cloned().collect();
         if default_enabled {
-            selected_features.extend(self.default_features())
+            let default_features: Vec<FeatureName> = self
+                .default_features()
+                .into_iter()
+                .filter(|f| f.package.is_none())
+                .map(|f| f.feature)
+                .collect();
+            selected_features.extend(default_features)
```
DelevoXDG, line 217: Is there any particular reason we disallow `dep/feat` syntax for default features 🤔 

They seem to be allowed in Cargo: https://github.com/rust-lang/cargo/issues/10788

scarb/src/core/manifest/mod.rs
maciektr: Can be `EnabledFeature` if you prefer it, I don't want to use `FeatureId`, as it's not intended to differentiate between features, but to deserialize user input (i.e. you can have two instances of this struct that encode the same feature, but one has package: None and other specifies it).

scarb/src/core/manifest/mod.rs
maciektr: Supported in next PR.

FroyaTheHen (approved):
🚀

maciektr:
Merged into https://github.com/software-mansion/scarb/pull/2245

Issue #2229, opened 2025-05-09
Starkup: Add completions generation to profile file upon install / update


DelevoXDG:
Blocked by #2214

Issue #2228, opened 2025-05-09
Scarb install script: Add completions generation to profile file upon install / update


DelevoXDG:
Blocked by #2214

PR #2251, opened 2025-05-16
Allow generating CLI completions

DelevoXDG:
This PR supersedes #2214 

### Stack
- #2253
- #2251

### Proposed changes
- Adds `scarb completions <shell>` subcommand that allows to generate completions for Scarb and bundled subcommands. 
- Completions are generated dynamically to allow resolving external commands
- The external subcommands that are not bundled with Scarb are listed as well, but no completions are generated for those.
- Completions can be loaded on shell start, or on first TAB hit
   - Example:
     ```zsh
     source <(scarb completions zsh)
     autoload -Uz compinit && compinit
     ```
   - Lazy load on TAB hit example:
     ```zsh
     _scarb_last_version=""
     _scarb() {
     local scarb_current_version="$(scarb --version 2>/dev/null)"
     if [[ $scarb_current_version != "$_scarb_last_version" ]]; then
           eval "$(scarb completions zsh)"
            _scarb_last_version=$scarb_current_version
     fi
     _scarb "$@"
     }
     compdef _scarb scarb
     autoload -Uz compinit && compinit
     ```
  - The latter seems preferential, as it allows to always load completions for exact shim used by asdf
  - All installer scripts should be updated accordingly (#2229, #2228)


scarb/src/bin/scarb/args.rs
```diff
@@ -156,7 +157,7 @@ pub enum Command {
     Remove(RemoveArgs),
     /// Compile current project.
     Build(BuildArgs),
-    /// Expand macros.
+    /// Expand proc macros.
```
maciektr, line 160:
```suggestion
    /// Expand macros.
```
This will actually expand all macros, both proc-* and built-in.

scarb/src/bin/scarb/args.rs
```diff
@@ -222,6 +225,7 @@ pub enum EmitTarget {
 /// Arguments accepted by the `build` command.
 #[derive(Parser, Clone, Debug)]
 pub struct BuildArgs {
+    /// Specify package(s) to operate on.
```
maciektr, line 228: Does it have any visible output, when added on `command(flatten)`? 🤔

scarb/src/bin/scarb/args.rs
```diff
@@ -345,7 +354,9 @@ pub struct MetadataArgs {
 /// Arguments accepted by the `new` command.
 #[derive(Parser, Clone, Debug)]
 pub struct NewArgs {
+    /// Path to the new package directory.
     pub path: Utf8PathBuf,
+    /// Initialization options.
```
maciektr, line 359: vide supra

scarb/src/bin/scarb/args.rs
```diff
@@ -572,6 +591,44 @@ pub struct LintArgs {
...
+#[clap(version, about = "Generate shell completions")]
+pub struct CompletionsArgs {
+    /// Target shell for completion generation
+    #[arg(value_enum)]
+    pub shell: Shell,
+}
+
+/// Target shell for completion generation.
+#[derive(ValueEnum, Clone, Debug)]
+pub enum Shell {
```
maciektr, line 605: Why not use `ClapShell` directly? 
This would parse from `SHELL` var automagically for us.

scarb/src/ops/subcommands.rs
```diff
@@ -1,21 +1,51 @@
...
 use crate::process::exec_replace;
 use crate::subcommands::{EXTERNAL_CMD_PREFIX, SCARB_MANIFEST_PATH_ENV, get_env_vars};
+use itertools::Itertools;
+use scarb_ui::components::Status;
+
+#[derive(Debug)]
+pub struct SubcommandDirs {
+    pub scarb_exe_dir: PathBuf,
+    pub path_dirs: Vec<PathBuf>,
+}
```
maciektr, line 25: I don't understand what's the purpose of this struct. Why not keep using Config? 🤔

scarb/src/ops/subcommands.rs
```diff
@@ -87,28 +118,91 @@ pub fn execute_test_subcommand(
 ///
 /// # Search order
 ///
-/// This function searches for an executable in the following locations, in order:
+/// This function (unless path dirs overrides are specified) searches for an executable in the following locations, in order:
```
maciektr, line 121: WDYM?

scarb/src/ops/subcommands.rs
```diff
@@ -87,28 +118,91 @@ pub fn execute_test_subcommand(
...
+                .trim_start_matches(prefix)
+                .trim_end_matches(suffix)
+                .to_string();
+            if cmd_name.is_empty() {
+                return None;
+            }
+            Some(ExternalSubcommand {
+                name: cmd_name,
+                path: path.clone(),
+                is_bundled: path.parent() == Some(scarb_dir),
```
maciektr, line 201: This heuristic is fine for the completions script, but shouldn't be generalized. It does make sense for the user to install extensions to scarb install dir.

scarb/src/bin/scarb/args.rs
DelevoXDG: No, it's leftover from previous implementation that relied on `ScarbArgs` being public, which in turn caused `#![deny(missing_docs)]` to produce errors for each field that doesn't have a doc comment.

I planned on reverting these changes, but didn't because: 
- Args located in `scarb-extensions-cli` *are* public, and such comments are necessary there to avoid errors. 
- Prior to these changes, we already had several `command(flatten)` fields annotated in a similar fashion in this file

Therefore, I thought we might as well keep doc comments consistent both across the Scarb args and the extension args.

There's no real *need* to have them though, I can remove it if you prefer though.

scarb/src/bin/scarb/args.rs
DelevoXDG: Same reasoning as above, let me know what's your preference

scarb/src/bin/scarb/args.rs
DelevoXDG: > Why not use ClapShell directly?
This would parse from SHELL var automagically for us.

I've intended for the install scripts (starkup + scarb's own installer) to add blocks to profile files necessary to load completions.

These block will be different for each shell:
```sh
zsh_completion_block() {
  cat <<'EOF'
_scarb_last_version=""
_scarb() {
  local scarb_current_version="$(scarb --version 2>/dev/null)"
  if [[ $scarb_current_version != "$_scarb_last_version" ]]; then
    eval "$(scarb_build completions zsh)"
    _scarb_last_version=$scarb_current_version
  fi
  _scarb "$@"
}
compdef _scarb scarb
autoload -Uz compinit && compinit
EOF
}

bash_completion_block() {
  cat <<'EOF'
_scarb_last_version=""
_scarb_complete() {
  local scarb_current_version="$(scarb --version 2>/dev/null)"
  if [[ $scarb_current_version != "$_scarb_last_version" ]]; then
    source <(scarb_build completions bash)
    _scarb_last_version=$scarb_current_version
  fi
  _scarb
}
complete -o default -F _scarb_complete scarb
EOF
}
```
Thus, there's not much value in automatically writing `scarb completions` rather than of `scarb completions <zsh/bash/etc.>`
 
Although we definitely can do it, even using the current shell struct:

```rust
    let shell: ClapShell = args.shell.map_or(
        ClapShell::from_env()
            .expect("Failed to resolve shell from environment, please specify manually."),
        Into::into,
    );
```

DelevoXDG: Back to the main question 😅

> Why not use `ClapShell` directly?

We can derive `clap::ValueEnum`, resulting in more detailed completions based on our docstrings 
```
scarb completions
bash        -- Generate CLI completions for `bash`
elvish      -- Generate CLI completions for `elvish`
fish        -- Generate CLI completions for `fish`
powershell  -- Generate CLI completions for `powershell`
zsh         -- Generate CLI completions for `zsh`
```

Replacing it with `ClapShell` will result in slightly drier completions:
```
scarb completions
bash        elvish      fish        powershell  zsh
```

I suppose descriptions are nice-to-have, but are not really essential. I don't have a strong say on this, can replace it if you think we shouldn't add an intermediate struct 😄

scarb/src/ops/subcommands.rs
DelevoXDG: Points 2. and 3. will not be accurate if `path_dirs_override` is set:

https://github.com/software-mansion/scarb/blob/dcb24fa41812f1cb4019690120b104e5cf9128c3/scarb/src/core/dirs.rs#L30-L30

Adding a good comment here is tricky though, as:
- `path_dirs_override` is processed when setting up `Config` and `AppDirs`
- and the intended behaviour is specified in `find_external_subcommand` and `list_external_subcommands`, which do not directly accept `path_dirs_override`

scarb/src/bin/scarb/args.rs
maciektr: > I can remove it if you prefer

Cool, idc, just making sure.

scarb/src/bin/scarb/args.rs
maciektr: > Thus, there's not much value in automatically writing scarb completions rather than of scarb completions <zsh/bash/etc.>

I do feel like there is some value, one example could be non standard installation methods - we do have brew formulas, there has been a nix overlay in the past, etc. People may be installing Scarb in various different ways for various reasons. 

>  Replacing it with ClapShell will result in slightly drier completions:

Tbh I don't think these comments are that valuable.

scarb/src/ops/subcommands.rs
maciektr: > So, I added the struct to simplify this, avoid duplication, and establish single source of truth.

Yeah, but why not just some method on Config? 🤔

scarb/src/ops/subcommands.rs
maciektr: I think it's too nitpicky. 
If you override PATH via config, you expect Scarb to use the override as PATH. 
That's not contradictory with the doc.

scarb/src/bin/scarb/args.rs
DelevoXDG: Updated

scarb/src/ops/subcommands.rs
DelevoXDG: Ok, reverted

scarb/src/ops/subcommands.rs
DelevoXDG: Discussed offline, removed field from `ExternalSubcommand`, moved logic  to completions.rs

scarb/src/ops/subcommands.rs
DelevoXDG: Tbh current approach seemed cleaner, as it brings related logic to one place 

> Yeah, but why not just some method on Config?

We'd need two methods, e.g. `subcommands_dirs()` and `scarb_exe_dir()`, which seems to me:
- Separates related logic across two methods
- Introduces two methods to `Config` that will not be used elsewhere

If those don't seem significant to you, I can change it to methods instead

scarb/src/bin/scarb/commands/completions.rs
```diff
@@ -0,0 +1,96 @@
...
+use std::io;
+
+#[tracing::instrument(skip_all, level = "info")]
+pub fn run(args: CompletionsArgs, config: &Config) -> Result<()> {
+    let mut cmd = build_command(config)?;
+    let shell = args.shell.or_else(Shell::from_env).ok_or_else(|| {
+        anyhow!(indoc! {r#"
+            could not automatically determine shell to generate completions for.
+            help: specify the shell explicitly: `scarb completions <shell>`.
+            For the list of supported shells, run `scarb completions --help`.
```
maciektr, line 26:
```suggestion
            could not automatically determine shell to generate completions for
            help: specify the shell explicitly: `scarb completions <shell>`
            for the list of supported shells, run `scarb completions --help`
```

scarb/src/ops/subcommands.rs
maciektr: > current approach seemed cleaner

Ok, let's keep it then

maciektr (approved):
Tried locally and it looks fabulous 😍

scarb/tests/completions.rs
```diff
@@ -0,0 +1,301 @@
...
+    Scarb::quick_snapbox()
+        .arg("completions")
+        .env("SHELL", "")
+        .assert()
+        .failure()
+        .stdout_eq(indoc!(
+            r#"
+            error: could not automatically determine shell to generate completions for.
+            help: specify the shell explicitly: `scarb completions <shell>`.
+            For the list of supported shells, run `scarb completions --help`.
```
maciektr, line 298:
```suggestion
            error: could not automatically determine shell to generate completions for
            help: specify the shell explicitly: `scarb completions <shell>`
            for the list of supported shells, run `scarb completions --help`
```

PR #2258, opened 2025-05-20
Set call site location in generated code mappings

maciektr:


scarb/src/compiler/plugin/proc_macro/v2/host/mod.rs
```diff
@@ -237,5 +242,18 @@ pub fn generate_code_mappings(token_stream: &TokenStream) -> Vec<CodeMapping> {
...
+        end: TextOffset::default().add_width(TextWidth::new_for_testing(call_site.end)),
+    };
+    mappings.push(CodeMapping {
+        span: TextSpan {
+            start: TextOffset::default(),
+            end: mappings.last().map(|m| m.span.end).unwrap_or_default(),
+        },
+        origin: CodeOrigin::CallSite(call_site),
+    });
+    dbg!(token_stream, &mappings);
```
piotmag769, line 257: dbg

Issue #2254, opened 2025-05-19
Add additional instructions after running scarb doc --build


FroyaTheHen:
For example:

> Saving output to: target/doc/token_sender
> Saving build output to: target/doc/token_sender/book
> 
> Run the following to see the results:
> $ mdbook serve target/doc/token_sender 
> 
> Or open the following in your browser
> $HOME/Code/Cairo/starknet-foundry-template/target/doc/token_sender/book/index.html

PR #2259, opened 2025-05-20
feat: add info to doc --build

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2254

### DIFF
--- a/extensions/scarb-doc/src/main.rs
+++ b/extensions/scarb-doc/src/main.rs
@@ -72,16 +72,25 @@ fn main_inner(args: Args, ui: Ui) -> Result<()> {
                     ScarbCommand::new()
                         .arg("mdbook")
                         .arg("--input")
-                        .arg(pkg_output_dir)
+                        .arg(pkg_output_dir.clone())
                         .arg("--output")
                         .arg(build_output_dir.clone())
                         .env("SCARB_UI_VERBOSITY", ui.verbosity().to_string())
                         .run()?;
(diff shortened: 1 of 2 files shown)


extensions/scarb-doc/src/main.rs
```diff
@@ -72,16 +72,20 @@ fn main_inner(args: Args, ui: Ui) -> Result<()> {
...
-                    let output_path = build_output_dir
+                    let build_output_path = build_output_dir
                         .strip_prefix(&metadata.workspace.root)
                         .unwrap_or(&build_output_dir)
                         .to_string();
-                    ui.print(Status::new("Saving build output to:", &output_path));
+                    ui.print(Status::new("Saving build output to:", &build_output_path));
+                    ui.print(format!(
+                        "\nRun the following to see the results: \n`mdbook serve {output_path}`\
+                         \n\nOr open the following in your browser: \n`{pkg_output_dir}/book/index.html`",
```
maciektr, line 87: But it's only there, if you add `--build` flag, right? 🤔

FroyaTheHen: yes, do you think we should print more info in all cases??

maciektr: I think we can show something like

> Run the following to see the results: `mdbook serve {output_path}` (you will need to have mdbook installed)
> Or build html docs by running `scarb doc --build` 

In the other case.

PR #2260, opened 2025-05-20
Use span without trivia as call site location

maciektr:


scarb/src/compiler/plugin/proc_macro/v2/host/conversion.rs
```diff
@@ -15,7 +15,7 @@ pub trait SpanSource {
 impl<T: TypedSyntaxNode> SpanSource for T {
     fn text_span(&self, db: &dyn SyntaxGroup) -> TextSpan {
         let node = self.as_syntax_node();
-        let span = node.span(db);
+        let span = node.span_without_trivia(db);
```
piotmag769, line 18: Lol, we have it this way in LS already for some reason xD That's why it worked for us

piotmag769 (approved):

PR #2265, opened 2025-05-21
Add path argument to linter

MKowalski8:


scarb/src/bin/scarb/args.rs
```diff
@@ -570,6 +570,10 @@ pub struct LintArgs {
     /// Should fail on any warning.
     #[arg(short, long, default_value_t = false, env = "SCARB_LINT_DENY_WARNINGS")]
     pub deny_warnings: bool,
+
+    /// Path to a file or directory to lint. If provided, only this file or directory will be linted.
+    #[clap(value_name = "PATH")]
```
maciektr, line 575:
```suggestion
    #[clap(value_name = "SCARB_FILE_PATH")]
```
wdyt? 
Overriding PATH may have unexpected effects. 
You can also change it in FmtArgs 😅

scarb/src/ops/lint.rs
```diff
@@ -53,6 +56,8 @@ pub fn lint(opts: LintOptions, ws: &Workspace<'_>) -> Result<()> {
         },
     )?;
 
+    let absolute_path = opts.path.as_ref().and_then(|path| canonicalize(path).ok());
```
maciektr, line 59: I don't think we should silently ignore the error, when the provided path is invalid

scarb/tests/lint.rs
```diff
@@ -580,3 +580,136 @@ fn lint_selected_targets() {
...
+            fn main() {
+                let x = true;
+                if x == false {
+                    println!("x is false");
+                }
+            }
+        "#})
+        .build(&t);
+
+    t.child("src/other.cairo")
```
maciektr, line 601: FYI you can do `.src(path, content)` on ProjectBuilder

scarb/tests/lint.rs
```diff
@@ -580,3 +580,136 @@ fn lint_selected_targets() {
...
+        warn: Plugin diagnostic: unnecessary double parentheses found after break. Consider removing them.
+         --> [..]/src/other.cairo:3:9
+                break ();
+                ^^^^^^^^^
+
+        "#});
+}
+
+#[test]
+fn lint_specific_module() {
```
maciektr, line 628: Nitpick:
```suggestion
fn lint_specific_directory() {
```
(as it's not semantic)

scarb/src/ops/lint.rs
```diff
@@ -166,7 +171,26 @@ pub fn lint(opts: LintOptions, ws: &Workspace<'_>) -> Result<()> {
                         .iter()
                         .flat_map(|diags| {
                             let all_diags = diags.get_all();
-                            all_diags.iter().for_each(|diag| match diag.severity() {
+
+                            let filtered = match &absolute_path {
+                                Some(path) => all_diags
+                                    .into_iter()
+                                    .filter_map(|diag| {
+                                        let file_id = diag.stable_location.file_id(&db);
```
maciektr, line 179: Note, that this can be a `FileId` of a virtual file (e.g. macro generated). 
Should you show lints from files generated from the provided file path? 
Please add a comment in a code documenting that you are making this decision intentionally.

maciektr (changes requested):

scarb/src/ops/lint.rs
MKowalski8: As far as I understand, cairo-lint  filters out virtual files, so diagnostics won't be produced for them. I can add a comment like `// This might be a FileId of a virtual file`, but with the current linter logic, this case shouldn't happen at all.

wawel37: Yup, that's true. It's being done in the linter's plugin itself https://github.com/software-mansion/cairo-lint/blob/b359855bd930e89ade76bd9c1852fb74b75b5861/src/plugin.rs#L105. And as the `ScarbDatabase` also uses our plugin, it's filtered out by default

scarb/src/ops/lint.rs
```diff
@@ -166,7 +171,26 @@ pub fn lint(opts: LintOptions, ws: &Workspace<'_>) -> Result<()> {
                         .iter()
                         .flat_map(|diags| {
                             let all_diags = diags.get_all();
-                            all_diags.iter().for_each(|diag| match diag.severity() {
+
+                            let filtered = match &absolute_path {
```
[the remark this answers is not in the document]
MKowalski8: The only place I can move the filtering to is before `flat_map`, because diagnostics are displayed inside this very `flat_map`. On the other hand, I could perform the `flat_map`, then `filter`, and move the display logic elsewhere, if you think that's a better idea.

scarb/src/bin/scarb/args.rs
maciektr: `TARGET` is iffy, as it can easily be confused with Scarb target dir (https://docs.swmansion.com/scarb/docs/writing-extensions/subcommands.html#environment-variables) or even Scarb targets  (https://docs.swmansion.com/scarb/docs/reference/targets.html#targets) 😅 

Maybe `SCARB_ACTION_PATH` or smth? 🤔

scarb/src/ops/lint.rs
maciektr: That's fine, If you can add a short note, I fell like this would be for me :)

scarb/src/ops/lint.rs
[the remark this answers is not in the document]
wawel37: I think we should filter before doing the flat_map and displaying

scarb/src/ops/lint.rs
```diff
@@ -53,6 +57,14 @@ pub fn lint(opts: LintOptions, ws: &Workspace<'_>) -> Result<()> {
         },
     )?;
 
+    let absolute_path = opts
+        .path
+        .as_ref()
+        .map(|path| {
+            canonicalize(path).map_err(|_| anyhow!("Provided `SCARB_ACTION_PATH` is invalid"))
+        })
+        .transpose()?;
```
maciektr, line 66:
```suggestion
        .map(canonicalize)
        .transpose()?;
```
Canonicalize will return an error that can be helpful for the user (https://github.com/software-mansion/scarb/blob/4388a3801ef5a7f4a72651b08873abeaa76a2fa4/scarb/src/internal/fsx.rs#L18)

If you need to add additional context, use `.with_context` instead of replacing the error fully. 

Also, I would not phrase it as `provided ENV_NAME`, as it can be passed via cli (mostly will).

maciektr (changes requested):

maciektr (approved):

PR #2272, opened 2025-05-23
feat: attributes mod

FroyaTheHen:


extensions/scarb-doc/src/attributes.rs
```diff
@@ -0,0 +1,52 @@
...
+    db: &ScarbDocDatabase,
+    id: &impl TopLevelLanguageElementId,
+) -> Option<String> {
+    let node = id.stable_location(db).syntax_node(db);
+
+    if let Some(attr) = node.find_attr(db, "doc") {
+        // Process the arguments of the "doc" attribute, if any.
+        if let OptionArgListParenthesized::ArgListParenthesized(args) = attr.arguments(db) {
+            for arg in args.arguments(db).elements(db).iter() {
+                if let ArgClause::Unnamed(clause) = arg.arg_clause(db) {
```
maciektr: I think we should use `doc(group: 'name')` syntax, i.e. the `Named` variant. wdyt? 🤔

extensions/scarb-doc/src/attributes.rs
maciektr, line 1: It would be nice to emit warnings / errors when doc attr with unsupported syntax is found (i.e. doc(group = 'name') vs doc(group: 'name') etc.).

extensions/scarb-doc/src/attributes.rs
```diff
@@ -0,0 +1,52 @@
...
+                if let Some(text) = child.text(db) {
+                    return Some(clean_string(&text));
+                }
+            }
+        }
+    }
+    None
+}
+
+fn clean_string(input: &SmolStr) -> String {
```
maciektr: Do we really need this? 🤔

extensions/scarb-doc/src/attributes.rs
```diff
@@ -0,0 +1,52 @@
...
+/// Extracts string group information from the "doc" attributes of a node.
+pub fn find_groups_from_attributes(
+    db: &ScarbDocDatabase,
+    id: &impl TopLevelLanguageElementId,
+) -> Option<String> {
+    let node = id.stable_location(db).syntax_node(db);
+
+    if let Some(attr) = node.find_attr(db, "doc") {
+        // Process the arguments of the "doc" attribute, if any.
+        if let OptionArgListParenthesized::ArgListParenthesized(args) = attr.arguments(db) {
```
maciektr: Not sure if it changes anything, but might be easier with `attr.structurize(db)` 🤔

extensions/scarb-doc/src/attributes.rs
FroyaTheHen: yeah, you're right.

extensions/scarb-doc/src/attributes.rs
FroyaTheHen: Right, we don't. will remove :)

extensions/scarb-doc/src/attributes.rs
FroyaTheHen: done in [separate PR](https://github.com/software-mansion/scarb/pull/2288)

maciektr (approved):

PR #2282, opened 2025-05-30
Implement `scarb tree`

### LINKED ISSUE #2188: Add `scarb tree` command
This should print a dependency tree of a Scarb project. 
See https://doc.rust-lang.org/cargo/commands/cargo-tree.html

mkaput:
fix #2188

### DIFF
--- a/scarb/src/bin/scarb/commands/tree.rs
+++ b/scarb/src/bin/scarb/commands/tree.rs
@@ -82,7 +82,7 @@ fn build(
             tree.package = Some(package_id);
 
             // Recursively visit dependencies if we haven't visited this package before
-            // or if no_dedupe is true, or if there is a cycle.
+            // or if no_dedupe is true, unless there is a cycle.
             let already_visited = self.visited.contains(&package_id) && !self.args.no_dedupe;
             let is_cycle = self.stack.iter().filter(|p| **p == package_id).count() > 1;
             tree.already_visited_duplicate = already_visited || is_cycle;


scarb/src/bin/scarb/args.rs
```diff
@@ -604,6 +609,30 @@ pub struct CompletionsArgs {
     pub shell: Option<Shell>,
 }
 
+/// Arguments accepted by the `tree` command.
+#[derive(Parser, Clone, Debug)]
+pub struct TreeCommandArgs {
```
mkaput, line 614: I've ignored the `--invert` flag (that `cargo tree` has) because it would employ a entirely different tree building algorithm -- and I don't need this feature for immediate use cases.

Same reason for `--edges`, though that one should be simpler to implement.

scarb/tests/tree.rs
```diff
@@ -0,0 +1,441 @@
...
+    ProjectBuilder::start()
+        .name("root")
+        .version("0.1.0")
+        .dep("dep1", &dep1)
+        .dep("dep2", &dep2)
+        .build(&root);
+
+    Scarb::quick_snapbox()
+        .arg("tree")
+        .arg("--no-dedupe")
```
maciektr, line 136: removing the arg does not change the output 🤔 Please add some deps to common, so they can be deduplicated.

maciektr:
Cycle + `--no-dedupe` will result in a infinite loop

```
#[test]
fn cycle() {

    let root = t.child("root");
    let dep1 = t.child("dep1");
    ProjectBuilder::start()
        .name("dep1")
        .version("0.1.0")
        .dep("root", &root)
        .build(&dep1);

    ProjectBuilder::start()
        .name("root")
        .version("0.1.0")
        .dep("dep1", &dep1)
        .build(&root);

    Scarb::quick_snapbox()
        .arg("tree")
        .arg("--no-dedupe")
        .current_dir(&root)
        .assert()
        .success();
}
```

scarb/tests/tree.rs
mkaput: 2d888d7353f2c0adbb9538bfdb68a7a12aa6799a

scarb/src/bin/scarb/commands/tree.rs
```diff
@@ -0,0 +1,313 @@
...
+                if depth > max_depth {
+                    tree.max_depth_reached = true;
+                    return Err(BuildError::MaxDepthReached);
+                }
+            }
+
+            tree.package = Some(package_id);
+
+            // Recursively visit dependencies if we haven't visited this package before
+            // or if no_dedupe is true, or if there is a cycle.
```
maciektr:
```suggestion
            // or if no_dedupe is true, unless there is a cycle.
```

Nit: I guess? 🤔

mkaput: grammar is hard here, ideally I'd need parentheses, because what I want here is:

```
(haven't visited this package before || if no_dedupe is true) || there is a cycle
```

cycles are handled no matter what value `no_dedupe` has

code has this logic demorganed

PR #2288, opened 2025-06-03
add diagnostic warnings for doc groups

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2133

**Stack**:
- #2275
- #2274
- #2273
- #2272

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #2274, opened 2025-05-23
feat: aggregate_groups - doc

FroyaTheHen:
**Stack**:
- #2273
- #2272

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #2291, opened 2025-06-04
doc groups for module reeksports

FroyaTheHen:
**Stack**:
- #2288
- #2275
- #2274
- #2273
- #2272

### DIFF
--- a/extensions/scarb-doc/src/types/groups.rs
+++ b/extensions/scarb-doc/src/types/groups.rs
@@ -1,4 +1,4 @@
-use crate::types::module_type::Module;
+use crate::types::module_type::{Module, ModulePubUses};
 use crate::types::other_types::{
     Constant, Enum, ExternFunction, ExternType, FreeFunction, Impl, ImplAlias, Struct, Trait,
     TypeAlias,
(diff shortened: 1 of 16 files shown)


extensions/scarb-doc/src/types/module_type.rs
```diff
@@ -338,6 +342,9 @@ impl Module {
         aggregate_extern_types_groups(&mut extern_types, &mut group_map);
         aggregate_extern_functions_by_group(&mut extern_functions, &mut group_map);
         aggregate_modules_by_group(&mut submodules, &mut group_map);
+        if !include_private_items {
+            aggregate_pub_uses_groups(&module_pubuses, &mut group_map);
+        }
```
maciektr, line 347: What would happen if the pub use reexports something from a different crate? 🤔

extensions/scarb-doc/src/types/groups.rs
```diff
@@ -65,22 +65,100 @@ macro_rules! aggregate_groups {
...
+aggregate_groups!(aggregate_impls_groups, Impl, impls);
+aggregate_groups!(aggregate_extern_types_groups, ExternType, extern_types);
 aggregate_groups!(
     aggregate_extern_functions_by_group,
-    Vec<ExternFunction>,
+    ExternFunction,
     extern_functions
 );
-aggregate_groups!(aggregate_modules_by_group, Vec<Module>, submodules);
+aggregate_groups!(aggregate_modules_by_group, Module, submodules);
```
maciektr: Can't we return result from aggregate_groups instead of replacing and use it instead of this -> just ignoring the return value? 🤔

extensions/scarb-doc/tests/dog_groups_in_reeksports.rs
```diff
@@ -0,0 +1,78 @@
...
+        .lib_cairo(DOC_GROUPS_REEKSPORTS_CODE);
+
+    WorkspaceBuilder::start().package(root).build(&root_dir);
+
+    Scarb::quick_snapbox()
+        .arg("doc")
+        .args(["--document-private-items", "--build"])
+        .current_dir(&root_dir)
+        .assert()
+        .success();
```
maciektr, line 77: why don't we assert the output here? 🤔

FroyaTheHen: I wanted to avoid storing all all generated html files because there are many and I don't find them relevant as they're a result of `mdbook build` execution which is an external tool.    
The output for markdown is checked in the `doc_groups_reeksports_markdown` test above and in this (`doc_groups_reeksports_markdown_doesnt_duplicate_groups`) test I wanted just to cover the build success and the case of potential duplicates in summary which would cause build to fail.

extensions/scarb-doc/src/types/module_type.rs
FroyaTheHen: Hmm, good question. 
Including this PR, the reeksported item wouldn't have its subpage created.

However, before the doc gropus stack gets merged the item would be documented in the wrong parent. Then the path for the foregin-crate-item wouldn't work either. 

That is imo a separate problem. The code does not handle the foregin-crates-reeksports correctly. 

created an issue https://github.com/software-mansion/scarb/issues/2302

extensions/scarb-doc/tests/dog_groups_in_reeksports.rs
maciektr: Makes sense! 🙇

PR #2321, opened 2025-06-16
Define fingerprint structure

maciektr:
Co-authored-by: Maksim Zdobnikau <maksim.zdobnikau@swmansion.com>

### DIFF
--- a/scarb/src/compiler/compilation_unit.rs
+++ b/scarb/src/compiler/compilation_unit.rs
@@ -11,6 +11,7 @@ use std::hash::{Hash, Hasher};
 use std::sync::Arc;
 use typed_builder::TypedBuilder;
 
+use crate::FINGERPRINT_DIR_NAME;
 use crate::compiler::Profile;
 use crate::compiler::plugin::proc_macro::ProcMacroInstance;
 use crate::core::{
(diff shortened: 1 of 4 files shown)


DelevoXDG (approved):

scarb/src/compiler/compilation_unit.rs
```diff
@@ -321,6 +321,10 @@ impl CairoCompilationUnit {
         ws.target_dir().child(self.profile.as_str())
     }
 
+    pub fn fingerprint_dir(&self, ws: &Workspace<'_>) -> Filesystem {
+        self.target_dir(ws).child(".fingerprint")
```
DelevoXDG: Nit: 
I think it's typical we store path names like these in `scarb/src/lib.rs`

maciektr: Personally, since we are not likely to use this in other places, hiding it under a method should be enough. It's also what we do in some other places, like the cache structure, e.g. 
https://github.com/software-mansion/scarb/blob/8ecd0bfa62a15f8115c16a3036004c04ddaeba45/scarb/src/core/dirs.rs#L71-L71
I worry, that storing every name in src/lib.rs will soon cause it to be unreadable. 
WDYT? 🤔

DelevoXDG: It wouldn't be the first time we add variable to this file that is only used once 😅 

Personally, I wouldn't worry about 2 extra lines, and it seems *slightly* cleaner to have a reference point for path names, should we ever use them elsewhere. 

With that said, this is a nitpick with a big n, so the final decision is up to you 😄

maciektr: Changed

PR #2322, opened 2025-06-16
refactor: MarkdownContent to store hash map of files, fix: unify nesting level

FroyaTheHen:


extensions/scarb-doc/src/docs_generation/markdown.rs
```diff
@@ -65,3 +83,48 @@ impl MarkdownContent {
         Ok(())
     }
 }
+
+pub struct IndexMap<K, V> {
```
maciektr, line 112: Pls add small doc saying why we need this over standard map / ordered map 🙏

extensions/scarb-doc/src/docs_generation/markdown.rs
```diff
@@ -65,3 +83,48 @@ impl MarkdownContent {
...
+            self.map.insert(key, (value, position));
+        }
+    }
+
+    /// Insert multiple key-value pairs, replaces any existing value for the key.
+    fn insert_multiple(&mut self, items: Vec<(K, V)>) {
+        for (key, value) in items {
+            self.insert(key, value);
+        }
+    }
```
maciektr: NIT: Can be done as extend trait
```
impl<K, V> Extend<(K, V)> for IndexMap<K, V>
where
    K: std::hash::Hash + Eq + Clone,
{
    fn extend<T: IntoIterator<Item = (K, V)>>(&mut self, iter: T) {
        for (k, v) in iter {
            self.insert(k, v);
        }
    }
}
```

extensions/scarb-doc/src/docs_generation/markdown.rs
```diff
@@ -28,10 +28,12 @@ const SHORT_DOCUMENTATION_AVOID_PREFIXES: &[&str] = &["#", "\n\n", "```\n", "- "
 
 type Filename = String;
 type GeneratedFile = (Filename, String);
+/// Stores `SUMMARY.md` files data: name and list indent.
+pub type SummaryIndexMap = IndexMap<String, (String, usize)>;
```
maciektr: NIT: Consider changing the pair to a named struct, to better convey the meaning of the usize (or at least `type NestingLevel = usize;`).

PR #2327, opened 2025-06-16
Add docs for Scarb shell completions

DelevoXDG:
Added docs so we can reference those on failure when installing via starkup / Scarb's installation script


website/download.md
```diff
@@ -109,7 +109,136 @@ version number.
 curl --proto '=https' --tlsv1.2 -sSf https://docs.swmansion.com/scarb/install.sh | sh -s -- -v {{ rel.sampleVersion }}
 ```
 
-### Uninstall
+## Shell completions (Optional)
```
maciektr: NIT(as in starkup PR) - maybe just "Shell completions"? 🤔

website/download.md
```diff
@@ -109,7 +109,136 @@ version number.
...
+# BEGIN SCARB COMPLETIONS
+_scarb() {
+  if ! scarb completions zsh >/dev/null 2>&1; then
+    return 0
+  fi
+  eval "$(scarb completions zsh)"
+  _scarb "$@"
+}
+compdef _scarb scarb
+autoload -Uz compinit && compinit
```
maciektr: (As in Starkup PR) Shouldn't compinit be run before compdef?

website/download.md
DelevoXDG: I mainly took after [Foundry](https://foundry-rs.github.io/starknet-foundry/getting-started/installation.html#set-up-shell-completions-optional) and [asdf](https://asdf-vm.com/guide/getting-started.html#set-up-shell-completions-optional) that have "optional" in the heading.

I don't have a strong say on this tbh, I don't like having `-optional` in the link, but it also makes sense to emphasize that this step is not necessary for installation.

WDYT?

maciektr: Personally, I would drop it. 
Most of our docs describe things you **can do**, not **have to do**. 
But yes, it's fine both ways 👍

DelevoXDG: Dropped it then 👍🏻

PR #2335, opened 2025-06-23
WIP: incremental

maciektr:
Take unit by reference in compiler.compile()

Emit fingerprints and cache blobs for corelib crate


maciektr:

- #2320
    - #2326
        - #2321
            - #2335 ◀

FroyaTheHen (approved):

scarb/src/compiler/compilation_unit.rs
```diff
@@ -325,6 +325,10 @@ impl CairoCompilationUnit {
         self.target_dir(ws).child(".fingerprint")
     }
 
+    pub fn incremental_cache_dir(&self, ws: &Workspace<'_>) -> Filesystem {
+        self.target_dir(ws).child("incremental")
```
DelevoXDG: Nit: 
I think it's typical we store path names like these in `scarb/src/lib.rs`

scarb/src/compiler/incremental.rs
```diff
@@ -1,10 +1,150 @@
...
+            fingerprint_dir.create_rw(&component_id, "fingerprint file", ws.config())?;
+        fingerprint_file
+            .deref()
+            .write_all(digest.as_bytes())
+            .with_context(|| {
+                format!(
+                    "failed to write fingerprint to `{}`",
+                    fingerprint_file.path()
+                )
+            })?;
```
DelevoXDG: Cargo uses separate hash for `<target>-<digest>` and for `<fingerprint_hash>` itself. The fingerprint hash is exhaustive, while the digest hash is limited identifier of the package. Which means that if, for example, compiler version changes, instead of creating a new fingerprint dir and a new cache blob, the **existing fingerprint** for given `<target>-<digest>` will be **overwritten**, same as the corresponding cache file.

Unlike Cargo and #2304 that attempted to replicated this behaviour, we instead use the same exhaustive hash in both cases.

I wonder if this is deliberate choice? It means less re-compilations, but also more cache files.

However, if that is deliberate decision, I wonder whether we need fingerprint file at all?

Screenshot 2025-06-25 at 13 08 53
Screenshot 2025-06-25 at 13 08 29

Correct me if I'm wrong, but it seems we nest and compare the same hash three times. We could instead not store fingerprint file at all, since cache file name basically contains fingerprint hash, hence cache file is never overwritten. It seems it would be sufficient to computer fingerprint hash, and check whether `<target>-<fingerprint_hash>.bin` exists, and load it.

maciektr: > I wonder if this is deliberate choice? It means less re-compilations, but also more cache files.

Yes, since we only cache corelib here, there is no point in differentiating between the two. 

> However, if that is deliberate decision, I wonder whether we need fingerprint file at all?

Futureproofing. We will introduce the split later on.

maciektr: > creating a new fingerprint dir and a new cache blob, the existing fingerprint for given <target>-<digest> will be overwritten,


AFAIK this is not true - rustc version should be used by Cargo for both fingerprint and metadata? 🤔 
https://github.com/rust-lang/cargo/blob/409fed7dc1553d49cb9a8c0637d12d65571346ce/src/cargo/core/compiler/fingerprint/mod.rs#L69

DelevoXDG: > AFAIK this is not true - rustc version should be used by Cargo for both fingerprint and metadata? 🤔

Correct me if I'm wrong, but it seems under some circumstances fingerprint dir can be generated based on `package_id` + `metadata_version` only: https://github.com/rust-lang/cargo/blob/68c26f6c3f9aedf005f28d61211b541e2f909328/src/cargo/core/compiler/build_runner/compilation_files.rs#L243-L243

> since we only cache corelib here, there is no point in differentiating between the two.

Can you please elaborate? It seems to me some of the circumstances in which fingerprint changes, but digest doesn't - e.g. changes to source paths - would overlap between corelib and any other package.

> Futureproofing. We will introduce the split later on.

If that's the case, it seems to me we should store fingerprint under name:
`.fingerprint/<target>-<hash>/<target>` instead of 
`.fingerprint/<target>-<hash>/<target>-<hash>`.

So, when the split occurs, we actually overwrite the old fingerprint file for given `<target>-<hash>`, instead of keeping invalidated fingerprint with no fresh corresponding cache file.

maciektr: 1) and 3) -> discussed offline 

2) I'll reverse the question - what is now used as part of the path, that you feel should not be used?

maciektr: discussed offline

scarb/src/compiler/fingerprint.rs
```diff
@@ -111,3 +110,24 @@ impl Fingerprint {
...
+pub fn is_fresh(
+    fingerprint: &Fingerprint,
+    fingerprint_dir: &Filesystem,
+    target_name: &str,
+) -> Result<bool> {
+    let digest = fingerprint.digest();
+    let component = format!("{target_name}-{digest}");
+    let fingerprint_dir = fingerprint_dir.child(&component);
+    let fingerprint_dir = fingerprint_dir.path_unchecked();
+    let old_digest_path = fingerprint_dir.join(&component);
```
DelevoXDG: As previously discussed offline https://swmansion.slack.com/archives/D05FFE13D51/p1750857269160039, I think we should avoid duplicate hash in fingerprint path, and change `{target}-{hash}` to just `{target}`

DelevoXDG (commented):
LGTM beside one note

scarb/src/compiler/fingerprint.rs
maciektr: Changed.

PR #2337, opened 2025-06-24
Pass adapted call site to generate code mappings fn

maciektr:


scarb/src/compiler/plugin/proc_macro/v2/host/attribute/span_adapter.rs
```diff
@@ -221,6 +221,17 @@ impl ExpandableAttrLocation {
                         };
                         CodeOrigin::Span(span)
                     }
+                    CodeOrigin::CallSite(span) => {
+                        let call_site = CairoTextSpan {
+                            start: span.start.add_width(TextWidth::new_for_testing(
+                                attr_start - self.item_start_offset,
```
Arcticae: this is reused many times, perhaps it could be a variable?

PR #2340, opened 2025-06-25
Include component dependencies in fingerprint digest

maciektr:


scarb/src/compiler/fingerprint.rs
```diff
@@ -50,9 +56,127 @@ pub struct Fingerprint {
...
+        for component in unit.components.iter() {
+            for dep in component
+                .dependencies
+                .iter()
+                .sorted_by_key(|dep| dep.component_id())
+            {
+                let dep_component_id = match dep {
+                    CompilationUnitDependency::Library(component_id) => component_id,
+                    CompilationUnitDependency::Plugin(component_id) => component_id,
+                };
```
DelevoXDG: Why not use `.component_id()`?

DelevoXDG: Alternatively, can rewrite loop 🤔 
```rs
            for dep_component_id in component
                .dependencies
                .iter()
                .map(|dep| dep.component_id())
                .sorted()
```

scarb/src/compiler/fingerprint.rs
```diff
@@ -50,9 +56,127 @@ pub struct Fingerprint {
...
+                let fingerprint = fingerprints.get(dep_component_id);
+                let fingerprint =
+                    fingerprint.expect("component fingerprint must exist in unit fingerprints");
+                let fingerprint = Rc::downgrade(fingerprint);
```
DelevoXDG: Nit: 
a bit cleaner IMO
```rs
                let fingerprint = fingerprints
                    .get(dep_component_id)
                    .map(Rc::downgrade)
                    .expect("component fingerprint must exist in unit fingerprints");
```

scarb/src/compiler/fingerprint.rs
```diff
@@ -125,17 +250,50 @@ impl Fingerprint {
     /// This uniquely identifies the compilation environment for a component,
     /// allowing to determine if the cache can be reused or if a recompilation is needed.
     pub fn digest(&self) -> String {
+        // We use the set to avoid cycles when calculating digests recursively for deps.
+        let mut seen = HashSet::<SmolStr>::new();
+        seen.insert(self.component_discriminator.clone());
+        Self::do_digest(self, &mut seen)
+    }
+
+    fn do_digest(fingerprint: &Fingerprint, seen: &mut HashSet<SmolStr>) -> String {
```
DelevoXDG: nit:
maybe `calculate_digest` instead?

maciektr: Sounds good, changed!

scarb/src/compiler/fingerprint.rs
maciektr: Sounds good, changed!

scarb/src/compiler/fingerprint.rs
maciektr: Sounds good, changed!

PR #2342, opened 2025-06-26
Profiling config

maciektr:
- **Add tracing-chrome profile config**
- **Add more instruments to compilation**
- **Add instruments to proc macro generate code**
- **fixup! Add more instruments to compilation**

### DIFF
--- a/scarb/src/bin/scarb/main.rs
+++ b/scarb/src/bin/scarb/main.rs
@@ -36,8 +36,8 @@ fn main() -> ExitCode {
 }
 
 fn init_logging(verbose: VerbositySpec, ui: &Ui) -> Option<impl Drop> {
+    use chrono::Local;
     use std::fs;
-    use std::time::SystemTime;
 
     use std::path::PathBuf;
     use tracing_chrome::ChromeLayerBuilder;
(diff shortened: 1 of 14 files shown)


scarb/src/ops/compile.rs
```diff
@@ -246,6 +247,12 @@ fn compile_unit_inner(unit: CompilationUnit, ws: &Workspace<'_>) -> Result<()> {
                     .context("procedural macro post processing callback failed")?;
             }
 
+            let span = trace_span!("drop_db");
+            {
+                let _guard = span.enter();
+                drop(db);
+            }
```
FroyaTheHen, line 254: why is that explicitly in separate scope?

scarb/src/bin/scarb/main.rs
```diff
@@ -46,6 +35,90 @@ fn main() -> ExitCode {
...
+            let s = var.as_str();
+            s == "true" || s == "1"
+        })
+        .unwrap_or(false);
+
+    let profile_layer = if tracing_profile {
+        let mut path = PathBuf::from(format!(
+            "./scarb-profile-{}.json",
+            SystemTime::UNIX_EPOCH.elapsed().unwrap().as_micros()
+        ));
```
FroyaTheHen, line 74: I am wondering if it wouldn't be more user friendly if something like for example utc formatted timestamp would be used instead of epoch so it is easier to navigate the files. Or is there some indication here i'm not aware of?

FroyaTheHen (approved):
awesome! ❤️

scarb/src/ops/compile.rs
maciektr: It takes suprisingly long (at least for me) to drop the db (e.g. 4 seconds for staking which compiles in 30s). We don't really use this observation now, but it's something to keep in mind for future.

scarb/src/bin/scarb/main.rs
maciektr: Changed.

PR #2348, opened 2025-06-27
scaffold oracle hint processor

mkaput:

PR #2351, opened 2025-06-30
reimplement `exec_replace`

mkaput:
This PR reimplements our `exec_replace` approach to fix two issues:
1. We have inconsistently added a `error: process did not exit successfully: exit code: 1` line only on Windows, and this was accounted for in tests in a very dirty way.
2. We never cared about destructing objects created up in the call stack. This PR actually fixes an unnoticed bug - calling `exec_replace` would result in broken (unterminated) trace files.

### DIFF
--- a/scarb/src/process.rs
+++ b/scarb/src/process.rs
@@ -28,11 +28,11 @@ pub use crate::internal::fsx::is_executable;
 /// In doing so (and by effectively ignoring it) we should emulate proxying Ctrl-C handling to
 /// the application at hand, which will either terminate or handle it itself.
 /// According to Microsoft's documentation at
-/// <https://docs.microsoft.com/en-us/windows/console/ctrl-c-and-ctrl-break-signals>.
-/// The Ctrl-C signal is sent to all processes attached to a terminal, which should include our
-/// child process. If the child terminates, then we will reap them in Cargo pretty quickly, and if
-/// the child handles the signal, then we won't terminate (and we shouldn't!) until the process
-/// itself later exits.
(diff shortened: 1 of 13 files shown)


mkaput:

- #2351 ◀
    - #2348

mkaput:
now I think we should just drop that error message and simulate exec syscall on Windows 🤔

scarb/src/process.rs
```diff
@@ -14,43 +14,82 @@ pub use crate::internal::fsx::is_executable;
...
 /// One aspect we fix here is that we specify a handler for the Ctrl-C handler.
 /// In doing so (and by effectively ignoring it) we should emulate proxying Ctrl-C handling to
 /// the application at hand, which will either terminate or handle it itself.
 /// According to Microsoft's documentation at
 /// <https://docs.microsoft.com/en-us/windows/console/ctrl-c-and-ctrl-break-signals>.
-/// the Ctrl-C signal is sent to all processes attached to a terminal, which should include our
-/// child process. If the child terminates then we'll reap them in Cargo pretty quickly, and if
-/// the child handles the signal then we won't terminate (and we shouldn't!) until the process
+/// The Ctrl-C signal is sent to all processes attached to a terminal, which should include our
+/// child process. If the child terminates, then we will reap them in Cargo pretty quickly, and if
```
maciektr:
```suggestion
/// child process. If the child terminates, then we will reap them in Scarb pretty quickly, and if
```
Should you mean Scarb? 🤔 I don't really understand what "reap" represents here, would you mind clarifying it for me a bit?

mkaput: done

PR #2353, opened 2025-06-30
scaffold oracle connection management

mkaput:


extensions/scarb-execute/src/oracle/hint_processor.rs
```diff
@@ -100,11 +107,33 @@ impl<'a> OracleHintProcessor<'a> {
...
+                .connect(&connection_string)?
+                .call(selector, calldata)
+        };
+
+        let result = invoke();
+
+        let mut response: Vec<Felt252> = vec![];
+        EncodableResult::from(result)
+            .encode(&mut response)
+            .expect("response encoding should never fail");
```
maciektr, line 154: Why? 🤔 Can't it depend on user input?

extensions/scarb-execute/src/oracle/connection.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl ConnectionManager {
+    pub fn new() -> Self {
+        Self(Default::default())
+    }
+
+    /// Establishes a connection to a given connection string and stores it in the connection pool.
+    ///
+    /// If the connection already exists in the pool, the existing connection is reused.
+    /// The same applies to connection errors, the pool will never reattempt to reconnect.
```
maciektr: Won't the first connection error break the execution anyway? 🤔 https://github.com/software-mansion/scarb/blob/de55b4d4f8811fb78da91bf17ccd26cdae13e11a/extensions/scarb-execute/src/oracle/hint_processor.rs#L81

Also, why do you want to avoid re-attempts? 🤔

mkaput: > Won't the first connection error break the execution anyway? 🤔

no, because user code can just ignore the error (I'm bubbling everything as `oracle::Result` here) and continue execution.

> Also, why do you want to avoid re-attempts? 🤔

hmm, this question actually highlighted to me that user code might attempt to use another oracle to fix the failing one 🤔 super unlikely but caching errors would indeed be weird here.

my original idea was to just cache the response for performance, but this is not a stateless system

extensions/scarb-execute/src/oracle/hint_processor.rs
mkaput: it depends on user input but that input is "serialized" into `let result = Result<Vec<Felt252>, anyhow::Error>` one line up here, and I & starknet-rs control full encoding process of this type and I know it never fails because nothing can fail there.

extensions/scarb-execute/src/oracle/connection.rs
mkaput: a17ae17c8e641b35fde03aaa58c0b2f4ef07d8b7

PR #2355, opened 2025-06-30
Reject dependency toml with invalid keywords

### LINKED ISSUE #130: Make deserialization of dependency specifications produce better error messages
See TODO comment, it describes more details:

https://github.com/software-mansion/scarb/blob/90372252a2888c4d4e27650a14194daf1b01162c/scarb/tests/e2e/git_source.rs#L194-L221

maciektr:
Previously, we've agreed to allow invalid keywords in dependency spec (see https://github.com/software-mansion/scarb/pull/2237#discussion_r2090618885). However, this seems to lead to a lot of hard to notice / debug errors, as users might mistype a spec. In such case, the project can still compile, but using a different revision than intended (e.g. if you type "commit" instead of "rev"). 
This PR disallows using invalid keywords in dependency spec and emits errors when parsing instead. 

fix #130

### DIFF
--- a/scarb/src/core/manifest/toml_manifest.rs
+++ b/scarb/src/core/manifest/toml_manifest.rs
@@ -271,45 +271,44 @@ mod serdex {
     use crate::core::{
         DetailedTomlDependency, MaybeWorkspace, TomlDependency, TomlWorkspaceDependency,
     };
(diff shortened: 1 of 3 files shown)


FroyaTheHen (approved):
lgtm 👌

PR #2356, opened 2025-07-01
[Linter] Fail on --deny-warnings after all packages were linted

wawel37:
Resolves https://github.com/software-mansion/cairo-lint/issues/375

### DIFF
--- a/scarb/src/ops/lint.rs
+++ b/scarb/src/ops/lint.rs
@@ -5,7 +5,7 @@ use crate::{
         CompilationUnit, CompilationUnitAttributes,
         db::{ScarbDatabase, build_scarb_root_database},
     },
-    core::{PackageId, TargetKind},
+    core::{PackageId, PackageName, TargetKind},
     ops,
 };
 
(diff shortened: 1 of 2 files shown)


scarb/src/ops/lint.rs
```diff
@@ -223,20 +227,25 @@ pub fn lint(opts: LintOptions, ws: &Workspace<'_>) -> Result<()> {
...
+                        was_error_in_package = true;
                     }
+                    diagnostics_per_cu.push((db, diagnostics));
+                }
+            }
+        }
+        if was_error_in_package {
+            return Err(anyhow!(
+                "lint checking `{package_name}` failed due to previous errors"
+            ));
```
piotmag769: Shouldn't we fail after all selected packages are linted? Idk though

wawel37: Was thinking about it, but wasn't sure. But yup, i think it's better to just fail after all the packages

wawel37:
Regarding the whole lint `lint.rs` file readability, it's awful and the task for the refactor is on the way https://github.com/software-mansion/cairo-lint/issues/376

scarb/src/ops/lint.rs
```diff
@@ -223,24 +236,53 @@ pub fn lint(opts: LintOptions, ws: &Workspace<'_>) -> Result<()> {
...
-
-                    if opts.fix {
-                        let fixes = get_fixes(&db, diagnostics);
-                        for (file_id, fixes) in fixes.into_iter() {
-                            ws.config()
-                                .ui()
-                                .print(Status::new("Fixing", &file_id.file_name(&db)));
-                            apply_file_fixes(file_id, fixes, &db, formatter_config.clone())?;
-                        }
+                        packages_with_error.push(package_name.clone());
```
maciektr, line 239: Won't this duplicate the package name, if a package has multiple CUs that are checked? E.g. one for integration and one for unit tests?

PR #2366, opened 2025-07-03
implement feature use yanked packages only if cached

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2178
connected to https://github.com/software-mansion/scarbs.xyz/pull/532

### DIFF
--- a/scarb/src/core/lockfile.rs
+++ b/scarb/src/core/lockfile.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::str::FromStr;
 
 use anyhow::{Context, Result, anyhow};
@@ -182,6 +182,19 @@ impl Lockfile {
         }
 
         Ok(doc)
+    }
+
+    /// Creates a set of [`PackageId`]s with source [`crate::core::source::SourceKind::Registry`] present in Lockfile.
+    pub fn create_yanked_whitelist(&self) -> HashSet<PackageId> {
+        self.packages
+            .values()
+            .flatten()
+            .filter_map(|p| {
+                p.source
+                    .filter(|source| source.is_registry())
+                    .map(|source| PackageId::new(p.name.clone(), p.version.clone(), source))
+            })
+            .collect::<HashSet<_>>()
     }
 
     pub fn render(&self) -> Result<String> {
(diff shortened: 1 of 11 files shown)


maciektr:
Please note that it should not care whether the yanked package is in cache or not -> but rather, if it's in the lockfile or not. 
It should still be valid, to use a yanked version when you freshly download project to your computer, as long as it's in the lockfile. 
This means that we would probably want to check it higher up in the abstractions we have. For instance, Cargo does it in `RegistrySource` implementation (https://github.com/rust-lang/cargo/blob/5f900981de60ce4df51f3161d5505fedbf179d1d/src/cargo/sources/registry/mod.rs#L796). 
The lockfile is accessible throughout the resolution https://github.com/software-mansion/scarb/blob/8c153113232a48dce86e7362b11d9ad052e81958/scarb/src/ops/resolve.rs#L313

scarb/src/core/registry/index/record.rs
```diff
@@ -15,6 +15,8 @@ pub struct IndexRecord {
     pub checksum: Checksum,
     #[serde(default = "default_false", skip_serializing_if = "is_false")]
     pub no_core: bool,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub yanked: Option<bool>,
```
maciektr: It should just deserialize to bool, since there is no difference between `None` and `Some(false)` (you can just copy the line from `no_core`).

scarb/src/core/lockfile.rs
```diff
@@ -184,6 +184,25 @@ impl Lockfile {
         Ok(doc)
     }
 
+    /// Creates a set of [`PackageId`]s present in Lockfile.
+    pub fn create_yanked_whitelist(&self) -> HashSet<PackageId> {
+        self.packages
+            .values()
+            .flatten()
+            .filter_map(|p| {
+                if p.source.is_some() {
```
maciektr: nit If this method is specialized for yanked whitelist, we could even filter to only keep reigstry sources, as yanking has no meaning for git ones (and paths are dropped already) 🤔

scarb/src/ops/resolve.rs
```diff
@@ -301,21 +301,23 @@ pub fn resolve_workspace_with_opts(
...
-
             let members_summaries = ws
                 .members()
                 .map(|pkg| pkg.manifest.summary.clone())
                 .collect::<Vec<_>>();
 
-            let lockfile: Lockfile = if opts.update {
-                Lockfile::new([])
+            let (lockfile, yanked_whitelist) = if opts.update {
+                (Lockfile::new([]), HashSet::new())
```
maciektr, line 310: 👍 
We can doc somewhere that update will invalidate usage of any yanked versions (e.g. https://docs.swmansion.com/scarb/docs/reference/lockfile.html#updating-locked-versions)

PR #2369, opened 2025-07-04
Calculate source checksums in fingerprint

maciektr:


maciektr:

- #2340
    - #2369 ◀

scarb/src/compiler/incremental/source.rs
```diff
@@ -0,0 +1,123 @@
...
+                return false;
+            }
+
+            // Skip `target` directory.
+            if entry.depth() == 1
+                && ({
+                    let f = entry.file_name();
+                    f == DEFAULT_TARGET_DIR_NAME
+                })
+            {
```
DelevoXDG:
```suggestion
            if entry.depth() == 1 && entry.file_name() == DEFAULT_TARGET_DIR_NAME {
```

DelevoXDG: Also, isn't `{package}/src/target`, unlike `{package}/target`, a valid path that could contain source files?

maciektr: Yes, but `entry.depth() == 1` guards against that

DelevoXDG: `{package}/src/target` does have `depth == 1`, though 🤔 

At least it appeared so after building one of the examples

maciektr: Depth == 0: 
- {package}

Depth == 1: 
- {package}/target
- {package}/src

Depth == 2: 
- {package}/src/target

maciektr: This should go like this. I'll recheck in runtime and let you know

maciektr: Ahh, disregard everything I was saying earlier. 
I forgot we are passing source paths here, not package roots!
You are right, this is wrong. Thanks!

maciektr: Changed.

scarb/src/compiler/incremental/source.rs
```diff
@@ -0,0 +1,109 @@
...
+pub(crate) fn create_local_fingerprints(source_paths: Vec<&Utf8Path>) -> Vec<LocalFingerprint> {
+    let filter = {
+        move |entry: &DirEntry| -> bool {
+            // Skip any subdirectories containing `Scarb.toml`.
+            if entry.path().join(MANIFEST_FILE_NAME).exists() {
+                return false;
+            }
+            true
+        }
+    };
```
DelevoXDG: Nit:

```suggestion
    // Skip any subdirectories containing `Scarb.toml`.
    let filter = |entry: &DirEntry| !entry.path().join(MANIFEST_FILE_NAME).exists();
```

If you believe we're likely to change this though, you may ignore this

scarb/src/compiler/incremental/source.rs
```diff
@@ -0,0 +1,109 @@
...
+
+            let Ok(path) = path.try_to_utf8() else {
+                warn!("failed to convert path to UTF-8: {}", path.display());
+                return Continue;
+            };
+
+            let Ok(content) = fsx::read_to_string(&path) else {
+                warn!("failed to read file: {}", path.to_string());
+                return Continue;
+            };
```
DelevoXDG: Did you consider what happens if, for example, there's a permission issue with the file, the file is corrupted, or the file got deleted before the walker runs?

While some of these are more and less realistic, I worry that silently ignoring such failures with `Continue` may produce stale cache files 🤔 

I wonder if it make sense to instead produce an `Error` in such circumstance, display a warning to the user, and treat such occurrence as a cache miss.
 
WDYT?

maciektr: Good question

Generally, I don't think it is completely safe to turn problems here into hard errors either, as we traverse any files under `/src` dir -  whether they are part of the package source code or not. At this point in compilation, we have no way of knowing, which files will be part of the module tree - determined by `mod` keywords basically. 

> Did you consider what happens if, for example, there's a permission issue with the file, the file is corrupted, or the file got deleted before the walker runs?

In a sense, this does not need to be a problem for us: 

If a file has been deleted before run -> it has been previously in the hash and now will be skipped -> hash will change -> cache will be overwritten. 

If a file is corrupt / has permissions that deny us from accessing it -> we won't be able to load it during compilation neither. (and if it has been accessible before, the hash will change, like in the previous example). 

It's possible that there is some edge case where it is a problem, but not sure if we have a better way of handling this 🤔

DelevoXDG: You make a good argument, I suppose you're right that it won't hurt us as much.

I would still consider printing a warning to UI rather than ignoring these silently, which will allow users to at least troubleshoot the problem, possibly determining reason for cache misses. Since there would be too many warnings, just one that lets the user know there was an issue with fingerprinting source files IMO would be sufficient, with rest checkable via tracing. WDYT?

maciektr: Yeah, I can do that. Changed!

scarb/src/compiler/incremental/source.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    let mut fingerprints = Vec::new();
+    while let Ok((path, checksum)) = rx.recv() {
+        fingerprints.push(LocalFingerprint { path, checksum });
+    }
+
+    if any_missed.load(Ordering::Acquire) {
+        ui.warn(formatdoc! {r#"
+            some files were skipped when calculating source checksums for `{target_name}` component
+            this may or may not affect the build
```
DelevoXDG: Nit:
```suggestion
            this may or may not affect the build and incremental compilation cache loading 
```

PR #2373, opened 2025-07-07
Caclulate digests for proc macro plugin fingerprints

maciektr:

PR #2372, opened 2025-07-07
Cache shared_lib_path function for proc macros (call to cargo metadata)

maciektr:

PR #2371, opened 2025-07-07
Fix incremental cache file name, update test case

maciektr:

PR #2382, opened 2025-07-09
Do not calculate diagnostics for cached crates

maciektr:


maciektr:

- #2371
    - #2340
        - #2369
            - #2372
                - #2373
                    - #2374
                        - #2382 ◀

scarb/src/compiler/helpers.rs
```diff
@@ -77,7 +88,12 @@ pub fn build_compiler_config<'c>(
             };
         }
     })
-    .with_ignore_warnings_crates(&ignore_warnings_crates);
+    .with_ignore_warnings_crates(&ignore_warnings_crates)
+    // If a crate is cached, we do not need to check it for diagnostics,
+    // as the cache can only be produced if the crate is diagnostic-free.
+    // So if there were any diagnotics here to show, it would mean that the cache is outdated - thus
+    // we should not use it in the first place.
```
DelevoXDG, line 92: Is this accurate in case of warnings too? 🤔

maciektr: Rewritten to always ensure main crate id is included.
Since we hide warnings for dependencies, we can ignore them.

scarb/src/compiler/helpers.rs
```diff
@@ -77,7 +85,12 @@ pub fn build_compiler_config<'c>(
...
```
DelevoXDG, line 92: Following up: https://github.com/software-mansion/scarb/pull/2382/files#r2205169439
Should we update this comment to clarify it applies to deps and not main crate?

DelevoXDG (approved):

scarb/src/compiler/helpers.rs
maciektr: Added.

PR #2386, opened 2025-07-10
add minimal JSON-RPC types implementation

mkaput:
It is surprising how complex existing JSON-RPC Rust implementations are, we don't need any kinds of async, extendibility or network stuff that they provide.

### DIFF
--- a/extensions/scarb-execute/src/oracle/jsonrpc.rs
+++ b/extensions/scarb-execute/src/oracle/jsonrpc.rs
@@ -1,0 +1,266 @@
+//! Simple JSON-RPC types that suit our needs and nothing more.
+//! It is surprising how complex existing JSON-RPC Rust implementations are,
+//! we don't need any kinds of async, extendibility or network stuff that they provide.
+
+use anyhow::{Result, bail};
+use serde::de::{Unexpected, Visitor};
+use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
+use std::fmt;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(untagged)]
+pub enum Message {
+    Request(Request),
+    Response(Response),
+    Notification(Notification),
+}
+
+impl Message {
+    pub fn kind(&self) -> &'static str {
+        match self {
+            Message::Request(_) => "request",
+            Message::Response(_) => "response",
+            Message::Notification(_) => "notification",
+        }
+    }
+
+    pub fn expect_request(self) -> Result<Request> {
+        match self {
+            Message::Request(request) => Ok(request),
+            _ => bail!("expected request, got {}", self.kind()),
+        }
+    }
+
+    pub fn expect_response(self) -> Result<Response> {
+        match self {
+            Message::Response(response) => Ok(response),
+            _ => bail!("expected response, got {}", self.kind()),
+        }
+    }
+
+    pub fn expect_notification(self) -> Result<Notification> {
+        match self {
+            Message::Notification(notification) => Ok(notification),
+            _ => bail!("expected notification, got {}", self.kind()),
+        }
+    }
+}
+
+impl From<Request> for Message {
+    fn from(request: Request) -> Message {
+        Message::Request(request)
+    }
+}
+
+impl From<Response> for Message {
+    fn from(response: Response) -> Message {
+        Message::Response(response)
+    }
+}
+
+impl From<Notification> for Message {
+    fn from(notification: Notification) -> Message {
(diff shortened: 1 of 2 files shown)


extensions/scarb-execute/src/oracle/jsonrpc.rs
```diff
@@ -0,0 +1,268 @@
...
+
+#![allow(dead_code, unused)] // TODO: Remove in next PRs
+
+use anyhow::{Result, bail};
+use serde::de::{Unexpected, Visitor};
+use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
+use std::fmt;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(untagged)]
```
maciektr, line 13: Should we bail from parsing on additional keys in the messages, since it's out of spec to send them outside params/data/result fields?

extensions/scarb-execute/src/oracle/jsonrpc.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct Request {
+    pub jsonrpc: TwoPointZero,
+    pub id: Id,
+    pub method: String,
+    #[serde(default = "serde_json::Value::default")]
+    #[serde(skip_serializing_if = "serde_json::Value::is_null")]
+    pub params: serde_json::Value,
```
maciektr, line 100: Should we ensure somewhere, that this is `Value::Object`, as spec says "A **Structured** value"?

extensions/scarb-execute/src/oracle/jsonrpc.rs
```diff
@@ -0,0 +1,268 @@
...
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct Response {
+    pub jsonrpc: TwoPointZero,
+    pub id: Id,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub result: Option<serde_json::Value>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub error: Option<ResponseError>,
+}
```
maciektr, line 111: Maybe w should validate during ser/de that this never have both error and result?

Or at least disallow creating this struct directly and expose some `::error()` / `::result()` constructors?

extensions/scarb-execute/src/oracle/jsonrpc.rs
mkaput: For `"ready"` request that will come from the oracle, I am going to do this check manually at the call site, so it's done there.

Otherwise, we are the ones sending requests, so we control the request generation logic.

extensions/scarb-execute/src/oracle/jsonrpc.rs
mkaput: I would have to write a manual serde implementation for these types in place of `#[serde(untagged)]` in order to achieve this :/ That's why I've decided to just ignore this problem.

maciektr: Can't we just add #[serde(deny_unknown_fields)] to all of the variants? 🤔

mkaput: I believe we can't because `#[serde(untagged)]` attempts to deserialize each variant and picks the first one that deserializes successfully. If a valid variant gets an unknown field, then it will fail at `untagged` stage and the error message will be unhelpful.

extensions/scarb-execute/src/oracle/jsonrpc.rs
mkaput: > Maybe w should validate during ser/de that this never have both error and result?

Would have to write custom serde impl, IMHO too much code for the scope of this thing.

> Or at least disallow creating this struct directly and expose some ::error() / ::result() constructors?

That's what `MessageFactory` is already for :)

maciektr: > That's what MessageFactory is already for :)

But you do not enforce it's usage

maciektr: > Would have to write custom serde impl, IMHO too much code for the scope of this thing.

For deserialization, could be simple `serde(try_from` 🤔

mkaput: discussed offline

maciektr (approved):
agreed to limit validations, since this is for internal use only

PR #2381, opened 2025-07-08
implement `StdioJsonRpcConnection`

mkaput:

PR #2390, opened 2025-07-10
future-proof oracle error types by treating it as an enum

mkaput:


mkaput:

- #2351
    - #2348
        - #2353
            - #2379
                - #2380
                    - #2386
                        - #2387
                            - #2381
                                - #2390 ◀

extensions/scarb-execute/src/oracle/encodable_result.rs
```diff
@@ -23,11 +23,12 @@ where
...
         match self {
             EncodableResult::Ok(value) => {
-                writer.write(Felt::ZERO);
+                writer.write(Felt::ZERO); // Result::Ok marker
                 value.encode(writer)
             }
             EncodableResult::Err(value) => {
-                writer.write(Felt::ONE);
+                writer.write(Felt::ONE); // Result::Err marker
+                writer.write(Felt::ZERO); // oracle::ErrorInner::ErrorMessage marker
```
maciektr, line 31: Sorry, I don't understand this. 
Can we talk about it more? 🤔

mkaput: This covers the fact that on the `oracle` package side I have wrapped the error ByteArray into an intermediary enum: https://github.com/software-mansion/cairo-oracle/blob/378b57a5f7b46cab451f954da67fec208d50a798/oracle/src/lib.cairo#L132-L139

The idea is to allow me to introduce new fields for this after the byte array by treating enum variant tag as kind of a "version".

Though I think I messed up the `oracle` side of this, because the Serde impl there needs to fall back to `ErrorInner::ErrorMessage` variant in case the tag is nonzero 🤔

mkaput: yeah, scrap this PR, it's nonsense

PR #2397, opened 2025-07-14
Include dependencies in fingerprint id

maciektr:


maciektr:
Merged into https://github.com/software-mansion/scarb/pull/2369

PR #2408, opened 2025-07-16
pipe oracle stderr as debug level instead of trace

mkaput:

PR #2409, opened 2025-07-16
fix oracle result encoding

mkaput:
The previous encoding logic had a tricky bug: for `Result<T, ByteArray>` it emitted unwanted felt which denoted
length of `T`. To visualize, it emitted:

```cairo
[0, 13, <13 felts that should decode into T>]
```

instead of

```cairo
[0, <13 felts that should decode into T>]
```

This bug was not repeated in `Err` encoding.


mkaput:

- #2408
    - #2409 ◀

extensions/scarb-execute/src/oracle/encodable_result.rs
```diff
@@ -1,37 +0,0 @@
...
-impl<T, E> Encode for EncodableResult<T, E>
-where
-    T: Encode,
-    E: fmt::Debug,
-{
-    fn encode<W: FeltWriter>(&self, writer: &mut W) -> Result<(), CodecError> {
-        match self {
-            EncodableResult::Ok(value) => {
-                writer.write(Felt::ZERO);
-                value.encode(writer)
```
mkaput, line 28: here was the bug. `self` was `Vec<Felt>`, so when encoded it included vector length

extensions/scarb-execute/src/oracle/hint_processor.rs
```diff
@@ -148,14 +147,24 @@ impl<'a> OracleHintProcessor<'a> {
...
-            .expect("response encoding should never fail");
-
-        res_segment.write_data(response.into_iter())?;
+        // Encode the result as Result<[felt252; N], ByteArray>. The oracle package interprets this
+        // as Result<T, oracle::Error>, where T is user-defined, and oracle::Error has an implicit
+        // invariant that it can always deserialise from encoded byte arrays.
+        match invoke() {
+            Ok(result) => {
+                res_segment.write(Felt252::ZERO)?;
+                res_segment.write_data(result.into_iter())?;
```
mkaput, line 156: this fix works by directly writing data to memory, without encoding roundtrip

PR #2417, opened 2025-07-18
Add docs for patch and features

maciektr:
- **Add dependency features docs**
- **Add dependency overrides docs**
- **Add context for source patch url parse error**

### DIFF
--- a/website/docs/reference/overriding-dependencies.md
+++ b/website/docs/reference/overriding-dependencies.md
@@ -1,0 +1,43 @@
+# Overriding Dependencies
+
+Overriding dependencies is a powerful feature in Scarb that allows you to replace all usages of some dependency with
+another revision, coming from a different source, during the workspace resolution.
+
+The desire to override a dependency can arise through a number of scenarios.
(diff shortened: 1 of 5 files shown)


website/docs/reference/overriding-dependencies.md
```diff
@@ -0,0 +1,43 @@
+# Overriding Dependencies
+
+Overriding dependencies is a powerful feature in Scarb that allows you to replace all usages of some dependency with
+another revision, coming from a different source, during the workspace resolution.
```
mkaput: we support same-source overrides

maciektr: Hmm, I meant it to hint you to the use case, rather than to suggest it's not possible to override the same source. 
Maybe:
Overriding dependencies is a powerful feature in Scarb that allows you to replace all usages of some dependency with
another revision, **for example** coming from a different source, during the workspace resolution.
Wdyt?

mkaput: that's much better

PR #2419, opened 2025-07-20
Add dependency cfg_set to fingerprint id

maciektr:


scarb/src/compiler/incremental/fingerprint.rs
```diff
@@ -303,14 +303,28 @@ impl Fingerprint {
...
-            .map(|dep| dep.component_discriminator.clone())
-            .sorted()
+            .sorted_by_key(|dep| dep.component_discriminator.clone())
         {
-            component_discriminator.hash(&mut hasher);
+            dep.component_discriminator.hash(&mut hasher);
+            let dep_fingerprint = dep.fingerprint.upgrade()
+                .unwrap_or_else(|| unreachable!(
+                "dependency fingerprint should never be dropped, as long as unit fingerprint is alive"
+            ));
```
mkaput: just `.expect` here, the outcome will be the same

mkaput (approved):

PR #2424, opened 2025-07-21
Create artifacts writer

maciektr:
Create artifacts writer

Move serialization of .test.sierra.json artifact to separate thread

Move serialization of .sierra.json artifact to separate thread

Move serialization of .sierra artifact to separate thread


maciektr:
I noticed that writing the artifact files can take suprisingly long. This PR moves the serialization / write to a separate thread, so it can happen while we generate inc compilation caches and wait for the salsa db to drop. See the screenshots below. 

Before: 
image

After: 
image

Before: 
image


After:
image

scarb/src/ops/compile.rs
```diff
@@ -253,6 +262,8 @@ fn compile_unit_inner(unit: CompilationUnit, ws: &Workspace<'_>) -> Result<()> {
                 drop(db);
             }
 
+            artifacts_writer.join()?;
```
mkaput: maybe wrap this in a tracing span so we can see on this thread that is has been hanging waiting for artifacts writing to finish?

scarb/src/internal/artifacts_writer.rs
```diff
@@ -0,0 +1,117 @@
...
+    handle: Option<thread::JoinHandle<Result<()>>>,
+}
+
+impl ArtifactsWriter {
+    pub fn new(request_stream: mpsc::Receiver<Request>, ws: &Workspace<'_>) -> Self {
+        let ws = unsafe {
+            // This should be safe, as we know we will join the artifact writer thread before
+            // the workspace is dropped.
+            mem::transmute::<&Workspace<'_>, &Workspace<'_>>(ws)
+        };
```
mkaput:
```suggestion
pub struct ArtifactsWriter<'a> {
    handle: Option<thread::JoinHandle<Result<()>>>,
    _phantom_data: PhantomData<Workspace<'a>>,
}

impl<'a> ArtifactsWriter<'a> {
    pub fn new(request_stream: mpsc::Receiver<Request>, ws: &Workspace<'a>) -> Self {
```

let's enforce this assumption in the type system

scarb/src/internal/artifacts_writer.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    pub fn join(mut self) -> Result<()> {
+        let result = if let Some(handle) = self.handle.take() {
+            handle
+                .join()
+                .expect("failed to join artifacts writer thread")
+        } else {
+            Ok(())
+        };
```
mkaput: nit, `match` would be shorter here

scarb/src/internal/artifacts_writer.rs
```diff
@@ -0,0 +1,117 @@
...
+            // the workspace is dropped.
+            mem::transmute::<&Workspace<'_>, &Workspace<'_>>(ws)
+        };
+        let handle = thread::Builder::new()
+            .name("scarb-artifacts-writer".into())
+            .spawn(move || {
+                let span = trace_span!("writer requests");
+                for request in request_stream.iter() {
+                    let _guard = span.enter();
+                    handle_request(request, ws)
```
mkaput: why not wrap `handle_request` with `#[tracing::instrument]` instead?

scarb/src/internal/artifacts_writer.rs
```diff
@@ -0,0 +1,117 @@
+use crate::compiler::helpers::{write_json, write_string};
+use crate::core::Workspace;
+use crate::flock::Filesystem;
+use anyhow::{Context, Result};
+use cairo_lang_sierra::program::{ProgramArtifact, VersionedProgram};
+use std::sync::{Arc, mpsc};
+use std::{mem, thread};
+use tracing::trace_span;
+
+pub enum Request {
```
mkaput: I wonder: this could be quite nicely generalised. We can just make this object to be able to _offload something_. This way the logic for actual writing would not be spread out between scheduling logic and this file. To do this, we can base the implementation on sendable FnOnces:

```rust
type BoxedWorkload<'a> = Box<dyn FnOnce(&'a Workspace<'a>) -> Result<()> + Send>;

struct Offloader<'a> { ... }

impl<'a> Offloader<'a> {
    pub fn new(request_stream: mpsc::Receiver<(&'static str, BoxedWorkload<'a>)>, ws: &'a Workspace<'a>) -> Self { ... }

    pub fn offload(&self, what: &'static str, workload: impl FnOnce(&'a Workspace<'a>) -> Result<()> + Send) { ... }
}
```

scarb/src/compiler/repository.rs
```diff
@@ -48,6 +49,7 @@ impl CompilerRepository {
     pub fn compile(
         &self,
         unit: CairoCompilationUnit,
+        artifacts_writer: mpsc::Sender<Request>,
```
mkaput: would be sane to provide typealias for this mpsc sender

scarb/src/ops/compile.rs
maciektr: It's kind of redundant, as we will be tracing requests inside it anyway, but I guess won't hurt :D

scarb/src/compiler/compilers/test.rs
```diff
@@ -85,9 +87,23 @@ impl Compiler for TestCompiler {
         let span = trace_span!("serialize_test");
         {
             let _guard = span.enter();
-            let sierra_program: VersionedProgram = test_compilation.sierra_program.clone().into();
-            let file_name = format!("{}.test.sierra.json", unit.main_component().target_name());
-            write_json(&file_name, "output file", &target_dir, ws, &sierra_program)?;
+            {
```
mkaput, line 90: why this nested block?

scarb/src/internal/offloader.rs
```diff
@@ -0,0 +1,74 @@
...
+    _phantom_data: PhantomData<Workspace<'a>>,
+    sink: OffloaderSink<'a>,
+}
+
+impl<'a> Offloader<'a> {
+    pub fn new(scope: &'a Scope<'a, '_>, ws: &'a Workspace<'a>) -> Self {
+        let ws = unsafe {
+            // This should be safe, as we know we will join the artifact writer thread before
+            // the workspace is dropped.
+            mem::transmute::<&Workspace<'_>, &Workspace<'_>>(ws)
```
mkaput: clippy will probably yell at this, but it'd be good to explicilyt name the lifetimes here

mkaput (approved):

scarb/src/compiler/compilers/test.rs
maciektr: Just so I can do `let target_dir = target_dir.clone();` shadowing the outer var.

mkaput (approved):

PR #2426, opened 2025-07-21
Include all component deps in fingerprint id

maciektr:


scarb/tests/incremental.rs
```diff
@@ -175,17 +172,9 @@ fn deps_are_fingerprinted() {
...
-    assert_eq!(fingerprints().len(), 8); // core, first, second, second, third, third, fourth, fifth
-    assert_eq!(digest(second_component_id.as_str()), second_digest);
-    let new_second_component_id = fingerprints()
-        .iter()
-        .find(|t| t.starts_with("second-") && t.as_str() != second_component_id)
-        .unwrap()
-        .to_string();
-    assert_ne!(digest(new_second_component_id.as_str()), second_digest);
+    // Note we have changed the edition of the third.
+    // Since editions are part of the fingerprint id, the third will change its id.
```
DelevoXDG, line 180: I'm not sure we actually verify this here? 😄

maciektr: The number of files is asserted to be changed. I can add value comparison.

maciektr: Done

PR #2431, opened 2025-07-22
Set dev inlining strategy to avoid

maciektr:


maciektr:
> What is the reason for those changes?? 

Basically, the compiler perfomance. 

> Also, shouldn't we consider updating Scarb documentation for inlining-strategy with default values, I think it could be somehow misleading when the default strategy isn't default?

Good point, changed.

PR #2436, opened 2025-07-23
Change error message and add a warning for executable target

maciektr:
- **Change error message in executable compilation**
- **Add a warning**


Arcticae (approved):
🙏

PR #2446, opened 2025-07-24
Use new linter group for linter diags

wawel37:
Related to https://github.com/software-mansion/cairo-lint/issues/406

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,7 @@ cairo-lang-test-runner = "*"
 cairo-lang-utils = { version = "*", features = ["env_logger"] }
 cairo-language-server = "*"
 cairo-lint = "*"
-cairo-vm = "2.2.0"
+cairo-vm = "2.3.1"
 camino = { version = "1", features = ["serde1"] }
 cargo_metadata = ">=0.18"
 chrono = "0.4"
@@ -114,7 +114,7 @@ proc-macro2 = "1"
 pubgrub = { git = "https://github.com/software-mansion-labs/pubgrub.git", branch = "dev" }
 quote = "1"
 ra_ap_toolchain = "0.0.218"
-redb = "2.6.0"
+redb = "2.6.1"
 reqwest = { version = "0.11", features = ["gzip", "brotli", "deflate", "json", "stream", "multipart"], default-features = false }
 salsa = { package = "rust-analyzer-salsa", version = "0.17.0-pre.6" }
 semver = { version = "1", features = ["serde"] }
@@ -158,7 +158,7 @@ typetag = "0.2"
 url = { version = "2", features = ["serde"] }
 walkdir = "2"
 which = "8"
-windows-sys = { version = "0.59", features = ["Win32_Foundation", "Win32_System", "Win32_System_Console"] }
+windows-sys = { version = "0.60", features = ["Win32_Foundation", "Win32_System", "Win32_System_Console"] }
 xshell = "0.2"
 xxhash-rust = { version = "0.8", features = ["xxh3"] }
 zip = { version = "0.6", default-features = false, features = ["deflate"] }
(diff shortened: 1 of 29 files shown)


Cargo.toml
```diff
@@ -196,8 +196,8 @@ cairo-lang-test-plugin = { git = "https://github.com/starkware-libs/cairo", rev
 cairo-lang-test-runner = { git = "https://github.com/starkware-libs/cairo", rev = "7675b74815370e6c16760e8b059ae5d773cade88" }
 cairo-lang-test-utils = { git = "https://github.com/starkware-libs/cairo", rev = "7675b74815370e6c16760e8b059ae5d773cade88" }
 cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo", rev = "7675b74815370e6c16760e8b059ae5d773cade88" }
-cairo-language-server = { git = "https://github.com/software-mansion/cairols", rev = "7ecba29dd8a72690af0c5d2c533a48e291662f2c" }
-cairo-lint = { git = "https://github.com/software-mansion/cairo-lint", rev = "3cb99ece36b89338a2efd300a88ac22ebdb5a70e" }
+cairo-language-server = { git = "https://github.com/software-mansion/cairols", rev = "79eab67cfd9e461a71e4144bf0c2a093ce8bbf53" }
```
wawel37: This points to https://github.com/software-mansion/cairols/pull/845 PR in ls. Will change it as soon as we merge that

scarb/src/compiler/db.rs
```diff
@@ -78,12 +78,12 @@ pub(crate) fn build_scarb_root_database(
 }
 
 #[cfg(feature = "scarb-lint")]
-fn append_lint_plugin(suite: &mut PluginSuite) {
+pub(crate) fn append_lint_plugin(suite: &mut PluginSuite) {
```
piotmag769, line 82: It is not released, no reason to not make it just `pub`

scarb/src/ops/lint.rs
```diff
@@ -326,3 +367,110 @@ fn find_integration_test_package_id(package: &Package) -> Option<PackageId> {
             .for_test_target(target.group_id.clone().unwrap_or(target.name.clone()))
     })
 }
+
+fn build_lint_database(
+    unit: &CairoCompilationUnit,
+    ws: &Workspace<'_>,
+) -> Result<LinterAnalysisDatabase> {
```
piotmag769, line 375: Does it differ from `build_scarb_root_database` with anything? If not maybe we can make this function generic? Or maybe just make `ScarbDatabase` implement `LinterGroup`?

scarb/src/compiler/db.rs
maciektr: It's okay either way, I wouldnt pick on that.

scarb/src/ops/lint.rs
wawel37: 1. Scarb database also includes proc macros, we dont need that in linter.
2. `ScarbDatabase` isn't actual database, it just includes the `RootDatabase` from compiler. And we don't want to make this rootdatabase implement linter db

piotmag769: Can't we make `build_root_database` which we would reuse here and in `build_scarb_root_database`?

wawel37: We would have to define a trait for `RootDatabaseBuilder` and `LinterDatabaseBuilder` and make it implemented in the compiler as well as inside the linter. Imho it's not worth the trouble. Or am I missing some easier implementation xd

piotmag769: Right, add a comment here and in `build_scarb_database` that the two need to be kept in sync. Imo worth adding an issue tbh, this is unmaintainable long term

piotmag769 (approved):

maciektr (approved):

PR #2447, opened 2025-07-24
scarb-execute: extend resource info with memory segments & improve formatting

m-kus:
This PR adds several extra fields that are useful for benchmarking:
- Memory segments size
- Max memory segment address

Also formatting is changed to improve readability:
- One property per line (easier to calculate diff)
- Colored values, thousand separators

image


maciektr (approved):
Thanks!!! <3

PR #2448, opened 2025-07-25
Rewrite token stream spans to start at zero, for optimized get_text method

maciektr:
- **Use 2.12.0-rc.2 parser in proc macro tests**
- **Pass whole token stream span to the span adapter**
- ** Rewrite token stream spans to start at zero**

This has to be done, as get_text method in the compiler now works by slicing the file content with the spans, so we cannot start the parser with non-zero initial offset for new virtual files with token content only.

### DIFF
--- a/scarb/src/compiler/plugin/proc_macro/v2/host/attribute/span_adapter.rs
+++ b/scarb/src/compiler/plugin/proc_macro/v2/host/attribute/span_adapter.rs
@@ -12,51 +12,38 @@
 //!
 //! See [`crate::compiler::plugin::proc_macro::v2::ProcMacroHostPlugin::parse_attribute`] for more context.
 //!
-//! We mitigate this problem, by following logic:
+//! We mitigate this problem by following logic:
 //! Spans in the expansion input and code mappings generated from the expansion output are moved
-//! around, as if the expandable attribute was the first attribute in the attributes list.
+//! around, as if the expandable attribute was the last token in the built token stream.
 //! *Note that no code is actually rewritten - only the corresponding token spans are modified.*
 //! Input `TokenStream` is built by the following rules:
-//! - All spans of tokens before the expandable attribute are moved towards the end of the file
-//!   by the expandable attribute length.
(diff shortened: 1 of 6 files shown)


cptartur (commented):
ACK

utils/scarb-test-support/src/cairo_plugin_project_builder.rs
```diff
@@ -128,11 +128,11 @@ impl CairoPluginProjectBuilder {
     }
 
     pub fn add_cairo_lang_parser_dep(self) -> Self {
-        self.add_dep(r#"cairo-lang-parser = { git = "https://github.com/starkware-libs/cairo.git", rev = "172195a5e8c889a2355c28f5c444e61da6ae8a2b" }"#)
+        self.add_dep(r#"cairo-lang-parser = { git = "https://github.com/starkware-libs/cairo.git", rev = "v2.12.0-rc.2" }"#)
```
piotmag769: rc3?

scarb/src/compiler/plugin/proc_macro/v2/host/attribute/span_adapter.rs
```diff
@@ -172,17 +163,24 @@ impl ExpandableAttrLocation {
         let attr_start = self.start_offset_with_trivia();
         let attr_end = self.end_offset_with_trivia();
         let attr_width = self.width_with_trivia();
```
piotmag769, line 165: [nit] not moving trivia

scarb/src/compiler/plugin/proc_macro/v2/host/attribute/span_adapter.rs
```diff
@@ -160,7 +151,7 @@ impl ExpandableAttrLocation {
 
     pub fn adapted_call_site(&self) -> AdaptedTextSpan {
         let start =
-            self.item_start_offset + self.span_without_trivia.start - self.span_with_trivia.start;
+            self.whole_item_span.end - self.width_without_trivia() - self.whole_item_span.start;
```
piotmag769, line 154: [nit] not moving trivia

PR #2449, opened 2025-07-25
separate cache dirs by root manifest

FroyaTheHen:


scarb/tests/build.rs
```diff
@@ -41,11 +43,18 @@ fn compile_simple() {
...
+    let hashed_manifest_path = hash_path(
+        fsx::canonicalize(t.child("Scarb.toml"))
+            .unwrap()
+            .to_str()
+            .unwrap(),
+    );
+
     cache_dir
-        .child("registry/std")
+        .child(format!("{hashed_manifest_path}/registry/std"))
```
maciektr, line 54: Isn't this kind of besides the point? 
I **do want** to share the downloaded registry packages between workspaces.

utils/scarb-test-support/src/filesystem.rs
```diff
@@ -53,3 +55,10 @@ pub fn path_with_temp_dir(t: &TempDir) -> OsString {
     let other_paths = env::split_paths(&os_path);
     env::join_paths(script_path.chain(other_paths)).unwrap()
 }
+
+pub fn hash_path(path: &str) -> String {
+    let mut hasher = DefaultHasher::new();
```
maciektr, line 60: Note, that the DefaultHasher is PRNG-seeded, so to make the hashes consistent between runs you need to use `scarb_stable_hash::StableHasher`.

Issue #1877, opened 2025-01-09
Registry source cache db advisory lock occasionally fails to release 


maciektr:
### Problem

Originally posted in: https://discord.com/channels/793094838509764618/1270121041116794962/1325874218352771102

>  I am running into an issue with scarb where it freezes when I run any command.
> 
> scarb build
> Blocking waiting for file lock on registry db cache
> 
> 
> I have closed all my terminals and restarted a single terminal but the issue persists.
> 
> Once I restart my operating system it works fine after, but then I eventually ran into the error again.
> 
> Can someone suggest how I can fix this issue for good.
> 
> OS: MacOS
> Scarb: 2.9.2
>

mkaput:
This issue also affects Ma'at so I'm adding it to its board.

PR #2467, opened 2025-07-30
fix: multiple parallel builds deadlock

FroyaTheHen:
fixes https://github.com/software-mansion/scarb/issues/1877

Problem: 
two processes - trying to acquire packages from different sources (registry and git) requiring  _registry cache db lock_ and  _package cache_ lock respectively - acquire each one lock. Then one of the processes tries to acquire package from the other source, requiring a lock that is being held by the other process, so it waits synchronously for the lock blocking the executor. Analogical actions of the second proces result in deadlock.   

Solution: 
await locks in separate threads to avoid blocking. 

Reproduce:
I have two simple projects with the same Scarb.toml content:

[package]
name = "brutus"
version = "0.1.0"
edition = "2024_07"

[dependencies]
openzeppelin_merkle_tree = "0.17.0"
alexandria_math = { git = "https://github.com/keep-starknet-strange/alexandria.git" }

[dev-dependencies]
cairo_test = "2.11.4"
run: scarb cache clean
for each project run: scarb build
it has to be done shortly one after another. I didn't need any extra tools on my mac to ensure that timespan is short enough. The second or so to switch terminal window would still reproduce the problem. Although if it took couple of seconds it was too late and both builds finished with success.

### DIFF
--- a/scarb/src/flock.rs
+++ b/scarb/src/flock.rs
@@ -8,9 +8,9 @@ use anyhow::{Context, Result, ensure};
 use camino::{Utf8Path, Utf8PathBuf};
 use fs4::tokio::AsyncFileExt;
 use fs4::{FileExt, lock_contended_error};
-use tokio::sync::Mutex;
-
 use scarb_ui::components::Status;
+use std::mem;
+use tokio::sync::Mutex;
 
 use crate::core::Config;
 use crate::internal::fsx;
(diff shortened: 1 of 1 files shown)


scarb/src/flock.rs
```diff
@@ -163,23 +163,17 @@ impl AdvisoryLock<'_> {
     /// An RAII structure is returned to release the lock, and if this process abnormally
     /// terminates the lock is also released.
     pub async fn acquire_async(&self) -> Result<AdvisoryLockGuard> {
-        let mut slot = self.file_lock.lock().await;
-
-        let file_lock_arc = match slot.upgrade() {
```
mkaput, line 168: reusing in-process locks should be preserved

maciektr (approved):

PR #2470, opened 2025-07-31
Parallelize calculating checksums for fingerprints

maciektr:
- **Parallelize calculating local checksums on fingerprint creation**
- **Add tracing span**
- **Parallelize calculating plugin checksums**
- **Warmup fingerprint digests**

### DIFF
--- a/scarb/src/compiler/incremental/fingerprint.rs
+++ b/scarb/src/compiler/incremental/fingerprint.rs
@@ -16,11 +16,11 @@ use futures::{StreamExt, stream};
 use itertools::Itertools;
 use scarb_stable_hash::{StableHasher, u64_hash};
 use smol_str::SmolStr;
-use std::cell::{OnceCell, RefCell};
 use std::collections::{HashMap, HashSet};
 use std::hash::{Hash, Hasher};
 use std::ops::Deref;
-use std::rc::{Rc, Weak};
+use std::sync::{Arc, OnceLock, RwLock, Weak};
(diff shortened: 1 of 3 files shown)


maciektr:
Before: 
image

After: 
image

maciektr:

- #2470 ◀
    - #2471

scarb/src/compiler/incremental/fingerprint.rs
```diff
@@ -108,27 +110,44 @@ pub enum ComponentFingerprint {
...
+                ),
+                ToFingerprint::Plugin(plugin) => (
+                    plugin.component_dependency_id.clone(),
+                    PluginFingerprint::try_from_plugin(plugin, unit, ws)
+                        .await
+                        .map(ComponentFingerprint::Plugin),
+                ),
+            }
+        })
+        .buffer_unordered(usize::MAX);
```
FroyaTheHen, line 147: nit pick: why not use some smaller value like in `collect_packages_from_resolve_graph` just maybe a pre-defined arbitrary constant?

maciektr: I don't think we need to add backpressure here. When downloading packages, we add a limit, as too much concurrent downloads would just exhaust the available internet connection throughput without any benefit. Here the task spawns some heavier piece of work and awaits it's completion - we want to use as much of the available thread pool processing power as possible, so adding a backpressure would only stop us from using all of it. Let me know if you think otherwise 🤔

Issue #704, opened 2023-09-25
Validate file stems on custom source path targets with virtual lib.cairo wrapper file 


maciektr:
### Problem

Validate file stems on custom source path targets with virtual lib.cairo wrapper file

mkaput:
BTW I wish we has `#[path]` on `mod`, like Rust has

PR #2487, opened 2025-08-04
validate file steams custom source path target

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/704

### DIFF
--- a/scarb/src/core/manifest/target.rs
+++ b/scarb/src/core/manifest/target.rs
@@ -1,14 +1,13 @@
-use std::hash::{Hash, Hasher};
-use std::ops::Deref;
-use std::sync::Arc;
-
-use anyhow::Result;
+use crate::core::{TargetKind, TomlExternalTargetParams};
+use crate::internal::restricted_names;
+use crate::internal::serdex::toml_merge;
+use anyhow::{Result, bail};
 use camino::{Utf8Path, Utf8PathBuf};
 use serde::{Deserialize, Serialize};
 use smol_str::SmolStr;
-
-use crate::core::{TargetKind, TomlExternalTargetParams};
-use crate::internal::serdex::toml_merge;
+use std::hash::{Hash, Hasher};
+use std::ops::Deref;
+use std::sync::Arc;
 
 /// See [`TargetInner`] for public fields reference.
 #[derive(Clone, Debug, Hash)]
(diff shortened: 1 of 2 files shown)


scarb/src/core/manifest/target.rs
```diff
@@ -67,14 +66,79 @@ impl Target {
...
+        }
+
+        if restricted_names::is_keyword(file_stem) {
+            bail!(
+                "the name `{file_stem}` cannot be used as a test target name, \
+                names cannot use Cairo keywords see the full list at https://starknet.io/cairo-book/appendix-01-keywords.html \
+                Consider renaming file: {}",
+                source_path
+            )
+        }
```
maciektr: Is there a reason why we would need to disallow using keywords here? 🤔 
Initially, I only thought about catching stuff that breaks the compiler in a strange way - e.g. If I call it `ąę.cairo`, I get: 
```
        3 + error: Missing token TerminalIdentifier.
        4 +  --> /private/var/folders/2t/7dt08_ss585ct08q57mvw_q40000gn/T/.tmpsDeZIR/tests/lib.cairo:1:4
        5 + mod ąę;
        6 +    ^
        7 + 
        8 + error: Expected either ';' or '{' after module name. Use ';' for an external module declaration or '{' for a module with a body.
        9 +  --> /private/var/folders/2t/7dt08_ss585ct08q57mvw_q40000gn/T/.tmpsDeZIR/tests/lib.cairo:1:4
       10 + mod ąę;
       11 +    ^
       12 + 
       13 + error: Skipped tokens. Expected: Const/Enum/ExternFunction/ExternType/Function/Impl/InlineMacro/Module/Struct/Trait/TypeAlias/Use or an attribute.
       14 +  --> /private/var/folders/2t/7dt08_ss585ct08q57mvw_q40000gn/T/.tmpsDeZIR/tests/lib.cairo:1:5
       15 + mod ąę;
       16 +     ^^^
       17 + 
       18 + error[E0005]: Module file not found. Expected path: /private/var/folders/2t/7dt08_ss585ct08q57mvw_q40000gn/T/.tmpsDeZIR/tests/.cairo
       19 +  --> /private/var/folders/2t/7dt08_ss585ct08q57mvw_q40000gn/T/.tmpsDeZIR/tests/lib.cairo:1:1
       20 + mod ąę;
       21 + ^^^
       22 + 
       23 + error: could not compile `hello_integrationtest` due to previous error
```
As people may not realize, it becomes a module name.

FroyaTheHen: I did it because some of the keywords break with: 
`error: 'trait' is a reserved identifier `
I would still validate those here because as you wrote 

> people may not realize, it becomes a module name

imo, it is not much to add one extra check to the custom error handling here and improvement of the error message is worth it. Although i did check with the `fn is_keyword ` instead of checking for _reserved identifier_ - that's just my bad. Any thoughts on that?

maciektr: Oh, I didn't realize that, that's good point! Let's do the validation then!

maciektr: Not sure what do you mean by:

> Although i did check with the fn is_keyword instead of checking for reserved identifier - that's just my bad. Any thoughts on that?

What's the alternative?

scarb/src/core/manifest/target.rs
```diff
@@ -67,14 +66,79 @@ impl Target {
     pub fn try_from_structured_params(
         kind: TargetKind,
         name: impl Into<SmolStr>,
-        source_path: impl Into<Utf8PathBuf>,
+        source_path: impl Into<Utf8PathBuf> + Clone,
         group_id: Option<SmolStr>,
         params: impl Serialize,
     ) -> Result<Self> {
+        if kind == TargetKind::TEST {
```
maciektr: I guessed we don't need this check - but for anything other than Test, the stem will always be the default one.

scarb/src/core/manifest/target.rs
```diff
@@ -67,14 +66,79 @@ impl Target {
...
+            Self::validate_test_target_file_stem(source_path.clone().into())?;
+        }
         let params = toml::Value::try_from(params)?;
         Ok(Self::new(kind, name, source_path, group_id, params))
     }
 
+    pub fn validate_test_target_file_stem(source_path: Utf8PathBuf) -> Result<()> {
+        let file_stem = source_path.file_stem().unwrap();
```
maciektr: Nit: Please replace unwrap with expect 🙏🏼

scarb/src/core/manifest/target.rs
```diff
@@ -67,14 +66,79 @@ impl Target {
...
+
+        if file_stem == ".cairo" {
+            bail!(
+                "empty string cannot be used as a test target name. \
+                Consider renaming file: {}",
```
maciektr: Nit: we usually do all lowercase without a dot at the end instead of full sentences in error msgs

scarb/src/core/manifest/target.rs
FroyaTheHen: > Not sure what do you mean by:
> 
> Although i did check with the fn is_keyword instead of checking for reserved identifier - that's just my bad. Any thoughts on that?

I mean that i used `fn is_keyword` but i didn't (at first) check if the compilation indeed fails for all of the keywords that the function checks. It does not. For 16 of the keywords it succeeds. So the alternative would be to write a very similar function that checks only the words that would fail as a mod name. Like: 
```
pub fn is_strict_keyword(name: &str) -> bool {
    [
    "as",
    "break",
    "const",
    "continue",
    "else",
    "enum",
    "extern",
    "false",
    "fn",
    "for", // ??
    "if",
    "implicits",
    "impl",
    "let",
    "loop",
    "macro", // ??
    "match",
    "mod",
    "mut",
    "nopanic",
    "of",
    "pub",
    "ref",
    "return",
    "struct",
    "trait",
    "true",
    "type",
    "use",
    "while",
].contains(&name)
} 
```
which could point to [Strict kewords](https://starknet.io/cairo-book/appendix-01-keywords.html#strict-keywords). 
(Note that `for` and `macro` are not documented - in the provided link section.)

I think it would be better to ban all keyowrds (like currently proposed in the PR) no matter if the compilation fails or not. I can hardly picture a situation when a user needs to stick with the name. 

The oppose argument in favour of creating `fn is_strict_keyword` would be that the  [is_keyword_terminal](https://github.com/starkware-libs/cairo/blob/e340b686af0fe3d06e24498f7e9151a0f2597e1f/crates/cairo-lang-syntax/src/node/kind.rs#L577) list is not often updated. 
 ____
My finally thoughts: this ☝️ is overthinking. I would ban all keywords as currently proposed in this PR.

maciektr: > I would ban all keywords as currently proposed in this PR.

Yeah, I agree.

PR #2488, opened 2025-08-04
update link in package name validation error message

FroyaTheHen:


scarb/src/core/package/name.rs
```diff
@@ -103,8 +103,7 @@ impl PackageName {
         if restricted_names::is_keyword(name.as_str()) {
             bail!(
                 "the name `{name}` cannot be used as a package name, \
-                names cannot use Cairo keywords see the full list at \
-                https://docs.cairo-lang.org/language_constructs/keywords.html"
+                names cannot use Cairo keywords see the full list at https://www.starknet.io/cairo-book/appendix-01-keywords.html"
```
DelevoXDG: Nit: 
is `www` necessary? 🤔

DelevoXDG (approved):

scarb/src/core/package/name.rs
FroyaTheHen: right, no 👍

maciektr (approved):

PR #2493, opened 2025-08-05
Do not wait for dangling requests when resolution completes

maciektr:

PR #2494, opened 2025-08-05
Do not query registry to build resolve graph

maciektr:
This does not really make sense on it's own, as we need to query the summaries in registry anyway, but will allow us to avoid doing so in some cases in future.


maciektr:

- #2493
    - #2494 ◀

PR #2506, opened 2025-08-07
Refactor file lock guards

mkaput:
1. Rust 1.89.0 brought all features that we used from fs4 into std 🎉
2. It turns out we don't have to explicitly unlock files because they are going to already be unlocked then the file is closed - which happens when we drop guards.
3. Renamed `(Async)FileLockGuard` to `(Async)LockedFile` because it's better naming now

### DIFF
--- a/scarb/src/ops/lockfile.rs
+++ b/scarb/src/ops/lockfile.rs
@@ -16,7 +16,7 @@ pub fn read_lockfile(ws: &Workspace<'_>) -> Result<Lockfile> {
         .open(ws.lockfile_path())
         .context("failed to open lockfile")?;
 
-    file.lock()
+    file.lock_shared()
         .context("failed to acquire shared lockfile access")?;
 
     let mut content = String::new();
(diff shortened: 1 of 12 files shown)


mkaput:
tests finally pass 🎉

scarb/src/ops/lockfile.rs
```diff
@@ -17,7 +16,8 @@ pub fn read_lockfile(ws: &Workspace<'_>) -> Result<Lockfile> {
         .open(ws.lockfile_path())
         .context("failed to open lockfile")?;
 
-    FileExt::lock_shared(&file).context("failed to acquire shared lockfile access")?;
+    file.lock()
```
maciektr: Why go from shared to exclusive?
Can you update the context msg?

mkaput: wow, what a mistake 🤦🏻 

```suggestion
    file.lock_shared()
```

PR #2542, opened 2025-08-14
add common trait for db to improve maintainability

FroyaTheHen:
I have mixed feelings for this PR but still think it will be better to maintain than the repetition. Looking forward to hear you opinion 🙏


FroyaTheHen:
> A lot of salsa stuff is being completely rewritten in Cairo repo. I suggest postponing such refactorings until we will settle down here

closing because of the above

PR #2546, opened 2025-08-14
Make proc macros expand first

piotmag769:
This fixes

```cairo
#[derive(Drop)]
#[type_hash(name: "My Struct")]
pub struct MyStruct { }
```

### DIFF
--- a/scarb/src/compiler/plugin/collection.rs
+++ b/scarb/src/compiler/plugin/collection.rs
@@ -159,7 +159,6 @@ struct PluginSuiteAssembler {
     base: PluginSuite,
     builtin: PluginSuite,
     proc_macro: PluginSuite,
-    post_proc_macros: PluginSuite,
 }
 
 impl Default for PluginSuiteAssembler {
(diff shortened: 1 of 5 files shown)


scarb/src/compiler/plugin/collection.rs
```diff
@@ -168,31 +167,24 @@ impl Default for PluginSuiteAssembler {
...
         Ok(())
     }
     pub fn add_proc_macro(&mut self, suite: PluginSuite) {
         self.proc_macro.add(suite);
     }
     pub fn assemble(self) -> PluginSuite {
-        let mut suite = self.base;
+        let mut suite = self.proc_macro;
+        suite.add(self.base);
         suite.add(self.builtin);
```
piotmag769, line 187: This doesn't solve all the possible issues, and never will - even proc macros themselves may fall into generating double impls for derives pretty easily. The only option to solve most of the problems is to make sure that we always remove an appropriate derive after expanding it. The problems that won't be solved - compiler macros with similar cases.

scarb/src/compiler/plugin/collection.rs
```diff
@@ -168,31 +167,24 @@ impl Default for PluginSuiteAssembler {
...
 impl PluginSuiteAssembler {
     pub fn add_builtin(&mut self, plugin: &dyn CairoPlugin) -> Result<()> {
-        let post_proc_macros = plugin.post_proc_macros();
         let instance = plugin.instantiate()?;
         let suite = instance.plugin_suite();
-        if post_proc_macros {
-            self.post_proc_macros.add(suite);
-        } else {
-            self.builtin.add(suite);
-        }
```
Arcticae, line 185: I understand the change of order - but why was this piece of code removed? What was `post_proc_macro`s usecase?

mkaput: introduced in https://github.com/software-mansion/scarb/pull/2186

that PR implemented this PR's behaviour, but only partially, relative to `ExecutablePlugin`

PR #2558, opened 2025-08-19
Make token streams for derive macros relative

piotmag769:


scarb/src/compiler/plugin/proc_macro/v2/host/derive/mod.rs
```diff
@@ -99,23 +102,25 @@ impl ProcMacroHostPlugin {
...
                 .instance(derive.package_id)
                 .try_v2()
                 .expect("procedural macro using v1 api used in a context expecting v2 api")
                 .generate_code(
                     derive.expansion.expansion_name.clone(),
-                    call_site.span.clone(),
+                    call_site.clone(),
                     TokenStream::empty(),
-                    token_stream,
+                    adapted_token_stream.clone(),
```
piotmag769, line 119: Not sure why it wasn't cloned before (edit: this is better for case with just one derive since we don't need to clone 🤦). Thought maybe it was bcs of the bump allocator, but all the tokens just keep `Rc<BumpWrap>` and `*const str` so I believe cloning is safe cc @mkaput

mkaput: because token stream was being built within for loop previously

piotmag769: I believe this is faster anyways - token stream building traverses syntax tree and does some span related logic. This just clones a bunch of pointers and spans.

Draggu (approved):

PR #2560, opened 2025-08-21
Add `require-audits` to Scarb.toml

DelevoXDG:
Add ability to enforce audit requirement:
```toml
[security]
require-audits = true
```
If enabled, error if unable to resolve dep tree with audited packages only

## Stack

- #2560
- #2561
- #2617
- #2563
- #2562 
- #2566
- #2575
- #2576

### DIFF
--- a/scarb/src/core/config.rs
+++ b/scarb/src/core/config.rs
@@ -323,7 +323,6 @@ pub struct ConfigBuilder {
     profile: Option<Profile>,
     load_proc_macros: bool,
     load_prebuilt_proc_macros: bool,
-    require_audits: bool,
 }
 
 impl ConfigBuilder {
@@ -345,7 +344,6 @@ impl ConfigBuilder {
             profile: None,
             load_proc_macros: true,
             load_prebuilt_proc_macros: true,
-            require_audits: false,
         }
     }
 
@@ -437,11 +435,6 @@ impl ConfigBuilder {
 
     pub fn load_prebuilt_proc_macros(mut self, load_prebuilt_proc_macros: bool) -> Self {
         self.load_prebuilt_proc_macros = load_prebuilt_proc_macros;
-        self
-    }
-
-    pub fn require_audits(mut self, require_audits: bool) -> Self {
-        self.require_audits = require_audits;
         self
     }
 }
--- a/scarb/src/core/config.rs
+++ b/scarb/src/core/config.rs
@@ -323,6 +323,7 @@ pub struct ConfigBuilder {
     profile: Option<Profile>,
     load_proc_macros: bool,
     load_prebuilt_proc_macros: bool,
+    require_audits: bool,
 }
 
 impl ConfigBuilder {
@@ -344,6 +345,7 @@ impl ConfigBuilder {
             profile: None,
             load_proc_macros: true,
             load_prebuilt_proc_macros: true,
+            require_audits: false,
         }
     }
 
@@ -435,6 +437,11 @@ impl ConfigBuilder {
 
     pub fn load_prebuilt_proc_macros(mut self, load_prebuilt_proc_macros: bool) -> Self {
         self.load_prebuilt_proc_macros = load_prebuilt_proc_macros;
+        self
+    }
+
+    pub fn require_audits(mut self, require_audits: bool) -> Self {
+        self.require_audits = require_audits;
         self
     }
 }
--- a/scarb/src/core/manifest/summary.rs
+++ b/scarb/src/core/manifest/summary.rs
@@ -30,6 +30,8 @@ pub struct SummaryInner {
     pub no_core: bool,
     #[builder(default)]
     pub checksum: Option<Checksum>,
+    #[builder(default = false)]
+    pub require_audits: bool,
 }
 
 impl Deref for Summary {
--- a/scarb/src/core/manifest/summary.rs
+++ b/scarb/src/core/manifest/summary.rs
@@ -30,8 +30,6 @@ pub struct SummaryInner {
     pub no_core: bool,
     #[builder(default)]
     pub checksum: Option<Checksum>,
-    #[builder(default = false)]
-    pub require_audits: bool,
 }
 
 impl Deref for Summary {
--- a/scarb/src/core/manifest/toml_manifest.rs
+++ b/scarb/src/core/manifest/toml_manifest.rs
@@ -47,7 +47,6 @@ pub struct TomlManifest {
     pub workspace: Option<TomlWorkspace>,
     pub dependencies: Option<BTreeMap<PackageName, MaybeWorkspaceTomlDependency>>,
     pub dev_dependencies: Option<BTreeMap<PackageName, MaybeWorkspaceTomlDependency>>,
-    pub security: Option<MaybeWorkspaceTomlSecurity>,
     pub lib: Option<TomlTarget<TomlLibTargetParams>>,
     pub executable: Option<TomlTarget<TomlExecutableTargetParams>>,
     pub cairo_plugin: Option<TomlTarget<TomlCairoPluginTargetParams>>,
@@ -103,10 +102,10 @@ type TomlToolsDefinition = BTreeMap<SmolStr, toml::Value>;
 #[serde(rename_all = "kebab-case")]
 pub struct TomlWorkspace {
     pub members: Option<Vec<String>>,
+    pub require_audits: Option<bool>,
     pub package: Option<PackageInheritableFields>,
     pub dependencies: Option<BTreeMap<PackageName, TomlDependency>>,
     pub dev_dependencies: Option<BTreeMap<PackageName, TomlDependency>>,
-    pub security: Option<TomlSecurity>,
     pub scripts: Option<BTreeMap<SmolStr, ScriptDefinition>>,
     pub tool: Option<TomlToolsDefinition>,
 }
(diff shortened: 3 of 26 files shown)


utils/scarb-test-support/src/registry/local.rs
```diff
@@ -53,23 +60,38 @@ impl fmt::Display for LocalRegistry {
     }
 }
 
-#[derive(Serialize, Deserialize)]
-struct Package {
-    pub v: String,
-    pub deps: Vec<String>,
-    pub cksum: String,
-    pub yanked: Option<bool>,
```
DelevoXDG, line 61: Note: 

This was buggy as this struct and its handling was not compliant with the format used by `IndexRecord`, including
- invalid `deps` field
- invalid handling of boolean fields (`yanked` and now `audited`), incorrectly writing `null` value to json, which is not allowed

Thus, the most obvious course of action was to remove `Package` struct and just directly use `IndexRecord` instead

utils/scarb-test-support/src/registry/local.rs
```diff
@@ -1,11 +1,18 @@
 use crate::command::Scarb;
-use anyhow::Result;
 use assert_fs::TempDir;
+use std::fmt;
+use url::Url;
+
+#[cfg(feature = "scarb-config")]
```
FroyaTheHen: nit pick: 
Maybe delete all `#[cfg(feature = "scarb-config")]` occurrences here and replace `utils/scarb-test-support/src/registry/mod.rs` content with 
```
#[cfg(feature = "scarb-config")]
pub mod http;
#[cfg(feature = "scarb-config")]
pub mod local;
```

website/docs/reference/specifying-dependencies.md
```diff
@@ -31,6 +31,24 @@ To use a custom, non-official registry you need to specify its url address as we
...
+
+You can enable this feature by specifying the `require-audits` key in the `[security]` section:
+
+```toml
+[security]
+require-audits = true
+```
+
+Setting this field to `true` will cause Scarb to ignore any versions of dependencies, including transitive ones, that are not marked as audited.
+If unable to resolve dependency tree due to this, Scarb will exit with an error.
```
FroyaTheHen:
```suggestion
Setting this field to `true` will cause Scarb to ignore any not audited dependency version, including transitive ones.
If unable to resolve the dependency tree, Scarb will exit with an error.
```
for me the suggestion sounds a bit more concise but up to you 🫡

DelevoXDG: I've updated it to:
```suggestion
Setting this field to `true` will cause Scarb to ignore any versions of dependencies, including transitive ones, that are not marked as audited in the registry.
If unable to resolve the dependency tree due to this, Scarb will exit with an error.
By default, this field is set to `false`.
```
While I understand the need to be concise, I feel we should be rather explicit with regard to this feature, including specifying clearly that the source of whether package is considered audited _is_ the registry - not any external service, and not a judgement made by Scarb, which is not something conveyed by `not audited`.
As for other things, stuff like "due to this" IMO allows the text to flow more freely, but I don't think matters much tbh 😄 

WDYT?

utils/scarb-test-support/src/registry/local.rs
DelevoXDG: Good idea, this will quiet simplify this file 😄

website/docs/reference/specifying-dependencies.md
FroyaTheHen: That's ok. Your explanation makes a lot of sense :) After all, I agree that it doesn't matter all that much and is all very subjective 👍

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -47,6 +47,7 @@ pub struct TomlManifest {
     pub workspace: Option<TomlWorkspace>,
     pub dependencies: Option<BTreeMap<PackageName, MaybeWorkspaceTomlDependency>>,
     pub dev_dependencies: Option<BTreeMap<PackageName, MaybeWorkspaceTomlDependency>>,
+    pub security: Option<MaybeWorkspaceTomlSecurity>,
```
maciektr: It's not obvious for me, why does it need to be a separate section? 🤔

DelevoXDG: Other sections didn't seem like a really good fit, aside from maybe `[tool.scarb]`, since it also hosts `allow-prebuilt-plugins`, which is similar in some ways.

If I understand correctly, it currently only allows inheriting tool config in its entirety
```toml
[tool]
scarb.workspace = true
```
but I should be able to tweak that. Aside from that, I don't have specific reasons *not to use it*, aside from minor reservations about whether the audit requirement is certainly something we consider "Scarb specific configuration," and that using this section for audit settings seemed slightly less user-friendly compared to `[security]`, which is more intuitive.

Where do you think it would be best placed?

scarb/src/core/config.rs
maciektr, line 1: Is this used anywhere? Cannot find any usages 🤔

scarb/src/core/workspace.rs
maciektr, line 1: This logic seems confusing for me. 

If you do:
```Scarb.toml
[package]
name = "first"

[security]
require-audits = true
```
It requires audits - as it falls into [this branch](https://github.com/software-mansion/scarb/blob/8e10fb5538f7a74355f729ca64e9138709ffd9f3/scarb/src/ops/workspace.rs#L160-L164) which uses `from_single_package` that relies on the package Summary.

However, if you then add a single line:

```Scarb.toml
[workspace]

[package]
name = "first"

[security]
require-audits = true
```
It does not require audits anymore, as it now falls into the other branch, that only checks [`[workspace.security]`](https://github.com/software-mansion/scarb/blob/8e10fb5538f7a74355f729ca64e9138709ffd9f3/scarb/src/ops/workspace.rs#L145-L149)

scarb/src/core/manifest/toml_manifest.rs
maciektr: Since: 
- this is basically a resolver input and resolver works workspace wide
- It cannot really be configured on per package bases independently
- Problems mentioned in https://github.com/software-mansion/scarb/pull/2560/files#r2307708266

I suggest we just make it a flag under `[workspace]` section - akin to resolver version in Cargo (https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions).
WDYT? 🤔

scarb/src/core/config.rs
DelevoXDG: removed

scarb/src/core/workspace.rs
DelevoXDG: Good catch!

I have now changed the logic to what you have suggested in the other comment - to locate `require-audits` under `[workspace]` section, which also fixes the problem you described.

Additionally, unlike Cargo's `resolver` option, I feel like this setting deserves a bit more clarity from the user regarding the policy used in the workspace, so instead of ignoring conflicts between the workspace root and member setting, I've added an error in such context.

website/docs/reference/workspaces.md
```diff
@@ -175,6 +175,26 @@ openzeppelin.workspace = true
...
+```toml
+[workspace]
+require-audits = true
+```
+
+Setting this field to `true` will cause Scarb to ignore any versions of dependencies, including transitive ones, that are not marked as audited in the registry.
+If unable to resolve the dependency tree due to this, Scarb will exit with an error.
+By default, this field is set to `false`.
+This policy applies applies to entire workspace.
+Members must not conflict with the root setting.
```
DelevoXDG: Do you think we should keep it in both places, or just keep one version and redirect to the other? 🤔

FroyaTheHen: I opt for the redirect. That way we do not risk updating only one page in case of any changes.

scarb/src/ops/workspace.rs
```diff
@@ -114,6 +113,15 @@ fn read_workspace_root<'c>(
             .map(AsRef::as_ref)
             .map(|package_path| {
                 let package_manifest = TomlManifest::read_from_path(package_path)?;
+                if package_path != manifest_path
+                    && let Some(member_ws) = package_manifest.get_workspace()
+                    && member_ws.require_audits.is_some()
+                {
+                    return Err(anyhow!(
+                        "only the workspace root may set `[workspace].require-audits`"
```
maciektr: I don't understand this check. 

Scarb run is always contained to a single workspace. So we should never read keys from other workspaces. 

We may (not sure if we should) decide to disallow workspaces intertwining like this altogether - but disallowing using some keys in one of them is pointless. 

Effectively, all you achieve here, is to disallow using the new setting in some workspaces based on values set in other workspaces, which should never happen.

maciektr (changes requested):

scarb/src/core/manifest/summary.rs
```diff
@@ -30,6 +30,8 @@ pub struct SummaryInner {
     pub no_core: bool,
     #[builder(default)]
     pub checksum: Option<Checksum>,
+    #[builder(default = false)]
+    pub require_audits: bool,
```
maciektr: We shouldn't need this now, as the value is set on Workspace level, not package metadata level - it makes more sense to store it in ws.

scarb/src/core/workspace.rs
```diff
@@ -77,7 +81,7 @@ impl<'c> Workspace<'c> {
     ) -> Result<Self> {
         let manifest_path = package.manifest_path().to_path_buf();
         let root_package = Some(package.id);
-
+        let require_audits = package.manifest.summary.require_audits;
```
maciektr: vide supra - shouldn't need to come from the package metadata

scarb/src/sources/registry.rs
```diff
@@ -99,6 +102,9 @@ impl Source for RegistrySource<'_> {
             if record.yanked && !self.yanked_whitelist.contains(&package_id) {
                 return None;
             }
+            if self.require_audits && dependency.kind == DepKind::Normal && !record.audited {
```
maciektr: Note, that currently the resolver deduplicates registry queries ignoring the dep kind - https://github.com/software-mansion/scarb/blob/eeca5cf4fc0e4e918592b63daafa583e8c9eed85/scarb/src/resolver/in_memory_index.rs#L24-L46

maciektr: I'd move this DepKind check logic to the next PR for more clarity around how we want to handle that - because I don't think this works as of now 🤔

scarb/tests/audits.rs
```diff
@@ -0,0 +1,241 @@
...
+        .stdout_matches(indoc! {r#"
+        error: cannot get dependencies of `hello_world@1.0.0`
+
+        Caused by:
+            cannot find package `foo ^1.0.0`
+        "#});
+
+    audit(registry.t.child("index/3/f/foo.json").path(), "1.0.0").unwrap();
+    Scarb::quick_snapbox()
+        .arg("build")
```
maciektr: Can this be `fetch` instead?

scarb/tests/audits.rs
```diff
@@ -0,0 +1,241 @@
...
+        .stdout_matches(indoc! {r#"
+            error: cannot get dependencies of `bar@1.0.0`
+
+            Caused by:
+                cannot find package `foo ^1.0.0`
+        "#});
+}
+
+#[test]
+fn require_audits_allows_audited_version() {
```
maciektr: Doesn't this duplicate `require_audits_disallows_non_audited_version`?

website/docs/reference/workspaces.md
DelevoXDG: After some further consideration, I think it'd just be more readable if we don't redirect. Let's change it in the future if there are problems 😄

PR #2561, opened 2025-08-21
Bypass audit requirement for dev deps

DelevoXDG:
The flag added in @2560 is automatically ignored for dev deps

## Stack

- #2560
- #2561
- #2617
- #2563
- #2562 
- #2566
- #2575
- #2576

### DIFF
--- a/scarb/src/core/registry/patch_map.rs
+++ b/scarb/src/core/registry/patch_map.rs
@@ -17,13 +17,26 @@ impl PatchMap {
 
     /// Lookup the `dependency` in this patch map and return patched dependency if found,
     /// or return `dependency` back otherwise.
-    pub fn lookup<'a>(&'a self, dependency: &'a ManifestDependency) -> &'a ManifestDependency {
+    pub fn lookup(&self, dependency: &ManifestDependency) -> ManifestDependency {
         let source_pattern = &dependency.source_id.canonical_url;
-        let result = self
+        let patch_dep = self
             .map
             .get(source_pattern)
-            .and_then(|patches| patches.get(&dependency.name))
-            .unwrap_or(dependency);
+            .and_then(|patches| patches.get(&dependency.name));
+
+        let result = if let Some(patch_dep) = patch_dep {
+            ManifestDependency::builder()
+                .name(patch_dep.name.clone())
+                .version_req(patch_dep.version_req.clone())
+                .source_id(patch_dep.source_id)
+                .kind(dependency.kind.clone())
+                .features(patch_dep.features.clone())
+                .default_features(patch_dep.default_features)
+                .build()
+        } else {
+            dependency.clone()
+        };
+
         self.unused.borrow_mut().remove(&(
             source_pattern.clone(),
             result.name.clone(),
--- a/scarb/src/resolver/provider.rs
+++ b/scarb/src/resolver/provider.rs
@@ -2,7 +2,7 @@ use crate::core::lockfile::Lockfile;
 use crate::core::registry::patch_map::PatchMap;
 use crate::core::{
     DepKind, DependencyFilter, DependencyVersionReq, ManifestDependency, PackageId, PackageName,
-    SourceId, Summary,
+    SourceId, SourceKind, Summary,
 };
 use crate::resolver::in_memory_index::VersionsResponse;
 use crate::resolver::{Request, ResolverState};
@@ -125,6 +125,7 @@ pub struct PubGrubDependencyProvider {
     lockfile: Lockfile,
     state: Arc<ResolverState>,
     request_sink: mpsc::Sender<Request>,
+    require_audits: bool,
 }
 
 impl PubGrubDependencyProvider {
@@ -134,6 +135,7 @@ impl PubGrubDependencyProvider {
         request_sink: mpsc::Sender<Request>,
         patch_map: PatchMap,
         lockfile: Lockfile,
+        require_audits: bool,
     ) -> Self {
         Self {
             main_package_ids,
@@ -144,6 +146,7 @@ impl PubGrubDependencyProvider {
             patch_map,
             lockfile,
             request_sink,
+            require_audits,
         }
     }
 
@@ -326,11 +329,17 @@ impl DependencyProvider for PubGrubDependencyProvider {
             .get(package)
             .cloned()
             .unwrap_or_default();
-        let dependency = package.to_dependency(range.clone(), kind);
+        let dependency = package.to_dependency(range.clone(), kind.clone());
         let summaries = self.wait_for_summaries(dependency)?;
         let summaries = summaries
             .into_iter()
             .filter(|summary| range.contains(&summary.package_id.version))
+            .filter(|summary| {
+                !self.require_audits
+                    || summary.package_id.source_id.kind != SourceKind::Registry
+                    || kind.is_test()
+                    || summary.audited
+            })
             .sorted_by_key(|summary| summary.package_id.version.clone())
             .collect_vec();
 
--- a/scarb/src/resolver/provider.rs
+++ b/scarb/src/resolver/provider.rs
@@ -1,7 +1,7 @@
 use crate::core::lockfile::Lockfile;
 use crate::core::registry::patch_map::PatchMap;
 use crate::core::{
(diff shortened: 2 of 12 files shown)


scarb/src/resolver/provider.rs
maciektr, line 1: Since this is basically driven by the resolver state (saved kinds), requires a lot of info to be passed down and causes problems with requests deduplication, I wonder if doing the filtering on registry client level was a right call. Maybe we should just save audit status as Summary field and filter the summaries in `choose_version` directly? Sounds even more straightforward to me 🤔  WDYT?

maciektr (changes requested):

scarb/src/resolver/provider.rs
DelevoXDG: Yes, after looking at it, the problem you describe may indeed arise. 

I moved the logic related to filtering out audited versions to `choose_version`, as you suggested.

To be consistent, I’ve also moved the yanked version filtering there (#2617), so we get the same error messages for yanked and non-audited versions.

I think the error format provided by the resolver is better since it demonstrates the full dependency path that lead to an error 🤔 

As we discussed offline, this is non-deterministic, hence I'm not sure if it would be possible for us to test this

maciektr: > As we discussed offline, this is non-deterministic, hence I'm not sure if it would be possible for us to test this

I don't think we need to tbh

scarb/src/core/registry/patch_map.rs
```diff
@@ -17,13 +17,26 @@ impl PatchMap {
...
         let source_pattern = &dependency.source_id.canonical_url;
-        let result = self
+        let patch_dep = self
             .map
             .get(source_pattern)
-            .and_then(|patches| patches.get(&dependency.name))
-            .unwrap_or(dependency);
+            .and_then(|patches| patches.get(&dependency.name));
+
+        let result = if let Some(patch_dep) = patch_dep {
```
maciektr, line 27: The point of this change is to make it work on patches? Or is there other reason? 🤔
If so, should we have a test case for that?

scarb/src/resolver/provider.rs
```diff
@@ -249,7 +285,14 @@ impl DependencyProvider for PubGrubDependencyProvider {
...
     fn prioritize(&self, package: &Self::P, range: &Self::VS) -> Self::Priority {
-        let dependency: ManifestDependency = package.to_dependency(range.clone());
+        let kind = self
+            .kinds
+            .read()
+            .unwrap()
+            .get(package)
+            .cloned()
+            .unwrap_or_default();
+        let dependency = package.to_dependency(range.clone(), kind);
```
maciektr: Do we need to pass the kind here?

scarb/src/resolver/provider.rs
```diff
@@ -279,11 +322,24 @@ impl DependencyProvider for PubGrubDependencyProvider {
...
         // Query available versions.
-        let dependency: ManifestDependency = package.to_dependency(range.clone());
+        let kind = self
+            .kinds
+            .read()
+            .unwrap()
+            .get(package)
+            .cloned()
+            .unwrap_or_default();
+        let dependency = package.to_dependency(range.clone(), kind.clone());
```
maciektr: vide supra

scarb/src/resolver/provider.rs
```diff
@@ -279,11 +322,24 @@ impl DependencyProvider for PubGrubDependencyProvider {
...
         let summaries = self.wait_for_summaries(dependency)?;
         let summaries = summaries
             .into_iter()
             .filter(|summary| range.contains(&summary.package_id.version))
+            .filter(|summary| {
+                !self.require_audits
+                    || summary.package_id.source_id.kind != SourceKind::Registry
```
maciektr, line 319: So this just allows packages from git and local when audits are required? 🤔 Why?

scarb/src/resolver/provider.rs
```diff
@@ -160,10 +166,22 @@ impl PubGrubDependencyProvider {
     pub fn fetch_summary(&self, package_id: PackageId) -> Result<Summary, DependencyProviderError> {
         let summary = self.packages.read().unwrap().get(&package_id).cloned();
         let summary = summary.map(Ok).unwrap_or_else(|| {
+            let package = PubGrubPackage {
```
maciektr: Please use From trait instead 
```
impl From<PackageId> for PubGrubPackage {
```

scarb/src/resolver/provider.rs
```diff
@@ -198,17 +216,35 @@ impl PubGrubDependencyProvider {
...
+                rewrite_path_dependency_source_id(summary.package_id, &original_dependency);
             let dependency = lock_dependency(&self.lockfile, dependency)?;
+            self.save_kind(&dependency);
             blocking_send(dependency);
         }
         Ok(())
     }
 
+    fn save_kind(&self, dep: &ManifestDependency) {
+        let package = PubGrubPackage {
```
maciektr: Please use From trait instead 
```
impl From<&ManifestDependency> for PubGrubPackage {
```

scarb/src/resolver/provider.rs
```diff
@@ -198,17 +216,35 @@ impl PubGrubDependencyProvider {
...
-                rewrite_path_dependency_source_id(summary.package_id, original_dependency);
+                rewrite_path_dependency_source_id(summary.package_id, &original_dependency);
             let dependency = lock_dependency(&self.lockfile, dependency)?;
+            self.save_kind(&dependency);
             blocking_send(dependency);
         }
         Ok(())
     }
 
+    fn save_kind(&self, dep: &ManifestDependency) {
```
maciektr: Please add a comment describing the logic here, as it does not simply store kind. 
I would consider changing the name 🤔

scarb/src/resolver/provider.rs
```diff
@@ -160,10 +166,22 @@ impl PubGrubDependencyProvider {
...
+                .read()
+                .unwrap()
+                .get(&package)
+                .cloned()
+                .unwrap_or_default();
             let dependency = ManifestDependency::builder()
                 .name(package_id.name.clone())
                 .source_id(package_id.source_id)
                 .version_req(DependencyVersionReq::exact(&package_id.version))
+                .kind(kind)
```
maciektr: Do we need it here? 🤔

scarb/tests/audits.rs
maciektr, line 1: Due to the way how the algo works, I would add one test case, when we depend on the same package in the single workspace, once as dev and once as normal. Maybe something like this:

```
fn() {
    let mut registry = LocalRegistry::create();
    registry.publish(|t| {
        ProjectBuilder::start()
            .name("foo")
            .version("1.0.0")
            .lib_cairo(r#"fn f() -> felt252 { 0 }"#)
            .build(t);
    });
    let pkg1 = t.child("first");
    ProjectBuilder::start()
        .name("first")
        .dev_dep("foo", Dep.version("1.0.0").registry(&registry))
        .build(&pkg1);
    let pkg2 = t.child("second");
    ProjectBuilder::start()
        .name("second")
        .dep("foo", Dep.version("1.0.0").registry(&registry))
        .build(&pkg2);
    WorkspaceBuilder::start()
        .add_member("first")
        .add_member("second")
        .require_audits(true)
        .build(&t);

    Scarb::quick_snapbox()
        .arg("fetch")
        .current_dir(&t)
        .assert()
        .failure();
}


```

scarb/src/resolver/provider.rs
maciektr: I guess https://github.com/software-mansion/scarb/pull/2566 resolves this

scarb/src/core/registry/patch_map.rs
DelevoXDG: Yes, that was the intention! I've added a test for this 🫡

scarb/src/resolver/provider.rs
DelevoXDG: No, now that we moved the filtering to `choose_version`, it's not needed. Removed.

scarb/src/resolver/provider.rs
DelevoXDG: Good idea, I've added a comment and changed name to describe it better

maciektr (approved):
👏 👏 👏

PR #2572, opened 2025-08-26
Make cfg plugin run first

piotmag769:


scarb/src/compiler/plugin/collection.rs
```diff
@@ -182,9 +173,14 @@ impl PluginSuiteAssembler {
...
-        let mut suite = self.proc_macro;
-        suite.add(self.base);
-        suite.add(self.builtin);
+        let mut suite = PluginSuite::default();
+        suite
+            // Config plugin must be first, as it removes items from the AST,
+            // and other plugins may add items prior to the removal of the original.
+            .add_plugin::<ConfigPlugin>()
+            .add(self.proc_macro)
+            .add(get_default_plugin_suite())
```
maciektr, line 184: This will add ConfigPlugin twice, right?

piotmag769: It will

maciektr (approved):

Issue #2262, opened 2025-05-20
Documenting macro declarations


maciektr:
Review whether any changes needs to be made after introduction of `ModuleItemId::MacroDeclaration`

maciektr:
Make sure to remove all todos
https://github.com/software-mansion/scarb/blob/08e7a606be623be32a40abea28dcecad5921e11a/extensions/scarb-doc/src/types/module_type.rs#L255-L255

PR #2574, opened 2025-08-26
feat: document macro declarations

FroyaTheHen:
documenting results of expanding macro calls is work in progress

closes https://github.com/software-mansion/scarb/issues/2262


extensions/scarb-doc/src/types/groups.rs
```diff
@@ -20,6 +20,7 @@ pub struct Group<'db> {
     pub impls: Vec<Impl<'db>>,
     pub extern_types: Vec<ExternType<'db>>,
     pub extern_functions: Vec<ExternFunction<'db>>,
+    pub macro_declarations: Vec<MacroDeclaration<'db>>,
```
FroyaTheHen, line 23: macro declaration cannot have a group attribute but the field is added for the sake of macro rule [`define_insert_function`](https://github.com/software-mansion/scarb/blob/036db00740c43fe376a157f4c99a9ab4936ded52/extensions/scarb-doc/src/types/module_type.rs#L171C14-L171C36) - I'm not sure about this so let me know if you think it's a bad idea. 
Another thing would be that it keeps the option open for adding the group attribute support for macro definitions (not sure what are the perspectives here).

maciektr: TBH I don't have strong opinion, but seems okay to me

maciektr (approved):
Looks great!
👏👏👏

PR #2563, opened 2025-08-21
Test `scarb update` in audit and regular context

DelevoXDG:
Adds test for `scarb update`:
- in regular context (no such test was present for registry deps)
- in audits context

## Stack

- #2560
- #2561
- #2617
- #2563
- #2562 
- #2566
- #2575
- #2576

PR #2562, opened 2025-08-21
Add `allow-no-audits` to Scarb.toml to bypass audit requirements

DelevoXDG:
## Stack

- #2560
- #2561
- #2617
- #2563
- #2562 
- #2566
- #2575
- #2576

PR #2575, opened 2025-08-27
Add `scarb list` to list registry package versions 

### LINKED ISSUE #2428: Add `scarb list`
`scarb list` is added that allows listing all package versions available on the default registry, checkmark is added next to "audited" ones

DelevoXDG:
Display package versions with yank and audit status

Closes #2428 

## Stack

- #2560
- #2561
- #2617
- #2563
- #2562 
- #2566
- #2575
- #2576

### DIFF
--- a/scarb/src/bin/scarb/commands/list.rs
+++ b/scarb/src/bin/scarb/commands/list.rs
@@ -1,0 +1,42 @@
+use crate::args::ListCommandArgs;
+use anyhow::{Context, Result};
+use scarb::core::ManifestDependency;
+use scarb::core::registry::DEFAULT_REGISTRY_INDEX;
+use scarb::core::registry::client::cache::RegistryClientCache;
+use scarb::core::{Config, DependencyVersionReq, SourceId};
+use scarb::sources::RegistrySource;
+use std::str::FromStr;
+use url::Url;
+
+#[tracing::instrument(skip_all, level = "info")]
+pub fn run(args: ListCommandArgs, config: &Config) -> Result<()> {
+    list_versions(args, config)
+}
+
+fn list_versions(args: ListCommandArgs, config: &Config) -> Result<()> {
+    let package_name = args.package_name;
+    let index = args.index.unwrap_or(Url::from_str(DEFAULT_REGISTRY_INDEX)?);
+
+    let source_id = SourceId::for_registry(&index)?;
+    let registry_client = RegistrySource::create_client(source_id, config)?;
+    let registry_client = RegistryClientCache::new(source_id, registry_client, config)?;
+
+    let dependency = ManifestDependency::builder()
+        .name(package_name)
+        .version_req(DependencyVersionReq::Any)
+        .build();
+    let records = config
+        .tokio_handle()
+        .block_on(registry_client.get_records_with_cache(&dependency))
+        .with_context(|| {
+            format!(
+                "failed to lookup for `{dependency}` in registry: {}",
+                source_id
+            )
+        })?;
+
+    for record in records.into_iter().rev() {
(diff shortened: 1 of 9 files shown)


scarb/tests/list.rs
maciektr, line 1: I'd add one assert on JSON format - it may actually be usefull sometimes 🤔

website/docs/guides/dependencies.md
```diff
@@ -14,12 +14,14 @@ To add a dependency, simply declare it in your `Scarb.toml`.
...
+
 This example adds a dependency on the [`openzeppelin_merkle_tree`](https://github.com/OpenZeppelin/cairo-contracts)
-package (note that OpenZeppelin is a collection of multiple packages, and we will use only one of them as an example in
-this guide). To see all available versions of some package, you can see the versions pane on the package's
-[registry page](https://scarbs.xyz/packages/openzeppelin_merkle_tree). At the time of writing this guide, the latest
-version of the `openzeppelin_merkle_tree` package is `0.17.0`, which is the version we will use.
+package (note that OpenZeppelin is a collection of multiple packages, and we will use only one of them as an example in this guide).
+To see all available versions of some package, use [`scarb list`](#listing-available-versions-of-the-package) command, or visit the "Versions" pane on the package's [registry page](https://scarbs.xyz/packages/openzeppelin_merkle_tree).
+
+At the time of writing this guide, the latest version of the `openzeppelin_merkle_tree` package is `0.17.0`, which is the version we will use.
```
maciektr, line 24:
```suggestion
At the time of writing this guide, the latest version of the `openzeppelin_merkle_tree` package is `2.0.0`, which is the version we will use.
```
:p

scarb/src/bin/scarb/commands/list.rs
```diff
@@ -0,0 +1,126 @@
...
+
+fn list_versions(args: ListCommandArgs, config: &Config) -> Result<VersionsList> {
+    let package_name = args.package_name;
+    let index = args.index.unwrap_or(Url::from_str(DEFAULT_REGISTRY_INDEX)?);
+
+    let source_id = SourceId::for_registry(&index)?;
+    let registry_client = RegistrySource::create_client(source_id, config)?;
+    let registry_client = RegistryClientCache::new(source_id, registry_client, config)?;
+
+    let dependency = ManifestDependency::builder()
```
maciektr: I wonder if we should check if package is builtin and handle it separately - querying the registry may not make sense for the user, as he can only use the built in version anyway 🤔

website/docs/guides/dependencies.md
DelevoXDG: #2588

scarb/src/bin/scarb/commands/list.rs
DelevoXDG: I agree that this maybe misleading.

But I think that the gist of the command is that it reflects the state of the registry, including potentially a custom one. And, considering that we do have builtin packages hosted on registry, at least in some form, - I believe that hiding them here might be confusing as well.

I would suggest:
- displaying the warning telling that there's only one version available with the current Scarb installation
- still displaying the list of other versions below

This would both inform the user of the state of registry, informing what alternative versions of the toolchain are available, and avoid the problem you mention. 

There's also a matter of json input, which I think we may show, but also may hide the warning, assuming the users who want to deserialize that info for some reason know what they're doing.

WDYT?

maciektr: I like your suggestion! Let's just show a warning before listing the regsitry state then. It should work fine with json format -> warning will be shown as a separate message before the output, as we do in other parts of scarb anyway.

PR #2576, opened 2025-08-27
Limit amount of versions listed by `scarb list`; Add `--all` flag

DelevoXDG:
## Stack

- #2560
- #2561
- #2617
- #2563
- #2562 
- #2566
- #2575
- #2576

### DIFF
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -643,8 +643,13 @@ pub struct ListCommandArgs {
     #[arg(long, value_name = "URL")]
     pub index: Option<Url>,
 
+    /// Show all available versions.
     #[arg(short, long, default_value_t = false)]
     pub all: bool,
+
+    /// Limit the number of results.
+    #[arg(short, long, default_value_t = 5, conflicts_with = "all")]
+    pub limit: usize,
 }
 
 /// Git reference specification arguments.
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -644,11 +644,11 @@ pub struct ListCommandArgs {
     pub index: Option<Url>,
 
     /// Show all available versions.
-    #[arg(short, long, default_value_t = false)]
+    #[arg(long, default_value_t = false)]
     pub all: bool,
 
     /// Limit the number of results.
-    #[arg(short, long, default_value_t = 5, conflicts_with = "all")]
+    #[arg(long, default_value_t = 5, conflicts_with = "all")]
     pub limit: usize,
 }
 
--- a/scarb/src/bin/scarb/args.rs
+++ b/scarb/src/bin/scarb/args.rs
@@ -648,7 +648,7 @@ pub struct ListCommandArgs {
     pub all: bool,
 
     /// Limit the number of results.
-    #[arg(long, default_value_t = 5, conflicts_with = "all")]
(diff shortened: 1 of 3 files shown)


scarb/src/bin/scarb/commands/list.rs
```diff
@@ -46,12 +46,20 @@ fn list_versions(args: ListCommandArgs, config: &Config) -> Result<VersionsList>
         .sorted_by_key(|r| std::cmp::Reverse(r.version.clone()))
         .collect();
 
-    Ok(VersionsList { records })
+    let display_limit = args.all.then(|| None).unwrap_or(Some(5));
```
maciektr: I'd just pass it from args directly. Cargo does:
```
      --limit <LIMIT>            Limit the number of results (default: 10, max: 100)
```

DelevoXDG: Generally, I feel like `--all` takes removes the unnecessary decision for a user regarding what number to specify.

Also I feel like `--limit` flag unnecessary *limiting* 😅 - I think there are cases where one would want all the versions, which is not something possible if we have max value here, or is not super user-friendly if the user has to type `--limit 999`

So, WDYT about just keeping both flags, perhaps making them conflict?

maciektr: 👌

scarb/src/bin/scarb/commands/list.rs
```diff
@@ -94,8 +103,12 @@ impl Message for VersionsList {
         let audit_header = "AUDIT";
         let status_header = "STATUS";
 
-        let version_width = self
-            .records
+        let total = self.records.len();
+        let limit = self.display_limit.unwrap_or(total);
+
+        let records = self.records.into_iter().take(limit).collect::<Vec<_>>();
```
maciektr, line 115: I think this are not guaranteed to be in order, so we should sort from highest ver to lowest ver between taking.

scarb/src/bin/scarb/args.rs
```diff
@@ -643,6 +643,14 @@ pub struct ListCommandArgs {
     /// Registry index URL to query package versions from.
     #[arg(long, value_name = "URL")]
     pub index: Option<Url>,
+
+    /// Show all available versions.
+    #[arg(long, default_value_t = false)]
+    pub all: bool,
+
+    /// Limit the number of results.
+    #[arg(long, default_value_t = 5, conflicts_with = "all")]
```
maciektr:
```suggestion
    #[arg(long, default_value_t = 10, conflicts_with = "all")]
```

5 seems low 😅

scarb/src/bin/scarb/commands/list.rs
DelevoXDG: We sort them based on semver when creating `VersionList` 🤔 

```
    let records = records
        .into_iter()
        .sorted_by_key(|r| std::cmp::Reverse(r.version.clone()))
        .collect();

    let display_limit = if args.all { None } else { Some(args.limit) };

    Ok(VersionsList {
        records,
        display_limit,
    })
```
But, I moved the sorting to `VersionsList::new` so it's a bit more clear

scarb/src/bin/scarb/args.rs
DelevoXDG: Updated

maciektr (approved):

PR #2580, opened 2025-08-28
Use Cargo features to make it possible to disable builtin oracle protocols

mkaput:

PR #2579, opened 2025-08-28
Allow providing protocols externally

mkaput:
Allow providing protocols externally

Also:
- Renamed `ConnectionManager` to `Connections`, for brevity.
- This enables turning some protocols on/off internally as everything is now controlled by one-liners in `OracleHintService::new`.

Extract builtin protocol registration logic into a dedicated function

PR #2578, opened 2025-08-28
Make `OracleCheatcodeSelector` opaque

mkaput:
I don't want anybody to play with this enum outside the hint service.

PR #2581, opened 2025-08-28
Implement `shell` oracle protocol

### LINKED ISSUE #2405: Implement `shell` oracle
### Problem

Implement an `shell` oracle (TBD if builtin or stdio based) that we will publish to scarbs.xyz.

It will take a shell invocation line, run it, and return back an `ByteArray` with stdout.

TBD if we want to handle stdin as well.

### Proposed Solution

_No response_

### Notes

_No response_

mkaput:
fix #2405


mkaput:

- #2578
    - #2579
        - #2580
            - #2581 ◀

utils/scarb-oracle-hint-service/src/connections/shell.rs
```diff
@@ -0,0 +1,157 @@
...
+        // Finalise, according to mode.
+        mode.finalise(exit_code, stdout.into())
+    }
+}
+
+#[derive(Eq, PartialEq)]
+enum Mode {
+    /// `eo`: return (exit code, stdout) tuple.
+    ReturnExitReturnOutput,
+    /// `co`: require successful exit code, return stdout.
```
maciektr, line 70: What does the `co` stand for? 🤔 
This seem kind of confusing for me, but I guess the flexibility is worth it.

utils/scarb-oracle-hint-service/src/connections/shell.rs
```diff
@@ -0,0 +1,157 @@
...
+            }
+        }
+    }
+}
+
+fn build_shell() -> Result<ShellState> {
+    Ok(ShellState::new(
+        env::vars_os().collect(),
+        env::current_dir().context("cannot get current dir")?,
+        Default::default(),
```
maciektr: Would it make sense to provide Scarb as a command, like in scripts? 🤔

utils/scarb-oracle-hint-service/src/connections/shell.rs
```diff
@@ -0,0 +1,157 @@
...
+        let _enter = span.enter();
+
+        // Decode arguments.
+        let mut calldata = calldata.iter();
+        let command: String = ByteArray::decode_iter(&mut calldata)?.try_into()?;
+        debug!("{command}");
+
+        // Run command.
+        let command = parse(&command)?;
+        let shell = build_shell()?;
```
maciektr, line 58: Would it make sense to share the state between calls? 🤔

utils/scarb-oracle-hint-service/src/connections/shell.rs
maciektr: (Would it be possible to keep one of them as a default?)

utils/scarb-oracle-hint-service/src/connections/shell.rs
mkaput: Hard to say because there is no state in this protocol - it's all one-off command runs. If somebody wants to keep a state here, they can always write the state to a file and read it back on next invocation.

utils/scarb-oracle-hint-service/src/connections/shell.rs
mkaput: **c**heck-**o**utput

These names will be more readable when I'll make an idiomatic Cairo library which wraps this protocol. Here, I must fit within shortstrings limits, and these one-letter suffixes were the best I could think of.

utils/scarb-oracle-hint-service/src/connections/shell.rs
mkaput: 4d5b984e279ade110df7a9f0a23f97588d0815f6

utils/scarb-oracle-hint-service/src/connections/shell.rs
maciektr: I guess sharing the state would make it possible to modify env vars with `export` between invocations - not sure if we disallow it on purpose, or by accident?

mkaput: Hmm, never thought of that, but I doubt that this is the way forward. Primarily because there is one shell instance for everything in runtime. You could set up some envs then call into some library, and it might unexpectedly override these. We would have to expose the state to the Cairo code.

I think there is a space for this to introduce as new selectors (spawn shell, run task within it) if/when we'll get feature requests for such thing...

Issue #2526, opened 2025-08-11
Create some ux for defining `build-external-contracts` for all test targets at once


maciektr:
Internal discussion: https://swmansion.slack.com/archives/C03N11B6JB0/p1753961305643769

Example syntax: 

```toml 
[target-defaults.test]
build-external-contracts = [
    "openzeppelin_test_common::mocks::account::DualCaseAccountMock"
]
```

May only accept `build-external-contracts` and reject other keys as errors.

PR #2583, opened 2025-08-28
support default params for test targets

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2526

### DIFF
--- a/scarb/src/core/manifest/toml_manifest.rs
+++ b/scarb/src/core/manifest/toml_manifest.rs
@@ -58,6 +58,13 @@ pub struct TomlManifest {
     pub tool: Option<BTreeMap<SmolStr, MaybeWorkspaceTomlTool>>,
     pub features: Option<BTreeMap<FeatureName, Vec<TomlFeatureToEnable>>>,
     pub patch: Option<BTreeMap<SmolStr, BTreeMap<PackageName, TomlDependency>>>,
+    pub target_defaults: Option<BTreeMap<TargetKind, TargetDefaults>>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct TargetDefaults {
+    pub build_external_contracts: Option<Vec<String>>,
 }
 
 type MaybeWorkspaceScriptDefinition = MaybeWorkspace<ScriptDefinition, WorkspaceScriptDefinition>;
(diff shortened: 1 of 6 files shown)


website/docs/reference/targets.md
maciektr, line 1: Please also update https://github.com/software-mansion/scarb/blob/main/website/docs/reference/manifest.md to showcase the new field 🙏

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -58,6 +58,13 @@ pub struct TomlManifest {
     pub tool: Option<BTreeMap<SmolStr, MaybeWorkspaceTomlTool>>,
     pub features: Option<BTreeMap<FeatureName, Vec<TomlFeatureToEnable>>>,
     pub patch: Option<BTreeMap<SmolStr, BTreeMap<PackageName, TomlDependency>>>,
+    pub target_defaults: Option<BTreeMap<TargetKind, TargetDefaults>>,
```
maciektr, line 61: This is good, but I think we should also allow inheriting this setting from ws, so all of these:

```Scarb.toml
[package] 
name = "hello"

[target-defaults.test]
build-external-contracts = [ "..." ]
```

and 

```Scarb.toml
[package] 
name = "hello"

[workspace.target-defaults.test]
build-external-contracts = [ "..." ]

[target-defaults]
test.workspace = true
```

and 

```Scarb.toml
[package] 
name = "hello"

[workspace.target-defaults.test]
build-external-contracts = [ "..." ]

[target-defaults.test]
build-external-contracts.workspace = true
```

would be valid (and equivalent). 

Can be a separate PR imho.

scarb/tests/build_targets.rs
```diff
@@ -1591,3 +1591,75 @@ fn disallowed_test_target_names() {
...
+            [target-defaults.test]
+            build-external-contracts = [
+                "hello::Balance",
+            ]
+        "#})
+        .dep_starknet()
+        .build(&world);
+
+    Scarb::quick_snapbox()
+        .arg("build")
```
maciektr: Maybe it would be enough to only check targets as represented in `scarb metadata` output - this would run much faster than compiling the contracts 🤔

FroyaTheHen: good call!

scarb/src/core/manifest/toml_manifest.rs
FroyaTheHen: Ok, will do 👌

PR #2586, opened 2025-09-01
Define macro fingerprint callback interface

maciektr:
Change proc macro post process callback into an enum

Define fingerprint callback interface for procedural macro instances

Add macro for defining fingeprint callbacks

PR #2587, opened 2025-09-01
Use macro defined fingerprint in digest calculation

maciektr:
Move proc macro repository access to a trait

Use macro defined fingerprint in digest calculation

Add a test case

### DIFF
--- a/scarb/tests/incremental.rs
+++ b/scarb/tests/incremental.rs
@@ -461,6 +461,114 @@ fn snforge_scarb_plugin_nondeterminism_hack() {
     assert_ne!(digest(hello_component_id.as_str()), hello_digest);
 }
 
+#[test]
+fn fingerprint_callback_can_force_recompilation() {
+    let cache_dir = TempDir::new().unwrap().child("c");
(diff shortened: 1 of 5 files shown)


maciektr:

- #2586
    - #2587 ◀

scarb/tests/incremental.rs
```diff
@@ -461,6 +461,114 @@ fn snforge_scarb_plugin_nondeterminism_hack() {
...
+        #[fingerprint]
+        fn some_value() -> u64 {
+            12
+        }
+
+        #[fingerprint]
+        fn env_hash() -> u64 {
+            let mut hasher = xxhash_rust::xxh3::Xxh3::default();
+            std::env::var("TEST_ENV").unwrap().hash(&mut hasher);
+            hasher.finish()
```
mkaput: I think you can shave some seconds from this test execution time by not depending on xxhash and LLM-rolling your custom hash function instead

maciektr: a4vof9

PR #2598, opened 2025-09-03
Fix completions_block install script

JulianGCalderon:
Running install.sh fails with the following error:

```
install.sh: line 573: bash_completion_block: command not found
```

This is because its trying to execute the function `bash_completion_block`, but the actual name is `bash_completions_block`.

This PR update the script with the correct function names.

With this PR, the instalation doesn't fail:

```
Scarb has been successfully installed and should be already available in your PATH.
Run 'scarb --version' to verify your installation. Happy coding!
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -539,17 +539,17 @@ add_completions_to_profile() {
   */zsh)
     _profile=$HOME/.zshrc
     _pref_shell=zsh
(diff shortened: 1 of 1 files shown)


maciektr (approved):
Thanks so much!!! ❤️

PR #2602, opened 2025-09-05
Read incremental cache files in parallel

maciektr:


scarb/src/compiler/incremental/compilation.rs
```diff
@@ -51,42 +53,70 @@ pub fn load_incremental_artifacts(
...
+        let fingerprints = UnitFingerprint::new(unit, ws).await;
+        let handles = unit
+            .components
+            .iter()
+            .map(|component| {
+                let fingerprint = fingerprints
+                    .get(&component.id)
+                    .expect("component fingerprint must exist in unit fingerprints");
+                let handle = {
+                    // HACK: We know that we will not use &Config outside the scope of this function,
```
FroyaTheHen:
```suggestion
                    // HACK: We know that we will not use &Workspace outside the scope of this function,
```

Issue #2593, opened 2025-09-02
`build-external-contracts` inheriting from workspace


FroyaTheHen:
https://github.com/software-mansion/scarb/pull/2583#discussion_r2310224016

PR #2603, opened 2025-09-06
support workspace target-defaults

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2593

docs update in separate [PR](https://github.com/software-mansion/scarb/pull/2604)

### DIFF
--- a/scarb/src/core/manifest/target.rs
+++ b/scarb/src/core/manifest/target.rs
@@ -1,4 +1,5 @@
-use crate::core::{TargetDefaults, TargetKind, TomlExternalTargetParams};
+use crate::core::manifest::target_defaults::TargetDefaults;
+use crate::core::{TargetKind, TomlExternalTargetParams};
 use crate::internal::restricted_names;
 use crate::internal::serdex::toml_merge;
 use anyhow::{Result, bail};
@@ -69,7 +70,7 @@ impl Target {
         source_path: impl Into<Utf8PathBuf> + Clone,
         group_id: Option<SmolStr>,
         params: impl Serialize,
-        target_defaults: Option<TargetDefaults>,
+        target_defaults: Option<&TargetDefaults>,
     ) -> Result<Self> {
         Self::validate_test_target_file_stem(source_path.clone().into())?;
         let params = toml::Value::try_from(params)?;
@@ -198,7 +199,7 @@ impl TestTargetProps {
 
     pub fn with_build_external_contracts(self, external: Vec<String>) -> Self {
         Self {
-            build_external_contracts: Some(external),
+            build_external_contracts: (!external.is_empty()).then_some(external),
             ..self
         }
     }
--- a/scarb/src/core/manifest/target_defaults.rs
+++ b/scarb/src/core/manifest/target_defaults.rs
@@ -1,0 +1,103 @@
+use crate::core::TargetKind;
+use crate::core::{MaybeWorkspace, WorkspaceInherit};
+use anyhow::{Result, bail};
+use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
+use std::fmt;
+use typetag::serde;
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct TargetDefaults {
+    pub build_external_contracts: Vec<String>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct TomlTargetDefaults {
+    pub build_external_contracts: MaybeWorkspaceBuildExternalContracts,
+}
+
+impl From<TargetDefaults> for TomlTargetDefaults {
+    fn from(value: TargetDefaults) -> Self {
+        Self {
+            build_external_contracts: MaybeWorkspace::Defined(value.build_external_contracts),
+        }
+    }
+}
+
+pub type MaybeWorkspaceTargetDefaults =
+    MaybeWorkspace<TomlTargetDefaults, TomlWorkspaceTargetDefault>;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct TomlWorkspaceTargetDefault {
+    workspace: bool,
+}
+
+impl WorkspaceInherit for TomlWorkspaceTargetDefault {
+    fn inherit_toml_table(&self) -> &str {
+        "target-defaults"
+    }
+
+    fn workspace(&self) -> bool {
+        self.workspace
+    }
+}
+
+pub type MaybeWorkspaceBuildExternalContracts =
+    MaybeWorkspace<Vec<String>, TomlWorkspaceBuildExternalContracts>;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct TomlWorkspaceBuildExternalContracts {
+    workspace: bool,
+}
+
+impl WorkspaceInherit for TomlWorkspaceBuildExternalContracts {
+    fn inherit_toml_table(&self) -> &str {
+        "target-defaults.test"
+    }
+
+    fn workspace(&self) -> bool {
+        self.workspace
+    }
+}
+
+#[derive(Debug, Clone, Serialize, PartialEq, Eq, Ord, PartialOrd, Hash, Deserialize)]
+#[serde(try_from = "TargetKind")]
+pub struct TomlTargetKindTestOnly(TargetKind);
+
+impl TomlTargetKindTestOnly {
+    #[inline(always)]
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+}
+
+impl TryFrom<TargetKind> for TomlTargetKindTestOnly {
+    type Error = anyhow::Error;
+
+    fn try_from(value: TargetKind) -> Result<Self> {
+        if value.is_test() {
+            Ok(Self(value))
+        } else {
+            bail!(
+                "only target kind `test` is allowed in `target_defaults`, but found `{}`",
+                value
+            );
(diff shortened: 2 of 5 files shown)


scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -728,7 +724,35 @@ impl TomlManifest {
...
+            .map(
+                |(target_kind, defaults)| -> Result<(TargetKind, TargetDefaultsDefinition)> {
+                    let inherit_ws = || {
+                        workspace
+                            .target_defaults
+                            .clone()
+                            .and_then(|tds| tds.get(&target_kind).cloned())
+                            .ok_or_else(|| {
+                                anyhow!(
+                                    "defaults for `{}` not found in workspace",
```
maciektr: nit
```suggestion
                                    "defaults for `{}` target not found in workspace",
```

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -992,6 +1037,7 @@ impl TomlManifest {
                     &package_name,
                     root,
                     None,
+                    target_defaults.clone(),
```
maciektr: NIT 
Why not pass by ref?

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -114,6 +109,7 @@ pub struct TomlWorkspace {
     pub dev_dependencies: Option<BTreeMap<PackageName, TomlDependency>>,
     pub scripts: Option<BTreeMap<SmolStr, ScriptDefinition>>,
     pub tool: Option<TomlToolsDefinition>,
+    pub target_defaults: Option<BTreeMap<TargetKind, TargetDefaultsDefinition>>,
```
maciektr: Since we only support the test target kind, I think it would make sense to validate that during deserialization 🤔 
I think this can be done with newtype and serde try_from 🤔

scarb/src/core/manifest/target_defaults.rs
```diff
@@ -0,0 +1,63 @@
...
+        "target-defaults.test"
+    }
+
+    fn workspace(&self) -> bool {
+        self.workspace
+    }
+}
+
+impl TargetDefaults {
+    pub fn resolve(
```
maciektr: It seems to me, this would be more idiomatic if we kept relying on `MaybeWorkspace`. 
We could do something like: 
```
use crate::core::{MaybeWorkspace, WorkspaceInherit};
use anyhow::{Context, bail};
use serde::{Deserialize, Serialize};

#[derive(Debug, Clone, Deserialize, Serialize)]
#[serde(rename_all = "kebab-case")]
pub struct TargetDefaults {
    pub build_external_contracts: Vec<String>,
}

#[derive(Debug, Clone, Deserialize, Serialize)]
#[serde(rename_all = "kebab-case")]
pub struct TomlTargetDefaults {
    pub build_external_contracts: MaybeWorkspaceBuildExternalContracts,
}

impl From<TargetDefaults> for TomlTargetDefaults {
    fn from(value: TargetDefaults) -> Self {
        Self {
            build_external_contracts: MaybeWorkspace::Defined(value.build_external_contracts),
        }
    }
}

pub type MaybeWorkspaceTargetDefaults =
    MaybeWorkspace<TomlTargetDefaults, TomlWorkspaceTargetDefault>;

#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(rename_all = "kebab-case")]
pub struct TomlWorkspaceTargetDefault {
    workspace: bool,
}

impl WorkspaceInherit for TomlWorkspaceTargetDefault {
    fn inherit_toml_table(&self) -> &str {
        "target-defaults"
    }

    fn workspace(&self) -> bool {
        self.workspace
    }
}

pub type MaybeWorkspaceBuildExternalContracts =
    MaybeWorkspace<Vec<String>, TomlWorkspaceBuildExternalContracts>;

#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(rename_all = "kebab-case")]
pub struct TomlWorkspaceBuildExternalContracts {
    workspace: bool,
}

impl WorkspaceInherit for TomlWorkspaceBuildExternalContracts {
    fn inherit_toml_table(&self) -> &str {
        "target-defaults.test"
    }

    fn workspace(&self) -> bool {
        self.workspace
    }
}
```
and then 
```
                    let inherit_ws = || {
                        workspace
                            .target_defaults
                            .clone()
                            .and_then(|tds| tds.get(&target_kind).cloned())
                            .map(Into::into)
                            .ok_or_else(|| {
                                anyhow!(
                                    "defaults for `{}` not found in workspace",
                                    target_kind.clone()
                                )
                            })
                    };
                    let target_defaults = defaults.resolve(target_kind.as_str(), inherit_ws)?;
                    let inherit_ws = || {
                        workspace
                            .target_defaults
                            .clone()
                            .and_then(|tds| tds.get(&target_kind).cloned())
                            .map(|td| td.build_external_contracts.clone())
                            .ok_or_else(|| {
                                anyhow!(
                                    "defaults for `{}` not found in workspace",
                                    target_kind.clone()
                                )
                            })
                    };
                    let resolved = TargetDefaults {
                        build_external_contracts: target_defaults
                            .build_external_contracts
                            .resolve("build-external-contracts", inherit_ws)?,
                    };
                    Ok((target_kind.clone(), resolved))
```
wdyt?

FroyaTheHen: Honestly, I have no idea 😅 I am trying to figure out "the best way" to handle this case and I do not really like any of the solutions. This is indeed more idiomatic and I will be happy to make the changes.  
However, I think that there is a space for making the _workspace inheritance_ code more concise. At least implementing some macros for the repetitive parts 🤔 But! That's probably just me coming to Rust from Python...

scarb/tests/build_targets.rs
```diff
@@ -1672,3 +1674,132 @@ fn test_target_defaults() {
...
+            version = "0.1.0"
+
+            [workspace.target-defaults.test]
+            build-external-contracts = [
+                "hello::Balance",
+            ]
+
+            {}
+
+            [[test]]
```
maciektr, line 1720: I think we should also have a test case, that tests adding the `build-external-contracts` to auto-detected test target through target defaults 🤔

scarb/src/core/manifest/target_defaults.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    fn workspace(&self) -> bool {
+        self.workspace
+    }
+}
+
+#[derive(Debug, Clone, Serialize, PartialEq, Eq, Ord, PartialOrd, Hash, Deserialize)]
+#[serde(try_from = "TargetKind")]
+pub struct TomlTargetKind(TargetKind);
```
maciektr:
```suggestion
pub struct TomlTargetKindTestOnly(TargetKind);
```
I think this should be very explicit about the context in which it can be used.

maciektr (changes requested):
Two comments, otherwise seems ready to go :D

scarb/tests/build_targets.rs
FroyaTheHen: good idea! see `test_target_defaults_does_not_override_auto_detected_targets` below. Is that what you had in mind?

scarb/src/core/manifest/target.rs
```diff
@@ -220,6 +221,41 @@ impl TryInto<TomlExternalTargetParams> for TestTargetProps {
     }
 }
 
+/// Custom merge function `toml_merge` alike,
+/// except an empty array in the source does not overwrite
+/// a non-empty array in the target.
+pub fn toml_merge_skip_empty_source<'de, T, S>(target: &T, source: &S) -> Result<T>
```
maciektr: I think it would be simpler to keep the same merge logic, but skip adding the `build_external_contracts` param for empty arrays in the auto-detect logic.

```
    pub fn with_build_external_contracts(self, external: Vec<String>) -> Self {
        Self {
-           build_external_contracts: Some(external),
+           build_external_contracts: (!external.is_empty()).then_some(external),
            ..self
        }
    }
```

scarb/tests/build_targets.rs
```diff
@@ -1672,3 +1674,193 @@ fn test_target_defaults() {
...
+        .dep_starknet()
+        .lib_cairo(format!("{BALANCE_CONTRACT}\n{HELLO_CONTRACT}"))
+        .build(&hello);
+
+    ProjectBuilder::start()
+        .name("world")
+        .edition("2023_01")
+        .version("0.1.0")
+        .dep("hello", Dep.path("../hello"))
+        .manifest_extra(formatdoc! {r#"
```
maciektr, line 1830: Please add an cairo integration test file here too.

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -1053,6 +1122,7 @@ impl TomlManifest {
                     &package_name,
                     root,
                     None,
+                    target_defaults.as_ref(),
                 )?);
             } else {
```
maciektr, line 1134: There is still one last problem to tackle 😅 

When compiling integration tests, we also compile contracts from the main package, so they are available in the runtime. We do this, by adding the package itself, to the build-external-contracts filed in the auto-detected test target. 
https://github.com/software-mansion/scarb/blob/4e3df993e627fcc39f6fce2343f7bd229bd1cfcd/scarb/src/core/manifest/toml_manifest.rs#L1088
Since we do this before the `try_from_structured_params` called, this would always overwrite the defaults. 
We need to find a way, so that: 
1) we can have the defaults for integration test targets
2) we do inject the main package, so that the contracts are compiled

FroyaTheHen: Sure! please see the changes! 😄

scarb/src/core/manifest/target.rs
FroyaTheHen: good idea 👌

PR #2620, opened 2025-09-10
Ignore flaky https_something_happens on windows runners

maciektr:
Getting tired of failed runs, I don't think it makes sense to keep running this test on windows

### DIFF
--- a/scarb/tests/git_source_network.rs
+++ b/scarb/tests/git_source_network.rs
@@ -9,6 +9,10 @@ use scarb_test_support::command::Scarb;
 use scarb_test_support::project_builder::{Dep, DepBuilder, ProjectBuilder};
 
 #[test]
+#[cfg_attr(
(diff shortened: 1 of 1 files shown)


scarb/tests/git_source_network.rs
```diff
@@ -9,6 +9,10 @@ use scarb_test_support::command::Scarb;
 use scarb_test_support::project_builder::{Dep, DepBuilder, ProjectBuilder};
 
 #[test]
+#[cfg_attr(
+    not(target_family = "unix"),
```
DelevoXDG, line 13: nit: why not?
```suggestion
    target_family = "windows",
```

maciektr: Just copied it - I see we have both. Shouldn't make a difference 🤷‍♂️

PR #2633, opened 2025-09-11
Handle selector ambiguities in `wasm` oracle

mkaput:


utils/scarb-oracle-hint-service/src/connections/wasm/mod.rs
DelevoXDG, line 1: Logic seems correct, but I notice we don't test ambiguities handling in #2634 - shouldn't we? 🤔

mkaput: Hmm, technically the `naked` test touches this functionality. It doesn't test actual ambiguity (i.e. two symbols that collide by short name). I'll add it there 🫡

mkaput: 1e2c5f333fcbf01ba55120de70351a362905b06e
7c4c5859583cd1d599b5540169a9d89ddecd61ab

PR #2644, opened 2025-09-16
Save trace data for cairo-profiler

THenry14:
This change adds support for generating trace data files compatible with cairo-profiler.
Related to https://github.com/software-mansion/cairo-profiler/issues/192

### DIFF
--- a/extensions/scarb-execute/src/lib.rs
+++ b/extensions/scarb-execute/src/lib.rs
@@ -189,7 +189,7 @@ pub fn execute(
             }
         })?;
 
-    let execution_resources = (cairo_run_config.relocate_mem && cairo_run_config.trace_enabled)
+    let execution_resources = (args.run.print_resource_usage || args.run.save_profiler_trace_data)
         .then(|| ExecutionResources::try_new(&runner, hint_processor.cairo_hint_processor).ok())
         .flatten();
 
@@ -277,7 +277,7 @@ pub fn execute(
              help: run `scarb build` to compile the package and try again.",
             build_target.name
         );
-        let tracked_resource = get_profiler_tracked_resource(package);
+        let tracked_resource = get_profiler_tracked_resource(package)?;
         let function_name: Option<String> = build_target
             .params
             .get("function")
--- a/extensions/scarb-execute/src/profiler.rs
+++ b/extensions/scarb-execute/src/profiler.rs
@@ -18,11 +18,18 @@ pub enum TrackedResource {
     SierraGas,
 }
 
-impl From<&str> for TrackedResource {
-    fn from(m: &str) -> Self {
+impl TryFrom<&str> for TrackedResource {
+    type Error = anyhow::Error;
+
+    fn try_from(m: &str) -> Result<Self, Self::Error> {
         match m {
-            "sierra-gas" => TrackedResource::SierraGas,
-            _ => TrackedResource::CairoSteps,
+            "sierra-gas" => Ok(TrackedResource::SierraGas),
+            "cairo-steps" => Ok(TrackedResource::CairoSteps),
+            other => Err(anyhow!(
+                "Invalid tracked resource set for profiler: {}\
+            \nhelp: valid options are: `cairo-steps` or `sierra-gas`",
+                other
+            )),
         }
     }
 }
@@ -81,14 +88,16 @@ pub fn build_profiler_call_trace(
     })
 }
 
-pub fn get_profiler_tracked_resource(package_metadata: &PackageMetadata) -> TrackedResource {
+pub fn get_profiler_tracked_resource(
+    package_metadata: &PackageMetadata,
+) -> Result<TrackedResource> {
     let tracked_resource = package_metadata
         .tool_metadata("cairo-profiler")
         .and_then(|val| val.get("tracked-resource"))
         .and_then(|val| val.as_str())
         .unwrap_or("cairo-steps");
 
-    TrackedResource::from(tracked_resource)
+    TrackedResource::try_from(tracked_resource)
 }
 
 fn build_profiler_call_entry_point(
--- a/extensions/scarb-execute/tests/build.rs
+++ b/extensions/scarb-execute/tests/build.rs
@@ -546,3 +546,37 @@ fn no_build_artifact_for_profiler_trace_file() {
         error: Missing sierra code for executable `hello`, file [..]hello.executable.sierra.json does not exist. help: run `scarb build` to compile the package and try again.
         "#});
 }
+
+#[test]
+fn invalid_tracked_resource_for_profiler_trace_file() {
+    let t = TempDir::new().unwrap();
+    executable_project_builder()
+        .manifest_extra(indoc! {r#"
+            [executable]
+            sierra = true
+            [cairo]
+            enable-gas = false
+            [tool.cairo-profiler]
+            tracked-resource = "whatever"
+        "#})
+        .lib_cairo(indoc! {r#"
+            #[executable]
+            fn main() -> felt252 { 1 }
+        "#})
+        .build(&t);
+
+    Scarb::quick_snapbox()
+        .arg("execute")
+        .arg("--save-profiler-trace-data")
+        .current_dir(&t)
+        .assert()
+        .failure()
+        .stdout_matches(indoc! {r#"
+        [..]Compiling hello v0.1.0 ([..]Scarb.toml)
+        [..]Finished `dev` profile target(s) in [..]
+        [..]Executing hello
+        Saving output to: target/execute/hello/execution1
+        error: Invalid tracked resource set for profiler: whatever
+        help: valid options are: `cairo-steps` or `sierra-gas`
+        "#});
+}


extensions/scarb-execute/src/profiler.rs
```diff
@@ -0,0 +1,109 @@
...
+        used_l1_resources: Default::default(),
+        nested_calls: Default::default(),
+        cairo_execution_info: Some(cairo_execution_info),
+    })
+}
+
+pub fn get_profiler_tracked_resource(package_metadata: &PackageMetadata) -> TrackedResource {
+    let tracked_resource = package_metadata
+        .tool_metadata("cairo-profiler")
+        .and_then(|val| val.get("tracked-resource"))
```
THenry14, line 96: **A bit of context first:**
Profiler shows different samples based on a "tracked resource"

the samples can be in either one of sets:
- steps, memory holes, *_builtin
- sierra gas

and up until now they were treated as exclusive to each other (for reasons, it should stay that way - at least for near future).

In snforge this was being determined by execution context itself (or to be more precise - by the transaction context; underneath everything went through cairo-vm, but based on some requirements, a user could imitate a native environment).

Here we're just executing everything through cairo-vm, but there is nothing preventing us from calculating sierra gas for a resulting trace as well. We need, however, a way for a user to express that this is indeed what one wants.

**Now off to a question:**
What is the best place to allow the user to set this? Right now I have added it under `tool.cairo-profiler`, but technically it is being used in scarb execute only (ie it is being used inside the execute to determine the final format of a trace data, the profiler just takes the info straight from the file).

extensions/scarb-execute/src/lib.rs
```diff
@@ -249,6 +255,35 @@ pub fn execute(
         fs::write(air_private_input_path, output_value)?;
     }
 
+    if args.run.save_profiler_trace_data {
+        let profiler_trace_path = Utf8PathBuf::from(format!(
```
maciektr: This seems problematic, as the sierra file is not guaranteed to be there - it's build with Scarb's `lib` target, while the execute runs on `executable` target, so the lib target may not even be defined in this package. We should add some validations on that.

extensions/scarb-execute/src/profiler.rs
maciektr: Would you be so kind as to elaborate on wdym by **execution / transaction context** and **based on some requirements, a user could imitate a native environment**? 🤔 

> is being used in scarb execute only (ie it is being used inside the execute to determine the final format of a trace data, the profiler just takes the info straight from the file).

Personally, I would still namespace it under profiler, as it's not usefull at all unless you use the profiler.

extensions/scarb-execute/src/profiler.rs
```diff
@@ -0,0 +1,109 @@
...
+            vm_trace: profiler_vm_trace,
+        },
+        source_sierra_path,
+    };
+
+    let mut execution_resources: ProfilerExecutionResources =
+        vm_resources.try_into().map_err(|e: String| anyhow!(e))?;
+
+    // currently (CallTraceV1) cairo-profiler uses vm_resources and gas_consumed to determine what was tracked
+    // so we need to reset a value for specific tracked resource, to not confuse it
```
maciektr: So did we track both - and now ignore one of them, to pretend that we didn'? 🤔

Cargo.toml
```diff
@@ -44,6 +44,7 @@ async-trait = "0.1"
 axum = { version = "0.6", features = ["http2"] }
 bincode = "2.0.1"
 bumpalo = "=3.17.0"
+cairo-annotations = "0.5.2"
```
maciektr: We might consider splitting the cairo-annotations lib with features, so we do not depend on cairo through this.

extensions/scarb-execute/src/profiler.rs
THenry14: Since sierra 1.7.0+ all declared classes are compiled via cairo-native in addition to "regular" compilation - and with this introduction of cairo native a problem arose - when we're executing a transaction (or a code for that matter), what resource should be tracked (so what resource should be consumed)? 

Up then the fee was based of number of steps and number of builtin application, but now (sierra 1.7.0+) it can also be based purely on sierra gas consumption. 

BUT, sierra version alone does not determine which resource will be consumed (you can still execute in the old way, tracking steps with newest sierras) - it depends on a parent call as well, as resource counting is hereditary. This means that, for example, if your account is "older" and its calls track cairo steps, the called code will always track it as well.

In snforge we are able to influence the tracked resource, by setting it in the transaction context (basically to emulate using older/newer accounts, contract calls etc). 

Technically there is nothing preventing us from allowing setting tracked resource in scarb as well, as people are still interested in profiling their code for both, that's why the proposed change.

Please let me know if this answers your question and gives you neccesary context. If something is not clear yet, let me know and I will try to help.

extensions/scarb-execute/src/lib.rs
THenry14: Oh yes, 100% right, I'm planning to add all the necessary validations to the final PR :+1:

extensions/scarb-execute/src/profiler.rs
THenry14: Not really - what we track in the profiler, are pc-s. Based on the execution context (on the tracked resource) and those pc-s, we calculate cairo steps, used builtins, memory holes and stuff OR sierra gas

maciektr: NIT
```suggestion
    // The `CallTraceV1` cairo-profiler trace format uses `vm_resources` and `gas_consumed` to determine what was tracked. 
    // Exactly one of the two should be defined - indicating the resource that has been tracked, while the other should be reset.
    // Since we cannot infer the tracked resources from execution directly, we mask one of the fields based on user provided configuration.
    // This maintains compatibility with other execution environments that produce traces in `CallTraceV1` format.
```

Cargo.toml
```diff
@@ -44,6 +44,7 @@ async-trait = "0.1"
 axum = { version = "0.6", features = ["http2"] }
 bincode = "2.0.1"
 bumpalo = "=3.17.0"
+cairo-annotations = {git = "https://github.com/software-mansion/cairo-annotations", branch = "szymczyk/bumps"}
```
THenry14: I have pinned the `starknet-types-core` to 0.2.0 (and 0.2.1 was yanked :+1:), so the compilation works now.
https://github.com/software-mansion/scarb/actions/runs/17981672922/job/51149087282

Will change it to new version after the PR is merged https://github.com/software-mansion/cairo-annotations/pull/103

THenry14: https://github.com/software-mansion/scarb/pull/2644/commits/9bd7e378e66e46ea57259a7c52b409abc8ed5908

extensions/scarb-execute/src/lib.rs
```diff
@@ -251,6 +261,59 @@ pub fn execute(
         fs::write(air_private_input_path, output_value)?;
     }
 
+    if args.run.save_profiler_trace_data {
+        ensure!(
+            package.targets.iter().any(|t| t.kind == "executable"
```
THenry14: https://github.com/software-mansion/scarb/pull/2676#discussion_r2403250564

maciektr: Checking for `any` is not enough - you need to compile **the target you are executing** to sierra. You know which target is executed from https://github.com/software-mansion/scarb/blob/c79063cbb9380a6cd4526a833dd1dc59c1638322/extensions/scarb-execute/src/lib.rs#L92

extensions/scarb-execute/src/lib.rs
```diff
@@ -251,6 +261,59 @@ pub fn execute(
...
+                && t.params.get("sierra").and_then(|v| v.as_bool()) == Some(true)),
+            "Failed to write profiler trace data into a file - missing sierra code. \
+            Set `sierra = true` under your `[executable]` target in the config and try again."
+        );
+        let profiler_trace_path = Utf8PathBuf::from(format!(
```
maciektr: Please make sure this file exists and show an error otherwise. See: https://github.com/software-mansion/scarb/blob/c79063cbb9380a6cd4526a833dd1dc59c1638322/extensions/scarb-execute/src/lib.rs#L414

extensions/scarb-execute/src/lib.rs
```diff
@@ -251,6 +261,59 @@ pub fn execute(
...
+            "{}/{}/{}.executable.sierra.json",
+            &metadata.workspace.root, &scarb_build_dir, &build_target.name
```
maciektr: scarb_build_dir should be enough, it is an absolute path.

```suggestion
            &scarb_build_dir, &build_target.name
```

THenry14: https://github.com/software-mansion/scarb/pull/2644/commits/6f7f34d4a6427217c0ef51132c80f4fcedc19e7c

extensions/scarb-execute/src/lib.rs
THenry14: https://github.com/software-mansion/scarb/pull/2644/commits/6f7f34d4a6427217c0ef51132c80f4fcedc19e7c

extensions/scarb-execute/src/lib.rs
THenry14: added check with build target in https://github.com/software-mansion/scarb/pull/2644/commits/6f7f34d4a6427217c0ef51132c80f4fcedc19e7c

extensions/scarb-execute/src/lib.rs
```diff
@@ -183,6 +189,10 @@ pub fn execute(
             }
         })?;
 
+    let execution_resources = (cairo_run_config.relocate_mem && cairo_run_config.trace_enabled)
```
maciektr: This changes the previous behaviour a bit.

Previously, for values `output=standard ; print-resource-usage=false` you would never construct `ExecutionResources` ; but now for values `output=standard ; print-resource-usage=false ; save-profiler-trace=false` you would construct it and never use. 

Maybe this should be `args.run.print_resource_usage || args.run.save_profiler_trace_data` instead?

extensions/scarb-execute/src/profiler.rs
```diff
@@ -0,0 +1,118 @@
...
+    #[default]
+    CairoSteps,
+    SierraGas,
+}
+
+impl From<&str> for TrackedResource {
+    fn from(m: &str) -> Self {
+        match m {
+            "sierra-gas" => TrackedResource::SierraGas,
+            _ => TrackedResource::CairoSteps,
```
maciektr: Shouldn't we do TryFrom and validate the input instead 🤔 e.g. if someone types sierragas it probably shouldn't be treated as cairo-steps.

THenry14: yeah, good suggestion :+1: it's better to be explicit here
https://github.com/software-mansion/scarb/pull/2644/commits/d11e082cdb4072e842a11f0559d4d8dd6ddafab5

extensions/scarb-execute/src/lib.rs
THenry14: sure https://github.com/software-mansion/scarb/pull/2644/commits/d11e082cdb4072e842a11f0559d4d8dd6ddafab5

PR #2645, opened 2025-09-17
Allow Proc Macro Server to load macros from arbitrary workspaces at runtime

integraledelebesgue:


scarb/src/ops/proc_macro_server/methods/discover_workspace.rs
```diff
@@ -0,0 +1,97 @@
...
+    ) -> anyhow::Result<Self::Response> {
+        let DiscoverWorkspaceParams { workspace } = params;
+        trace!(
+            "[PMS] Running discoverWorkspace at {}",
+            workspace.manifest_path.to_string_lossy().to_string()
+        );
+
+        let config = Config::builder(workspace.manifest_path.to_string_lossy().to_string())
+            .load_proc_macros(true)
+            .build()?;
```
piotmag769: This should be taken from [this](https://github.com/software-mansion/scarb/pull/2645/files#diff-c91d519e89944f141ae6a755728acd33b8aaebdeffc554d9a1390522f18faa9eR14) place with only manifest file overridden

maciektr: I'd even consider removing the manifest_path field from Config / making it optional / etc and sharing the config throughout all workspaces, as it has basically always been treateed as singleton .
It does stuff like kicking off tokio threadpools, so we do want to share those.

scarb/src/ops/proc_macro_server/store.rs
```diff
@@ -0,0 +1,26 @@
+use std::{collections::HashMap, sync::Arc};
+
+use scarb_proc_macro_server_types::scope::{ProcMacroScope, Workspace};
```
maciektr: NIT: I'd consider renaming this import to `WorkspaceScope` or using as `scope::Workspace` - but I'm very used to seeing `Workspace`  in Scarb :p

scarb/src/ops/proc_macro_server/methods/discover_workspace.rs
```diff
@@ -0,0 +1,97 @@
...
+        trace!("[PMS] Config built");
+
+        let ws = ops::read_workspace(config.manifest_path(), &config)?;
+        let resolve = ops::resolve_workspace_with_opts(&ws, &Default::default())?;
```
maciektr: Note this is very likely to fail in real-world scenarios, for instance due to version requirements conflict in user input. How is this case handled? 🤔

scarb/src/ops/proc_macro_server/methods/discover_workspace.rs
integraledelebesgue: Fixed. Now the config is shared between workspaces. I temporarily achieved it using `mem::transmute`, but I'm going to use `Pin` instead.

scarb/src/ops/proc_macro_server/methods/discover_workspace.rs
integraledelebesgue: All errors returned from `resolve_workspace_with_opts` are sent to the LS and displayed in the logs. Not much more we can do here, that's the standard way of handling errors in proc macro server.

scarb/src/ops/proc_macro_server/store.rs
integraledelebesgue: Sure, it can be `scope::Workspace` if you find it more readable

integraledelebesgue: Done

scarb/src/ops/proc_macro_server/methods/discover_workspace.rs
integraledelebesgue: Fixed. Now the config is shared between workspaces. I used a hack with `mem::transmute` to make it `'static`. Otherwise it cannot be propagated between threads.

scarb/src/ops/proc_macro_server/methods/discover_workspace.rs
piotmag769: Also if this fails it means metadata failed as well, we wouldn't have sent the request in the first place. And if someone changed sth in Scarb.toml in the mean time and it made it fail, we will reload the workspace on LS side (and handle it accordingly), so it's fine as well

scarb/src/ops/proc_macro_server/methods/discover_workspace.rs
```diff
@@ -0,0 +1,90 @@
...
+    let compilation_units = ops::generate_compilation_units(
+        &resolve,
+        &FeaturesOpts {
+            features: FeaturesSelector::AllFeatures,
+            no_default_features: false,
+        },
+        &ws,
+        CompilationUnitsOpts {
+            ignore_cairo_version: true,
+            load_prebuilt_macros: config.load_prebuilt_proc_macros(), // TODO: Control this via flag
```
piotmag769: We don't need to now, no?

scarb/src/ops/proc_macro_server/methods/discover_workspace.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use crate::{
+    compiler::{
+        self, CompilationUnit, CompilationUnitAttributes, plugin::collection::WorkspaceProcMacros,
+    },
+    core::Config,
+    ops::{self, CompilationUnitsOpts, FeaturesOpts, FeaturesSelector, store::ProcMacroStore},
+};
+
+pub fn discover_workspace(
```
piotmag769: I assume this is ~copy-paste code that was used when starting PMS before

scarb/src/ops/proc_macro_server/store.rs
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    pub fn get_workspace_macros(
+        &self,
+        workspace: &scope::Workspace,
+    ) -> Option<&WorkspaceProcMacros> {
+        self.workspace_macros.get(workspace)
+    }
+
+    pub fn get_plugin(&self, scope: &ProcMacroScope) -> Option<Arc<Vec<ProcMacroHostPlugin>>> {
```
piotmag769:
```suggestion
    pub fn get_plugins(&self, scope: &ProcMacroScope) -> Option<Arc<Vec<ProcMacroHostPlugin>>> {
```

scarb/src/ops/proc_macro_server/methods/discover_workspace.rs
maciektr: > Also if this fails it means metadata failed as well, we wouldn't have sent the request in the first place

Are you sure you guarantee that the manifest cannot be changed in the meantime?

maciektr: >All errors returned from resolve_workspace_with_opts are sent to the LS and displayed in the logs. Not much more we can do here, that's the standard way of handling errors in proc macro server.

Great! Just making sure it's handled :)

piotmag769: >Are you sure you guarantee that the manifest cannot be changed in the meantime?

We don't, but see the second sentence of that comment. If any Scarb.toml changes we wipe all the crates from the db and attempt to load them once again

scarb/src/ops/proc_macro_server/methods/discover_workspace.rs
integraledelebesgue: Indeed, I deleted the comment

scarb/src/ops/proc_macro_server/methods/discover_workspace.rs
integraledelebesgue: Yes, that's approximately the same logic. The only change is the usage of "internal" functions like `resolve_workspace_with_opts` instead of wrapper functions (e.g. `resolve_workspace`) which spawn unnecessary threads

scarb/src/compiler/plugin/proc_macro/compilation.rs
```diff
@@ -211,12 +211,20 @@ impl<'c> From<CargoCommand<'c>> for Command {
...
+        |line: &str| {
+            trace!(line);
+            config.ui().print(PipedText::new(line))
+        },
+        |line: &str| {
+            trace!(line);
+            config.ui().print(PipedText::new(line))
+        },
+    );
+    trace!("[PMS-Compiler] exec finished");
```
maciektr: This is not exclusive for PMS.

scarb/src/ops/compile.rs
```diff
@@ -217,17 +217,25 @@ pub fn compile_unit(unit: CompilationUnit, ws: &Workspace<'_>) -> Result<()> {
...
 #[tracing::instrument(skip_all, level = "trace")]
-fn compile_unit_inner(unit: CompilationUnit, ws: &Workspace<'_>) -> Result<()> {
+pub fn compile_unit_inner(unit: CompilationUnit, ws: &Workspace<'_>) -> Result<()> {
     let package_name = unit.main_package_id().name.clone();
 
     let result = match unit {
         CompilationUnit::ProcMacro(unit) => {
             if unit.prebuilt.is_some() {
+                trace!(
+                    "[PMS-compiler] {} unit is prebuilt",
```
maciektr: This is not exclusive to PMS.

scarb/src/ops/resolve.rs
```diff
@@ -343,14 +344,18 @@ pub fn resolve_workspace_with_opts(
             patch_map.warn_unused(ws.config().ui());
 
             let packages = collect_packages_from_resolve_graph(&resolve, &patched).await?;
+            trace!("[PMS-Resolve] Packages collected");
```
maciektr: vide supra

maciektr (changes requested):
Looks great!

piotmag769 (approved):
Really good job 🚀  Please merge the requests we talked about

scarb/src/compiler/plugin/proc_macro/compilation.rs
integraledelebesgue: Those traces were used only for my own debugging. I deleted all of them

scarb/src/ops/proc_macro_server/store.rs
integraledelebesgue: Done

maciektr (approved):

scarb/src/ops/proc_macro_server/store.rs
```diff
@@ -0,0 +1,29 @@
...
+pub struct ProcMacroStore {
+    workspace_macros: HashMap<scope::Workspace, WorkspaceProcMacros>,
+}
+
+impl ProcMacroStore {
+    pub fn insert(&mut self, workspace: scope::Workspace, workspace_macros: WorkspaceProcMacros) {
+        self.workspace_macros.insert(workspace, workspace_macros);
+    }
+
+    pub fn get_workspace_macros(
```
piotmag769: Is it used anywhere now?

integraledelebesgue: No, it isn't anymore. Thanks for the catch

scarb/src/ops/proc_macro_server/methods/defined_macros.rs
```diff
@@ -1,81 +1,157 @@
...
 }
+
+/// Converts the given path to UTF-8.
+fn convert_manifest_path_to_utf8(manifest_path: &Path) -> Result<&Utf8Path> {
+    let os_string = manifest_path.as_os_str();
+    let regular_string = manifest_path
+        .to_str()
+        .with_context(|| format!("manifest path {os_string:?} contains non-UTF-8 characters"))?;
+    Ok(Utf8Path::new(regular_string))
+}
```
maciektr: Maybe just: 
https://github.com/software-mansion/scarb/blob/0d4e505cc05d6dedbd7121d20518fd45a0e7fe64/scarb/src/internal/fsx.rs#L167-L167

maciektr (approved):

scarb/src/ops/proc_macro_server/methods/defined_macros.rs
integraledelebesgue: Didn't notice that method, thank you :)

utils/scarb-proc-macro-server-types/src/scope.rs
```diff
@@ -32,9 +34,15 @@ impl CompilationUnitComponent {
     }
 }
 
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize, Hash)]
```
Draggu, line 38: doc

PR #2656, opened 2025-09-22
Do not calculate warnings for cached crates if printing warnings is disabled

maciektr:


maciektr:

- #2656 ◀

scarb/src/compiler/helpers.rs
```diff
@@ -57,13 +57,31 @@ where
...
         .collect_vec();
-    let crates_to_check: HashSet<CrateId<'db>> = db
-        .crates()
-        .iter()
-        .filter(|crate_id| !cached_crates.contains(&crate_id.long(db).clone().into_crate_input(db)))
-        .chain(main_crate_ids)
-        .copied()
-        .collect();
+    // If a crate is cached, we do not need to check it for diagnostics,
+    // as the cache can only be produced if the crate is diagnostic-free.
```
DelevoXDG: Nit
This seems to contradict the statement below?

> // Note we may need to add the main crates to display warnings generated from them.
> // This is because warnings do not fail compilation, so we can produce caches for crates with them.

I'd rather clarify that cache can be produced only for crates free of error diagnostics, but since we do not produce warnings for dependencies, those may only emit error diagnostics and thus are not included in "crates to check"

DelevoXDG (approved):

FroyaTheHen (approved):
lgtm 👌

scarb/src/compiler/helpers.rs
maciektr: 👍 reworded

PR #2658, opened 2025-09-22
Save information about warnings in cache

maciektr:
This way we can skip calculating diagnostics for main crates.

### DIFF
--- a/scarb/src/compiler/incremental/compilation.rs
+++ b/scarb/src/compiler/incremental/compilation.rs
@@ -3,7 +3,6 @@ use crate::compiler::incremental::fingerprint::{
 };
 use crate::compiler::{CairoCompilationUnit, CompilationUnitComponent};
 use crate::core::Workspace;
-use crate::internal::fsx;
 use anyhow::{Context, Result};
 use cairo_lang_filesystem::db::{CrateConfiguration, FilesGroup};
 use cairo_lang_filesystem::ids::{BlobLongId, CrateInput};
@@ -13,9 +12,10 @@ use cairo_lang_lowering::db::LoweringGroup;
 use cairo_lang_utils::Intern;
 use itertools::Itertools;
 use salsa::{Database, par_map};
-use std::io::Write;
+use std::io::{BufReader, Write};
 use std::ops::Deref;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::{env, mem};
 use tokio::task::spawn_blocking;
 use tracing::{debug, error, trace_span};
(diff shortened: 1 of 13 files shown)


maciektr:
Before:
```
Benchmark 1: /Users/maciektr/Projects/scarb/target/release/scarb build -w --test
  Time (mean ± σ):     864.7 ms ±  21.1 ms    [User: 730.4 ms, System: 75.4 ms]
  Range (min … max):   837.8 ms … 887.5 ms    4 runs

Benchmark 1: /Users/maciektr/Projects/scarb/target/release/scarb build -w
  Time (mean ± σ):     455.1 ms ±  11.7 ms    [User: 294.6 ms, System: 34.3 ms]
  Range (min … max):   440.6 ms … 467.7 ms    4 runs

Benchmark 1: /Users/maciektr/Projects/scarb/target/release/scarb build -w --test
  Time (mean ± σ):     39.602 s ±  0.561 s    [User: 76.977 s, System: 6.189 s]
  Range (min … max):   38.901 s … 40.232 s    4 runs

Benchmark 1: /Users/maciektr/Projects/scarb/target/release/scarb build -w
  Time (mean ± σ):     30.995 s ±  0.478 s    [User: 54.296 s, System: 3.481 s]
  Range (min … max):   30.620 s … 31.653 s    4 runs
  ```
  After:
  ```
  Benchmark 1: /Users/maciektr/Projects/scarb/target/release/scarb build -w --test
  Time (mean ± σ):     727.6 ms ±   7.8 ms    [User: 596.9 ms, System: 70.8 ms]
  Range (min … max):   716.1 ms … 733.4 ms    4 runs

Benchmark 1: /Users/maciektr/Projects/scarb/target/release/scarb build -w
  Time (mean ± σ):     402.4 ms ±   5.4 ms    [User: 218.1 ms, System: 31.5 ms]
  Range (min … max):   397.2 ms … 409.4 ms    4 runs

Benchmark 1: /Users/maciektr/Projects/scarb/target/release/scarb build -w --test
  Time (mean ± σ):     36.060 s ±  0.223 s    [User: 57.429 s, System: 4.008 s]
  Range (min … max):   35.881 s … 36.367 s    4 runs

Benchmark 1: /Users/maciektr/Projects/scarb/target/release/scarb build -w
  Time (mean ± σ):     18.767 s ±  0.670 s    [User: 28.766 s, System: 1.859 s]
  Range (min … max):   18.073 s … 19.666 s    4 runs
  ```

**Review by FroyaTheHen — APPROVED** [review]
lgtm 🍾

File: scarb/src/compiler/incremental/compilation.rs [hunk 1]
```diff
@@ -120,6 +170,8 @@ pub fn load_incremental_artifacts(
     Ok(IncrementalContext::Enabled {
         fingerprints,
         cached_crates,
+        cached_crates_with_warnings,
+        warnings_found: AtomicBool::new(false),
```
DelevoXDG, line 197: Can you clarify why are we setting this to false regardless of `has_warnings`? 🤔

scarb/src/compiler/incremental/compilation.rs
```diff
@@ -193,9 +247,12 @@ pub fn save_incremental_artifacts(
     ctx: IncrementalContext,
     ws: &Workspace<'_>,
 ) -> Result<()> {
+    let warnings_found = ctx.warnings_found();
     let IncrementalContext::Enabled {
         fingerprints,
         cached_crates: _,
+        cached_crates_with_warnings: _,
+        warnings_found: _,
     } = ctx
```
DelevoXDG: Nit: maybe add `fingerprints -> Option<UnitFingerprint>` helper, considering we have helpers for everything else and this takes plenty of space? 😄

DelevoXDG (changes requested):

scarb/src/compiler/incremental/compilation.rs
maciektr: The `has_warnings` tells us, than in the **previous** compilation there has been some warnings, so we need to calculate diags for the main crates.

The `warnings_found` is only set by the `report_warnings` method, to tell us, than in **this** compilation, there has been some warnings.

So we don't really care whether `has_warnings` was true or false - this only changes for which crates we calculate diags. If in previous compilation, main crates had some warnings - we will calculate warnings for them here as well, so `warnings_found` will be set accordingly. If not, we will not find any this time either.

scarb/src/compiler/incremental/compilation.rs
maciektr: Changed.

scarb/src/compiler/incremental/compilation.rs
DelevoXDG: I see, makes sense

PR #2667, opened 2025-09-24
Remove forgotten dbg!

mkaput:


maciektr (changes requested):
Please add #![deny(clippy::dbg_macro)] so we catch this in future

Issue #2624, opened 2025-09-11
Implement assets design doc


mkaput:
Research and implement a way to bundle generated wasm blobs into scarb packages and then probably compiled artifacts, so that they can be reliaby referenced by oracle hint service; even if executed outside scarb context!

PR #2668, opened 2025-09-24
Disable WASM oracle protocol until it's fully implemented

mkaput:

PR #2671, opened 2025-09-24
Implement assets mechanism on the oracle hint service side

mkaput:
cc #2624

### DIFF
--- a/utils/scarb-oracle-hint-service/src/assets.rs
+++ b/utils/scarb-oracle-hint-service/src/assets.rs
@@ -1,6 +1,7 @@
-use anyhow::{Result, anyhow};
+use anyhow::{Result, bail};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use tracing::{debug, trace};
 
 #[derive(Clone, Debug)]
 pub struct Assets {
@@ -33,10 +34,26 @@ impl Assets {
     ///
     /// If the asset is not found, this function returns an error.
     pub fn fetch(&self, name: &str) -> Result<PathBuf> {
-        self.search_paths
-            .iter()
-            .map(|p| p.join(name))
-            .find(|p| p.exists())
-            .ok_or_else(|| anyhow!("asset not found: {name}"))
+        debug!(
+            name,
+            search_paths = ?(
+                self.search_paths
+                    .iter()
+                    .map(|p|p.display())
+                    .collect::<Vec<_>>()
+            ),
+            "looking for asset",
+        );
+        for p in self.search_paths.iter() {
+            let path = p.join(name);
+            let exists = path.exists();
+            trace!(path=?path.display(), exists=?exists, "trying");
+            if exists {
+                debug!(path=?path.display(), "found");
+                return Ok(path);
+            }
+        }
+        debug!(name, "asset not found");
+        bail!("asset not found: {name}")
     }
 }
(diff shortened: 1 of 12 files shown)


mkaput:

- #2578
    - #2579
        - #2580
            - #2581
                - #2582
                    - #2619
                        - #2631
                            - #2632
                                - #2633
                                    - #2634
                                        - #2635
                                            - #2667
                                                - #2668
                                                    - #2671 ◀

utils/scarb-oracle-hint-service/src/assets.rs
```diff
@@ -0,0 +1,59 @@
...
+            search_paths = ?(
+                self.search_paths
+                    .iter()
+                    .map(|p|p.display())
+                    .collect::<Vec<_>>()
+            ),
+            "looking for asset",
+        );
+        for p in self.search_paths.iter() {
+            let path = p.join(name);
```
maciektr: Do we allow this names to point outside the search path? 🤔 E.g. `../../../smth.wasm`?

utils/scarb-oracle-hint-service/src/hint_service.rs
```diff
@@ -19,26 +23,32 @@ enum OracleCheatcodeSelectorInner {
...
+    /// executor.
+    ///
+    /// ## Panics
+    /// This function will panic if the executable path is not a file.
+    pub fn new(executable_path: Option<&Path>) -> Self {
+        Self {
+            connections: Connections::default(),
+            protocols: builtin_protocols(),
+            assets: executable_path
+                .map(Assets::for_executable)
```
maciektr, line 41: Isn't this too restrictive from the perspective of the executor? 🤔 
The dd said: 

> 1. Let OracleHintService be configurable on where it allows protocols to look for assets. Basically establish a PATH-like mechanism:
>     1. Have builder-like method for adding custom paths.
>         1. Executors will be obliged use this to pass path to a directory where executed sierra/casm is located. This mechanism will allow loading assets that reside next to sierra/casm files. It can only be a convention (enforced by public shaming).

Which suggests to me that we should not limit this interface to executable location only 🤔

maciektr (commented):
One question attached.

utils/scarb-oracle-hint-service/src/hint_service.rs
mkaput: It's true that the design doc asked for more general interface, but for current purposes the limited interface as this one is enough and I thought it might be good to just wait for feedback if we need to expand it. Note that this interface is a logical subset of the one in the DD

utils/scarb-oracle-hint-service/src/assets.rs
mkaput: 16ef28cbe95354ee6797042ce310524eb4dda145

utils/scarb-oracle-hint-service/src/assets.rs
```diff
@@ -0,0 +1,118 @@
...
+        }
+        debug!(name, "asset not found");
+        bail!("asset not found: {name}")
+    }
+}
+
+/// Normalize a path, including `..` without traversing the filesystem.
+/// Returns an error if normalization would leave leading `..` components.
+/// This is unstable in Rust's std.
+pub fn normalize_lexically(path: &Path) -> Result<PathBuf> {
```
maciektr, line 65: Can't we normalize as we would do normally and then ensure it's under the base path by running https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.strip_prefix and checking it for error? 🤔

mkaput: you mean ours `fsx::canonicalize`? it accesses the filesystem which I think is better to avoid. and `std::path::absolute` doesn't handle `..`

maciektr (approved):

PR #2673, opened 2025-09-25
Add `assets` field to Scarb.toml

mkaput:
cc #2624


mkaput:

- #2673 ◀
    - #2671

scarb/src/ops/compile.rs
```diff
@@ -268,6 +273,10 @@ fn compile_unit_inner(unit: CompilationUnit, ws: &Workspace<'_>) -> Result<()> {
                     offloader.join()?;
                 }
 
+                if result.is_ok() {
+                    copy_assets(assets, &target_dir)?;
+                }
```
maciektr, line 278: So, this introduces an unstated assumption, that assets that collide with compilation artifacts by name will overwrite the latter. 

While I am rather okay with that, I wonder if it would make sense to introduce some validation on the file names instead. 

This would obviously force us to require structs that implement the [`Compiler` trait](https://github.com/software-mansion/scarb/blob/5f5c857b4a3ac9f86c79d32cfa35c5b59435c49c/scarb/src/compiler/mod.rs#L21) to declare files they produce - but it might be usefull in other cases anyway. 

For instance we could use this info to calculate hashes of compilation artifacts + compilation unit fingerprints and skip entire compilation if the inputs have not changed and all the outputs are up to date 🤔 

Any opinions?

mkaput: this makes sense but is a topic for separate issue imho: #2677

maciektr: Ok

PR #2676, opened 2025-09-26
Allow compiling executable to Sierra

maciektr:


scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -449,7 +449,14 @@ pub struct TomlExecutableTargetParams {
...
     /// In general, syscalls are not allowed in executables, as they are currently not checked.
     pub allow_syscalls: Option<bool>,
+    /// Program entrypoint.
+    ///
+    /// Executable must have exactly one entrypoint that will be used during the execution.
+    /// This must be a function with the `#[executable]` attribute.
     pub function: Option<String>,
+    /// Emit a compiled Sierra program for the executable function, in JSON format.
+    /// This may be useful, for example, for debugging or local profiling.
+    pub sierra: Option<bool>,
```
THenry14, line 459: Should this be visible in package under targets, when set like this:
```
[executable]
sierra = true
```
? 

I'd presume so, but it seems it is not 🤔 (it is, however, when set under `[[target.executable]]`)

maciektr: Hmm, are you sure it does not?
It works for me 🤔 
Added a test case here: https://github.com/software-mansion/scarb/blob/9dd940e26de2c4722efc70b0c51bfe339e034c35/scarb/tests/build_targets.rs#L1891-L1904

THenry14: I think so, yes. Although I do not have the target defined, I only have this:

```
[package]
name = "simple_nameless"
version = "0.1.0"
edition = "2024_07"

[dependencies]
cairo_execute = "2.12"

[executable]
sierra = true

[cairo]
enable-gas = false
```

then I get:
```
[extensions/scarb-execute/src/lib.rs:265:9] &package.targets = [
    TargetMetadata {
        kind: "executable",
        name: "simple_nameless",
        source_path: "/(...)/lib.cairo",
        params: Object {},
        extra: {},
    },
    TargetMetadata {
        kind: "test",
        name: "simple_nameless_unittest",
        source_path: "/(...)/lib.cairo",
        params: Object {
            "build-external-contracts": Array [],
            "test-type": String("unit"),
        },
        extra: {},
    },
]
```

But, when I set the target:

```
$ git diff Scarb.toml
diff --git a/simple_nameless/Scarb.toml
index d9f1159..0c79402 100644
--- a/simple_nameless/Scarb.toml
+++ b/simple_nameless/Scarb.toml
@@ -6,7 +6,7 @@ edition = "2024_07"
 [dependencies]
 cairo_execute = "2.12"

-[executable]
+[[target.executable]]
 sierra = true

 [cairo]
```

All works as expected.

maciektr: @THenry14 
I've tested your example, and sierra has been produced. We have tests for outputing sierra for both `executable` and `target.executable` notation. I added a test which asserts that the `sierra` flag is shown in  metadata as well. 
Since this is tested, I'll close this conversation. I'll compile a dev build, so you can test it out this way.

THenry14: Yes, sierra is produced for me as well, it's just missing from the package metadata (as shown above), not sure if this is intended, that's why I have approved this PR with a comment.

I can see all your tests also define [[target.executable]], which was missing from my repro. Can you confirm you tried without setting the target explicitly? 🙏

maciektr: Hmm, this test passes 🤔 

```
#[test]
fn executable_target_can_compile_to_sierra() {
    ProjectBuilder::start()
        .name("executable_test")
        .dep_cairo_execute()
        .manifest_extra(indoc! {r#"
            [executable]
            sierra = true

        "#})
        .lib_cairo(indoc! {r#"
            #[executable]
            fn first() -> felt252 {
                42
            }
        "#})
        .build(&t);
    let meta = Scarb::quick_snapbox()
        .arg("--json")
        .arg("metadata")
        .arg("--format-version")
        .arg("1")
        .current_dir(&t)
        .stdout_json::<Metadata>();
    let pkg = meta
        .packages
        .iter()
        .find(|p| p.name == "executable_test")
        .unwrap();
    let targets = pkg
        .targets
        .iter()
        .filter(|t| t.kind == "executable")
        .collect::<Vec<_>>();
    assert_eq!(targets.len(), 1);
    for t in targets {
        assert!(
            t.params
                .as_object()
                .unwrap()
                .get("sierra")
                .unwrap()
                .as_bool()
                .unwrap(),
        );
    }
}
```
I'll ping you tomorrow with a nightly and ask you to try on that 🤔

PR #2717, opened 2025-10-06
Share cache and incremental dirs in-process in tests

### LINKED ISSUE #2690: Share target and cache dirs in-process in tests
A lot of our test time is spent on recompiling `core` package whenever we compile some Cairo code because we run tests in target/cache dir isolation. Let's optimize this by leveraging our newly introduced incremental compilation mechanics.

In `ProjectBuilder`, if scarb cache/target dir is not explicitly overridden, let's "inject" a pregenerated one. The pregenerated directories should be a static `OnceLock<(PathBuf, PathBuf)>` that are generated by creating a basic scarb package (it must have a unique name, like `this_is_a_cache_warmup`) that is then compiled with `scarb build`. This project can be created and compiled in a tempdir, just make sure this directory is not deallocated before test suite halts (so keep it in another `static OnceLock`). The injection should be just copying entire contents of pregenerated dirs into writable target/cache dirs.

mkaput:
fix #2690

### DIFF
--- a/utils/scarb-test-support/src/command.rs
+++ b/utils/scarb-test-support/src/command.rs
@@ -5,11 +5,11 @@ use assert_fs::prelude::*;
 use serde::de::DeserializeOwned;
 use snapbox::cmd::Command as SnapboxCommand;
 use std::ffi::OsString;
-use std::fs;
 use std::io::BufRead;
 use std::path::{Path, PathBuf};
 use std::process::Command as StdCommand;
 use std::sync::LazyLock;
+use std::{env, fs};
 
 pub struct Scarb {
     cache: EnvPath,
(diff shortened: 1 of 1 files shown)


utils/scarb-test-support/src/command.rs
```diff
@@ -9,37 +12,40 @@ use std::path::{Path, PathBuf};
 use std::process::Command as StdCommand;
 use std::sync::LazyLock;
 
-use crate::cargo::cargo_bin;
-
 pub struct Scarb {
-    cache: EnvPath,
+    cache: Option<EnvPath>,
     config: EnvPath,
```
maciektr: Can we doc that None = shared? Or changed to enum? 🤔

maciektr (approved):

utils/scarb-test-support/src/command.rs
mkaput: fe398a1a58efd1c7266b8f2070f20835c2f949e5

maciektr (approved):

mkaput:
if [6ba7957](https://github.com/software-mansion/scarb/pull/2717/commits/6ba79577deb4bb685e154197cd89e3b3cdee6364) doesn't help then let's close this PR

PR #2721, opened 2025-10-08
Do not query registry to build resolve graph

maciektr:


maciektr:

- #2721 ◀
    - #2722

scarb/src/resolver/mod.rs
```diff
@@ -640,45 +659,6 @@ mod tests {
         );
     }
 
-    #[test]
-    fn can_resolve_target_kind_dep() {
```
maciektr, line 644: We no longer build a resolution graph at this point in execution.

mkaput (approved):

PR #2732, opened 2025-10-14
Update `reqwest` to 0.12

mkaput:


maciektr (approved):

Cargo.toml
```diff
@@ -115,7 +115,7 @@ quote = "1"
 ra_ap_toolchain = "0.0.218"
 rayon = "1.11.0"
 redb = "2.6.3"
-reqwest = { version = "0.11", features = ["gzip", "brotli", "deflate", "json", "stream", "multipart"], default-features = false }
+reqwest = { version = "0.12", features = ["gzip", "brotli", "deflate", "json", "stream", "multipart"] }
```
maciektr, line 118: Why do we remove `default-features = false`? 🤔

mkaput: Because it had virtually no positive effect in this particular case - especially, this disabled HTTP2 support which was contradictory to our design decisions 😃

maciektr (approved):

PR #2734, opened 2025-10-15
Update wasm.md to remove future capabilities note

mkaput:
Removed note about future limitations on capabilities as we're going to omit this feature.


mkaput:
merged without reviewer because this is nobrainer

PR #2749, opened 2025-10-16
Refactor of scarb lint

wawel37:


scarb/src/ops/lint/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+    let mut packages_with_error: Vec<PackageName> = Default::default();
+
+    for (package, compilation_unit) in compilation_units_for_linting.iter() {
+        ws.config()
+            .ui()
+            .print(Status::new("Linting", &compilation_unit.name()));
+
+        let db = build_lint_database(compilation_unit, ws)?;
+        let diagnostics = get_linting_diagnostics(&db, package, &absolute_path, compilation_unit)?;
+        display_diagnostics(ws, &db, &diagnostics);
```
FroyaTheHen: this is going to display warnings even if `--no-warnings` flag is passed, is it on purpose (and if so what's the reasoning behind)?

scarb/src/ops/lint/mod.rs
```diff
@@ -0,0 +1,395 @@
...
+                }
+            }
+        }
+    }
+}
+
+/// Returns an error message indicating which packages failed linting due to previous errors.
+fn get_package_linting_error_message(packages_with_error: &[PackageName]) -> String {
+    if packages_with_error.is_empty() {
+        "".to_string()
```
maciektr, line 207: NIT: this function could return Result directly, you could then only do `check_linting_errors()?;` in the caller.

scarb/src/ops/lint/db.rs
```diff
@@ -0,0 +1,178 @@
...
+use cairo_lang_semantic::{
+    db::{SemanticGroup, semantic_group_input},
+    ids::AnalyzerPluginLongId,
+    plugin::PluginSuite,
+};
+use cairo_lint::LinterAnalysisDatabase;
+use salsa::Setter;
+use std::{collections::HashMap, sync::Arc};
+
+/// Keep it in sync with [crate::compiler::db::build_scarb_root_database].
```
FroyaTheHen, line 31: I think it's a perfect time to come back to the question of maintainability of those functions. 
see https://github.com/software-mansion/scarb/pull/2542#pullrequestreview-3135421836
cc @mkaput @piotmag769 are we there yet?

mkaput: yes. we are currently on latest stable salsa everywhere

scarb/src/ops/lint/mod.rs
wawel37: I thought about it previously, but it would make this function a little but counter intuitive, as if this function would be called that way, it would **always** return an error. I think it's better to decide upon throwing an error in a main function, and let this one decide upon just the error message. Wdyt @maciektr ?

maciektr: Whathever you prefer

PR #2766, opened 2025-10-20
Support module-level inline macro

### LINKED ISSUE #2532: Support  inline proc macros at module top level items use
It should be possible to use inline proc macros at top level of the module, eg 

'''
mod name { macro!() }
'''

DelevoXDG:
Closes #2532

### DIFF
--- a/scarb/tests/proc_macro_v2_expand.rs
+++ b/scarb/tests/proc_macro_v2_expand.rs
@@ -2615,10 +2615,10 @@ fn module_level_inline_macro_with_args() {
     let temp = TempDir::new().unwrap();
     let t = temp.child("some");
     CairoPluginProjectBuilder::default()
-            .add_cairo_lang_parser_dep()
-            .add_cairo_lang_syntax_dep()
-            .add_primitive_token_dep()
-            .lib_rs(indoc! {r##"
+        .add_cairo_lang_parser_dep()
+        .add_cairo_lang_syntax_dep()
+        .add_primitive_token_dep()
+        .lib_rs(indoc! {r##"
             use cairo_lang_macro::{ProcMacroResult, TokenStream, inline_macro, quote};
             use cairo_lang_parser::utils::SimpleParserDatabase;
             use cairo_lang_syntax::node::with_db::SyntaxNodeWithDb;
@@ -2637,8 +2637,8 @@ fn module_level_inline_macro_with_args() {
                 result
             }
             "##})
-            .add_dep(r#"indoc = "*""#)
-            .build(&t);
+        .add_dep(r#"indoc = "*""#)
+        .build(&t);
 
     let project = temp.child("hello");
     ProjectBuilder::start()


scarb/tests/proc_macro_v2_expand.rs
maciektr, line 1: I miss a test like this one: https://github.com/software-mansion/scarb/blob/3084ae340d54c0182d6573abee1d9f3c506dfd5d/scarb/tests/proc_macro_v2_expand.rs#L1285
where args would be parsed with the cairo parser

scarb/tests/proc_macro_v2_expand.rs
```diff
@@ -2540,3 +2540,474 @@ fn can_use_two_derive_macros() {
...
+            [..]Finished `dev` profile target(s) in [..]
+            [..]Executing hello
+            Program output:
+            100
+            Saving output to: target/execute/hello/execution1
+        "#});
+}
+
+#[test]
+fn module_level_inline_macro_in_nested_module() {
```
maciektr: I would say that 
```
        .lib_cairo(indoc! {r#"
            mod hello {
                some!();
            }

            #[executable]
            fn main() -> felt252 {
                hello::foo()
            }
```
Is also a nested module test.
Can we call it directly in the module tree root, i.e.
```
        .lib_cairo(indoc! {r#"
            some!();

            #[executable]
            fn main() -> felt252 {
                hello::foo()
            }
```
? 🤔

scarb/tests/proc_macro_v2_expand.rs
```diff
@@ -2540,3 +2540,474 @@ fn can_use_two_derive_macros() {
...
+
+            // The first token is the opening parenthesis "(", so take the next token
+            let TokenTree::Ident(arg) = token_stream
+                .tokens
+                .get(1)
+                .expect("macro expected at least one argument");
+
+            let content = arg.content.as_ref();
+            let slice_len: usize = core::cmp::min(7, content.len());
+            let slice = &content[..slice_len];
```
maciektr: I don't get why do you need this 🤔

scarb/tests/proc_macro_v2_expand.rs
DelevoXDG: Changed

scarb/tests/proc_macro_v2_expand.rs
DelevoXDG: > Is also a nested module test.

Valid point! And yes, we can call it directly at module tree root. I've changed the test to check that instead, since testing another level of nesting is unnecessary

scarb/tests/proc_macro_v2_expand.rs
DelevoXDG: It seemed to me that test would be easier to understand if emitted diagnostic mapped to args in the original token stream more clearly, but perhaps I was mistaken. Updated to be more like `code_mappings_preserve_inline_macro_error_locations`

PR #2770, opened 2025-10-21
Remove cairo-test from archive verify step

maciektr:
Failure example: https://github.com/software-mansion/scarb/actions/runs/18682848804/job/53273349432


mkaput (changes requested):
I suggest merging #2771 because I'm already running dev build from it and it has more elaborative description

Issue #2654, opened 2025-09-22
Make it possible to create a documentation for workspace packages with common SUMMARY.md as a root


FroyaTheHen:
Now when running `scarb doc --workspace` it creates separate books for each package. Make it possible to generate documentation as one book for all workspace packages. 

To consider: 
Either change the current setup to work as such always when flag `--workspace` supplied or add a possibility to customise the behaviour (for example by additional flag). 

see Alexandria as an example use-case.

PR #2780, opened 2025-10-23
One mdbook for workspace packages

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2654


extensions/scarb-doc/.DS_Store
maciektr, line 1: pls remove

utils/scarb-ui/src/args/packages_filter.rs
```diff
@@ -36,7 +36,7 @@ pub struct PackagesFilter {
     package: Vec<String>,
     /// Run for all packages in the workspace.
     #[arg(short, long, conflicts_with = "package")]
-    workspace: bool,
+    pub workspace: bool,
```
maciektr: I'd prefer a getter

extensions/scarb-doc/src/docs_generation/markdown.rs
```diff
@@ -64,8 +65,32 @@ impl MarkdownContent {
         })
     }
 
+    pub fn from_workspace(packages: &[PackageInformation]) -> Result<Self> {
+        // TODO: consider generating book.toml content from workspace metadata
```
maciektr: pls add issue number

extensions/scarb-doc/src/main.rs
```diff
@@ -194,6 +233,55 @@ fn output_markdown(
     Ok(())
 }
 
+fn workspace_output_markdown(
```
maciektr: Overall this seems to mostly duplicate `output_markdown`. Can this be shared (as it seems to mostly differ by accepting single  `PackageInformation` vs an array of them), or at least be more composable? 🤔

extensions/scarb-doc/src/main.rs
```diff
@@ -35,24 +37,34 @@ fn main_inner(args: Args, ui: Ui) -> Result<()> {
...
     let workspace_root = metadata.workspace.root.clone();
     let mut output = match args.output_format {
         OutputFormat::Json => OutputEmit::for_json(output_dir, workspace_root, ui.clone()),
         OutputFormat::Markdown => {
             OutputEmit::for_markdown(output_dir, workspace_root, args.build, ui.clone())
         }
     };
+    let contexts: Vec<PackageContext> = metadata_for_packages
+        .iter()
+        .map(|pm| generate_package_context(&metadata, pm, args.document_private_items))
```
maciektr: This change is actually quite dangerous, as package context owns the compiler db. https://github.com/software-mansion/scarb/blob/efc5efe689ceef486575e46b08fe491ec6bde808/extensions/scarb-doc/src/lib.rs#L51C10-L51C11
So creating and using multiple contexts at the same time may lead to drastically more memory usage.

maciektr: Can we refactor this, so we only keep built output format in memory? 🤔

FroyaTheHen: Good call, thanks, Maciej! I'll change it.

extensions/scarb-doc/src/main.rs
FroyaTheHen: definitely can be improved, will do 🙌

PR #2787, opened 2025-10-27
Implement unit artifacts fingferprint

maciektr:
Implement unit artifacts fingferprint

Register compilation artifacts with unit artifacts fingerprint

### DIFF
--- a/scarb/src/compiler/incremental/artifacts_fingerprint.rs
+++ b/scarb/src/compiler/incremental/artifacts_fingerprint.rs
@@ -24,10 +24,13 @@ use std::io::BufReader;
 ///  was bound to change, we would invalidate the fingerprint due to the first part anyway.
 #[derive(Serialize, Deserialize)]
 pub struct UnitArtifactsFingerprint {
+    /// A hash of the `UnitComponentsFingerprint` corresponding to the compilation unit input.
     #[serde(skip)]
     pub unit: u64,
+    /// A hash of the Scarb targets of the compilation unit.
     #[serde(skip)]
     pub target: u64,
+    /// A list of `LocalFingerprint`s of compilation artifacts.
     pub local: Vec<LocalFingerprint>,
 }
 
@@ -66,19 +69,15 @@ pub fn save_unit_artifacts_fingerprint(
 ) -> anyhow::Result<()> {
     let digest = fingerprint.digest();
     let fingerprint_dir = unit.fingerprint_dir(ws);
-    let target_name = unit.main_component().target_name();
-    let target_kind = unit.main_component().target_kind();
-    let unit_id = short_hash((target_kind, target_name.clone()));
-    let filename = format!("unit-{target_name}-{unit_id}");
     write_string(
-        &filename,
+        &unit.cache_filename(),
         "unit artifacts fingerprint digest file",
         &fingerprint_dir,
         ws,
         digest,
     )?;
     write_json(
-        &format!("{filename}.json"),
+        &unit.fingerprint_filename(),
         "unit artifacts fingerprint file",
         &fingerprint_dir,
         ws,
@@ -93,10 +92,7 @@ pub fn load_unit_artifacts_local_paths(
     ws: &Workspace<'_>,
 ) -> anyhow::Result<Option<Vec<LocalFingerprint>>> {
     let fingerprint_dir = unit.fingerprint_dir(ws);
-    let target_name = unit.main_component().target_name();
-    let target_kind = unit.main_component().target_kind();
-    let unit_id = short_hash((target_kind, target_name.clone()));
-    let filename = format!("unit-{target_name}-{unit_id}.json");
+    let filename = unit.fingerprint_filename();
     if !fingerprint_dir.path_unchecked().join(&filename).exists() {
         return Ok(None);
     }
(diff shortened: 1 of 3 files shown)


scarb/src/compiler/incremental/artifacts_fingerprint.rs
```diff
@@ -0,0 +1,140 @@
...
+use scarb_stable_hash::{StableHasher, short_hash, u64_hash};
+use serde::{Deserialize, Serialize};
+use std::hash::{Hash, Hasher};
+use std::io::BufReader;
+
+/// Fingerprint of compiled unit artifacts.
+///
+/// This can be seen as a fingerprint that captures the relation between compilation inputs
+/// (as defined via `UnitComponentsFingerprint`) and outputs (i.e., targets of the main component
+/// of the [`CairoCompilationUnit`] and `LocalFingerprint`s of compilation artifacts).
```
DelevoXDG:
```suggestion
/// (as defined via [`UnitComponentsFingerprint`]) and outputs (i.e., targets of the main component
/// of the [`CairoCompilationUnit`] and [`LocalFingerprint`]s of compilation artifacts).
```

scarb/src/compiler/incremental/artifacts_fingerprint.rs
```diff
@@ -0,0 +1,140 @@
...
+/// knowing the artifacts' locations beforehand. This works on the assumption that produced artifacts
+/// are deterministic in relation to the compilation inputs and targets, thus if the artifact list
+///  was bound to change, we would invalidate the fingerprint due to the first part anyway.
+#[derive(Serialize, Deserialize)]
+pub struct UnitArtifactsFingerprint {
+    #[serde(skip)]
+    pub unit: u64,
+    #[serde(skip)]
+    pub target: u64,
+    pub local: Vec<LocalFingerprint>,
```
DelevoXDG, line 34: I think it may be useful to add description to these fields just so it's more immediately clear what these refer to

scarb/src/ops/compile.rs
```diff
@@ -300,6 +262,78 @@ fn compile_unit_inner(unit: CompilationUnit, ws: &Workspace<'_>) -> Result<()> {
...
+    let assets = collect_assets(&unit)?;
+
+    // This scope limits the offloader lifetime.
+    thread::scope(|s| {
+        let offloader = Offloader::new(s, ws);
+        let target_dir = unit.target_dir(ws);
+
+        let ctx = load_incremental_artifacts(&unit, &mut db, ws)?;
+
+        let is_fresh_unit_artifacts = ctx.cached_crates_with_warnings().is_empty()
```
FroyaTheHen: Wouldn't it be more efficient to separately check if `ctx.cached_crates_with_warnings().is_empty()` and only reprint the cached warnings `if is_fresh_unit_artifacts & !ctx.cached_crates_with_warnings().is_empty()`? please consider

scarb/src/compiler/incremental/artifacts_fingerprint.rs
```diff
@@ -0,0 +1,140 @@
...
+    unit: &CairoCompilationUnit,
+    fingerprint: UnitArtifactsFingerprint,
+    ws: &Workspace<'_>,
+) -> anyhow::Result<()> {
+    let digest = fingerprint.digest();
+    let fingerprint_dir = unit.fingerprint_dir(ws);
+    let target_name = unit.main_component().target_name();
+    let target_kind = unit.main_component().target_kind();
+    let unit_id = short_hash((target_kind, target_name.clone()));
+    let filename = format!("unit-{target_name}-{unit_id}");
```
DelevoXDG: Should we add helper method for this `filename`, similarly to `IncrementalArtifactsProvider` helpers?

Would both reduce duplication (the intermediate variables don't appear to be used in this context), and avoid potential mismatch between `save_unit_artifacts_fingerprint`, `load_unit_artifacts_local_paths`, etc.

scarb/src/ops/compile.rs
maciektr: So the thing is, we don't really keep the warnings in cache - just a single boolean flag that tells you, whether there are any warnings. To show the warnings to the user, you need to actually recompile the crate. (for both unit artifacts and normal incremental compilation)
Caching the warnings value is something to consider, but out of scope of this PR I think.

DelevoXDG: Should we avoid recompilation on `--no-warnings` / `--quiet` verbosity (check `ws.config().ui().verbosity().should_print_warnings()`)?

maciektr: Good idea, done.

scarb/src/compiler/incremental/artifacts_fingerprint.rs
maciektr: Good idea, done.

scarb/src/compiler/incremental/artifacts_fingerprint.rs
maciektr: @DelevoXDG turns out I cannot link to them in docs, as these structs are private. See: https://doc.rust-lang.org/rustdoc/lints.html#private_intra_doc_links

PR #2806, opened 2025-10-31
Add `quote_format!` macro with non-Rust Cairo syntax support

### LINKED ISSUE #2290: Support non-rust syntax in cairo-lang-macro quote! macro
Non-rust syntax in quote macro 
```
quote! {
      assert(1 + 1 == 2, 'fail')
}
```
Results in 
````
        2 + error: character literal may only contain one codepoint
        3 +   --> src/lib.rs:16:32
        4 +    |
        5 + 16 |             assert(1 + 1 == 2, 'fail')
        6 +    |                                ^^^^^^
        7 +    |
        8 + help: if you meant to write a string literal, use double quotes
        9 +    |
       10 + 16 -             assert(1 + 1 == 2, 'fail')
       11 + 16 +             assert(1 + 1 == 2, "fail")
       12 +    |
       13 + error: could not compile `some` (lib) due to 1 previous error
       14 + error: process did not exit successfully: exit status: 101
       15 + error: could not compile `some` due to previous error
```

DelevoXDG:
- #2806 
- #2812
___
Closes #2290

### DIFF
--- a/plugins/cairo-lang-quote/Cargo.toml
+++ b/plugins/cairo-lang-quote/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cairo-lang-quote"
-version = "0.1.1"
+version = "0.2.0"
 edition = "2021"
 
 authors.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1025,7 +1025,7 @@ dependencies = [
  "cairo-lang-macro-attributes 0.2.1",
  "cairo-lang-macro-stable 2.0.0",
  "cairo-lang-primitive-token",
- "cairo-lang-quote 0.1.1",
+ "cairo-lang-quote 0.2.0",
  "linkme",
  "serde",
  "serde_json",
@@ -1171,7 +1171,7 @@ dependencies = [
 
 [[package]]
 name = "cairo-lang-quote"
-version = "0.1.1"
+version = "0.2.0"
 dependencies = [
  "proc-macro2",
  "quote",
--- a/plugins/cairo-lang-macro/Cargo.toml
+++ b/plugins/cairo-lang-macro/Cargo.toml
@@ -18,7 +18,7 @@ bumpalo.workspace = true
 cairo-lang-macro-attributes = { path = "../cairo-lang-macro-attributes", version = "0.2.1" }
 cairo-lang-macro-stable = { path = "../cairo-lang-macro-stable", version = "2.0.0" }
 cairo-lang-primitive-token.workspace = true
-cairo-lang-quote = { path = "../cairo-lang-quote", version = "0.1.0" }
+cairo-lang-quote = { path = "../cairo-lang-quote", version = "0.2.0" }
 linkme.workspace = true
 serde = { workspace = true, optional = true }
 


utils/scarb-test-support/Cargo.toml
```diff
@@ -7,6 +7,7 @@ publish = false
 [features]
 default = []
 scarb-config = ["dep:scarb"]
+proc-macro-server-types = ["dep:scarb-proc-macro-server-types"]
```
maciektr:
```suggestion
proc-macro-server = ["dep:scarb-proc-macro-server-types"]
```

scarb/tests/proc_macro_v2_quote.rs
maciektr, line 1: I think we need to have: 
- a test case that demonstrates that Cairo syntax not compatible with Rust syntax can be used
- code blocks delimited with curly braces `{`, `}` ; e.g. a whole function, can be returned

plugins/cairo-lang-quote/Cargo.toml
```diff
@@ -15,5 +15,9 @@ repository.workspace = true
 proc-macro = true
 
 [dependencies]
```
maciektr, line 17: Just to be sure, we can hide `quote_format` behind a feature flag, so it's possible to slim down the cairo-lang-macro crate unless you need this - as this crate is compiled for each run of each proc macro.

plugins/cairo-lang-quote/src/lib.rs
```diff
@@ -127,6 +135,150 @@ pub fn quote(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+            }
+            Piece::NextArgument(arg) => {
+                let expr = match arg.position {
+                    Position::ArgumentIs(idx) | Position::ArgumentImplicitlyIs(idx) => {
+                        if let Some(expr) = args.get(idx).copied() {
+                            expr
+                        } else {
+                            return Error::new(
+                                fmtstr.span(),
+                                formatdoc! (r#"format arg index {} is out of range (the format string contains {} args)."#,
```
maciektr: Can we just use `format!`? The less deps the better here.

plugins/cairo-lang-quote/Cargo.toml
```diff
@@ -15,5 +15,9 @@ repository.workspace = true
...
+cairo-lang-parser.workspace = true
+ra-ap-rustc_parse_format.workspace = true
+indoc.workspace = true
 proc-macro2.workspace = true
 quote.workspace = true
+syn = { workspace = true, features = ["full"] }
```
maciektr: Do we need all features here? 🤔

scarb/tests/proc_macro_v2_quote.rs
DelevoXDG: > code blocks delimited with curly braces `{`, `}` ; e.g. a whole function, can be returned

Added `quote_format_macro_with_code_block`

> a test case that demonstrates that Cairo syntax not compatible with Rust syntax can be used

We already check this in `quote_format_macro` test, where we use non-Rust syntax (`'fail'`):
```rs
                let tokens = quote_format! {
                    r#"assert(1 + 1 == {}, 'fail')"#,
                    x
                };
```

plugins/cairo-lang-quote/Cargo.toml
DelevoXDG: No, good catch! I changed it so only `"parsing"` is enabled

plugins/cairo-lang-quote/Cargo.toml
DelevoXDG: Should we though, now that I have removed `cairo-lang-parser` and `indoc` deps?

Also, do you want to hide only `quote_format!` behind feature flag, or `quote!` as well?

If it's the former, I feel like it would be confusing since we'd be increasing visibility of macro that doesn't work in all cases, and reducing said visibility for the macro that does cover them 🤔

utils/scarb-test-support/Cargo.toml
DelevoXDG: Since we eliminated the dependency that lead to the error, do you think it's worth to keep it just in case, or remove it for now?

plugins/cairo-lang-quote/Cargo.toml
maciektr: It's fine now when we slimmed it down. I don't think we need the flag.

utils/scarb-test-support/Cargo.toml
maciektr: TBH I don't think it matters either way

scarb/tests/proc_macro_v2_quote.rs
maciektr: Cool!

plugins/cairo-lang-quote/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cairo-lang-quote"
-version = "0.1.0"
+version = "0.1.1"
```
maciektr:
```suggestion
version = "0.2.0"
```
We can do 0.2 here to underline this is a new feature and then 0.2.2 in cairo-lang-macro to keep semver compatibility in the interface.

plugins/cairo-lang-quote/src/lib.rs
```diff
@@ -127,6 +131,163 @@ pub fn quote(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+}
+
+/// Basic tokenizer for `quote_format!` macro.
+///
+/// Intentionally simplified to avoid full parsing of Cairo syntax.
+/// Only splits strings into tokens and preserves whitespace.
+/// Token kinds and spans are ignored as spans always set to call site.
+/// Additionally, it's expected placeholders (`{}`, `{0}`, etc.) are already
+/// stripped out by the format parser before this function is called.
+fn tokenize_basic(string: &str) -> Vec<QuoteToken> {
```
maciektr, line 161: I feel like for what we need, something as simple as 
```
    string
        .split(" ")
        .map(|s| QuoteToken::Content(s.to_string()))
        .flat_map(|content| [QuoteToken::Whitespace, content])
        .skip(1)
        .collect()
```
Would do the job 🤔

plugins/cairo-lang-quote/src/lib.rs
```diff
@@ -127,6 +131,163 @@ pub fn quote(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+            Piece::Lit(string) => {
+                for token in tokenize_basic(string) {
+                    match token {
+                        QuoteToken::Content(content) => {
+                            output_token_stream.extend(rust_quote! {
+                              quote_macro_result.push_token(::cairo_lang_macro::TokenTree::Ident(::cairo_lang_macro::Token::new(::std::string::ToString::to_string(#content), ::cairo_lang_macro::TextSpan::call_site())));
+                            });
+                        }
+                        // Vars are handled via placeholders, so they should not appear here.
+                        QuoteToken::Var(_) => {}
```
maciektr:
```suggestion
                        QuoteToken::Var(_) => { unreachable!("tokenizer cannot return a var quote token type") }
```

plugins/cairo-lang-quote/src/lib.rs
DelevoXDG: Updated

Issue #2793, opened 2025-10-29
(mvp) add generate .mdx files in scarb doc


FroyaTheHen:
add a possibility to use .mdx instead od .md and expose it (as additional flag for example scarb doc --fromat mdx?) 

(don't expose in documentation)

Issue #2775, opened 2025-10-22
Research mdbook + mintlify


FroyaTheHen:
Ideally we want some automated way of using scarb-doc generated mdbook files with mintlify

PR #2811, opened 2025-11-03
add optional .mdx files generation to scarb doc

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2793


extensions/scarb-doc/src/docs_generation/markdown.rs
```diff
@@ -11,12 +11,55 @@ mod book_toml;
...
+        match format {
+            OutputFormat::Markdown => OutputFilesExtension::Md,
+            OutputFormat::Mdx => OutputFilesExtension::Mdx,
+            OutputFormat::Json => OutputFilesExtension::Json,
+        }
+    }
+}
+
+// Global, run-scoped output extension accessor.
+pub static OUTPUT_EXTENSION: OnceLock<&'static str> = OnceLock::new();
```
DelevoXDG: Why pub?

extensions/scarb-doc/src/main.rs
```diff
@@ -55,7 +58,7 @@ fn main_inner(args: Args, ui: Ui) -> Result<()> {
     } else {
         let mut output = match args.output_format {
             OutputFormat::Json => OutputEmit::for_json(output_dir, workspace_root, ui.clone()),
-            OutputFormat::Markdown => {
+            OutputFormat::Markdown | OutputFormat::Mdx => {
                 OutputEmit::for_markdown(output_dir, workspace_root, args.build, ui.clone())
```
DelevoXDG, line 63: It seems we're just emitting regular markdown with `.mdx` extension - are you sure that's something we can safely do? 

MDX doesn't seem to be truly a superset of MD, as it has some deviations that make a valid MD an invalid MDX:
- https://mdxjs.com/docs/what-is-mdx/#markdown
- https://github.com/micromark/mdx-state-machine?tab=readme-ov-file#72-deviations-from-markdown


I think these specifically have a chance to be problematic for us 🤔
 
> HTML syntax doesn’t work in MDX as it’s replaced by JSX ( to ). Instead of HTML comments, you can use JavaScript comments in braces: {/* comment! */}
> Unescaped left angle bracket / less than (<) and left curly brace ({) have to be escaped: \< or \{ (or use expressions: {'<'}, {'{'})

FroyaTheHen: I'm aware, the use case for this is that documentation for corelib is generated by mintify, and the very first task for mvp is to change just the files extension, changing the file structure and generating a relevant syntax are out of the scope [see plan](https://github.com/software-mansion/scarb/issues/2775#issue-3540546664).

extensions/scarb-doc/src/docs_generation/markdown.rs
FroyaTheHen: my bad 🫡

utils/scarb-extensions-cli/src/doc.rs
```diff
@@ -20,6 +20,9 @@ pub enum OutputFormat {
     /// This may be useful if you want to generate documentation files by yourself.
     /// The precise output structure is not guaranteed to be stable.
     Json,
+    /// Generates documentation alike Markdown format with `.mdx` extension instead.
+    #[value(hide = true)] // do not advertise in --help but accept from CLI
```
maciektr, line 24: 👍 
Sounds like enough precautions to me :D

FroyaTheHen: do you think there's no need to change to `--output-format=unstable-mdx` ??

DelevoXDG: Not a strong opinion, but I'd rather add it

DelevoXDG (approved):
I'd maybe add a comment for us internally that this is work in progress and mdx syntax support is to be added: https://github.com/software-mansion/scarb/blob/31ce5df97cefe94e85781dd4ffea76cc2e12dcb6/extensions/scarb-doc/src/main.rs#L62

Looks good otherwise.

utils/scarb-extensions-cli/src/doc.rs
maciektr: I think it is completely fine to use just `mdx` as long as it's not suggested by the --help.

maciektr (approved):

PR #2828, opened 2025-11-07
Deduplicate cargo metadata calls across the whole run

maciektr:


scarb/src/ops/package.rs
```diff
@@ -274,6 +274,8 @@ fn prepare_archive_recipe(
     });
 
     if pkg.is_cairo_plugin() && !pkg.manifest.targets.iter().any(is_builtin) {
+        // Start getting the cargo metadata.
+        spawn_cargo_metadata(pkg);
```
DelevoXDG: Why not move this inside `package_crate`? 🤔

maciektr: We can, moved.

PR #2830, opened 2025-11-07
Allow to skip compiler optimizations

piotmag769:


scarb/src/compiler/compilers/starknet_contract/validations.rs
```diff
@@ -120,12 +123,26 @@ pub fn check_allowed_libfuncs<'db>(
...
+            CompilerOptimizations::Enabled {
+                inlining_strategy: InliningStrategy::Default,
+            } => {}
+            CompilerOptimizations::Disabled => {
+                ws.config().ui().error(indoc! {r#"
+                    this build runs in `release` profile, but has compiler optimizations turned off
+                    turning off the optimizations may significantly increase gas costs of calling entrypoints of the compiled contract
+                    please make sure the compiler configuration in your manifest file is intended
+                    see https://docs.swmansion.com/scarb/docs/reference/manifest.html#skip-optimizations for more info"#});
+            }
```
piotmag769, line 137: Idk if this shouldn't just be a hard error - no one in their right mind does it cc @maciektr

scarb/src/compiler/helpers.rs
```diff
@@ -134,7 +134,6 @@ where
             diagnostics_reporter
         },
         replace_ids: unit.compiler_config.sierra_replace_ids,
-        inlining_strategy: unit.compiler_config.inlining_strategy.clone().into(),
```
piotmag769, line 137: Unused in compiler anyways, no point setting it

scarb/src/compiler/compilers/starknet_contract/validations.rs
maciektr: We didn't have an error like this before, but I guess this setting is so aggressive, that it would make sense to turn it into an error.

scarb/src/compiler/helpers.rs
maciektr: Can this be removed in Cairo as well then? 🤔

website/docs/reference/manifest.md
```diff
@@ -401,6 +401,25 @@ inlining-strategy = 18
 > If you need to deploy your contracts on Starknet, it is recommended to use the `release` strategy for their compilation.
 > You can use profile settings overwriting, for more granular control of which builds use the `avoid` and `release` strategy.
 
+### `skip-optimizations`
+
+If enabled, Scarb will skip as much compiler optimization as possible when compiling to Sierra.
+Since inlining is an optimization as well, setting this field to `true` will cause inlining to behave
+as if [`inlining-strategy`](#inlining-strategy) was set to `avoid`, regardless of the value of
```
maciektr: Let's throw an error unless the `inlining-strategy` is unset or explicitly set to `avoid`.

website/docs/reference/manifest.md
```diff
@@ -401,6 +401,25 @@ inlining-strategy = 18
...
+
+```toml
+[cairo]
+skip-optimizations = false
+```
+
+> [!WARNING]
+> Setting this field to `true` may result in faster compilation, but **much** slower execution of the compiled code.
+> Please use with caution.
+> If you need to deploy your contracts on Starknet, it is **strongly** recommended to set this field to `false` for their compilation.
```
maciektr:
```suggestion
> If you need to deploy your contracts on Starknet, **you should never** compile it with this field set to `true`.
```
Or smth

piotmag769: But maybe `always` and `true` bcs ppl won't read, see `true` etc. xD

scarb/src/compiler/helpers.rs
piotmag769: Planning to do so

scarb/src/compiler/compilers/starknet_contract/validations.rs
piotmag769: Should I exit the process there or just print the error msg?

maciektr: Just bail!

scarb/src/compiler/helpers.rs
piotmag769: https://github.com/starkware-libs/cairo/pull/8634

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -1275,19 +1306,48 @@ impl TomlManifest {
...
+            return Ok(());
+        };
+
+        if cairo.skip_optimizations == Some(true)
+            && cairo
+                .inlining_strategy
+                .as_ref()
+                .is_none_or(|inlining_strategy| inlining_strategy == &InliningStrategy::Avoid)
+        {
+            ui.error(formatdoc! {r#"
```
maciektr: Can we add a test case that hits this?

maciektr: nit: you can just do `bail!(formatdoc`, but both are fine

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -1275,19 +1306,48 @@ impl TomlManifest {
...
+    fn verify_toml_cairo_section(profile_definition: &TomlProfile, ui: Ui) -> Result<()> {
+        let Some(cairo) = &profile_definition.cairo else {
+            return Ok(());
+        };
+
+        if cairo.skip_optimizations == Some(true)
+            && cairo
+                .inlining_strategy
+                .as_ref()
+                .is_none_or(|inlining_strategy| inlining_strategy == &InliningStrategy::Avoid)
```
maciektr:
```suggestion
            && !cairo
                .inlining_strategy
                .as_ref()
                .is_none_or(|inlining_strategy| inlining_strategy == &InliningStrategy::Avoid)
```
Isn't `None | "avoid"` actually what we want the user to use?

scarb/src/core/manifest/toml_manifest.rs
piotmag769: Will add tests when I fix contract compilation with `Optimizations::Disabled` in compiler

piotmag769:
Closing in favour of #2850

PR #2831, opened 2025-11-07
support exposed macro calls in doc

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2643


extensions/scarb-doc/src/types/other_types.rs
```diff
@@ -43,11 +43,29 @@ pub struct ItemData<'db> {
...
+                "{}::{}",
+                doc_full_path(&id.parent_module(db), db),
+                id.name(db).long(db)
+            )
+        }
+        ModuleId::MacroCall { id, .. } => {
+            format!("{}", doc_full_path(&id.parent_module(db), db))
+        }
+    }
+}
```
FroyaTheHen, line 61: if we didn't omit the macro modules the path would look like for example"
now: `hello::func_macro_fn_outter` vs  before: `hello::define_fn_outter!(func_macro_fn_outter);::expose! {\n\t\t\tpub fn func_macro_fn_outter() -> felt252 { \n\t\t\t\tprintln!(\"hello world\");\n\t\t\t\t10 }\n\t\t}::func_macro_fn_outter`

1. We simply cannot have a file name `hello_define_fn_outter!(func_macro_fn_outter);_expose! {\n\t\t\tpub fn func_macro_fn_outter() -> felt252 { \n\t\t\t\tprintln!(\"hello world\");\n\t\t\t\t10 }\n\t\t}_func_macro_fn_outter.md`. The implementation for the _Fully qualified path_ would enforce creating such file. 

2. I imagine that once a macro exposes an item (in the example below a module named `an_exposed_module`) for example a module (in the example below named `root_module` the expected path documentation is `root_module::an_exposed_module` which calls for omitting the macro module in the path returned from compiler.

extensions/scarb-doc/tests/data/expose_macros/src/hello_world.md
```diff
@@ -0,0 +1,46 @@
+# hello_world
+
+Fully qualified path: [hello_world](./hello_world.md)
+
+
+## [Modules](./hello_world-modules.md)
+
+| | |
+|:---|:---|
+| [not_public](./hello_world-not_public.md) | — |
```
maciektr: Is it expected, that private items are doced here? 🤔

extensions/scarb-doc/src/types/module_type.rs
```diff
@@ -176,6 +172,19 @@ impl<'db> ModulePubUses<'db> {
             && self.use_submodules.is_empty()
             && self.use_macro_declarations.is_empty()
     }
+
+    fn add(&mut self, other: Self) {
```
maciektr, line 176: Why only some items are added here? 🤔  (i.e. why `use_extern_types` are, but `use_extern_functions` are not)

extensions/scarb-doc/src/types/module_type.rs
```diff
@@ -687,3 +681,222 @@ where
         })
         .collect::<Maybe<Maybe<Vec<K>>>>()?
 }
+
+type ModuleItems<'db> = (
```
maciektr: nit: it seems even simpler, to make it into a struct, and construct/destruct by names, but it's fine too.

extensions/scarb-doc/src/types/other_types.rs
maciektr: Please add a short comment describing this behaviour as well 🙏

extensions/scarb-doc/src/types/module_type.rs
FroyaTheHen: my bad

extensions/scarb-doc/tests/data/expose_macros/src/hello_world.md
FroyaTheHen: yes, because it's `2023_01` edition, although makes more sense to test this against private items. I'll change accordingly.

extensions/scarb-doc/src/types/module_type.rs
maciektr: NIT: If we want to use all, we can ensure this via destructuring 
```
    fn add(&mut self, other: Self) {
        let Self {
            use_constants,
            use_free_functions,
            use_structs,
            use_enums,
            use_module_type_aliases,
            use_impl_aliases,
            use_traits,
            use_impl_defs,
            use_extern_types,
            use_extern_functions,
            use_submodules,
            use_macro_declarations,
        } = other;
        self.use_constants.extend(use_constants);
        self.use_free_functions.extend(use_free_functions);
        self.use_structs.extend(use_structs);
        self.use_enums.extend(use_enums);
        self.use_module_type_aliases.extend(use_module_type_aliases);
        self.use_impl_aliases.extend(use_impl_aliases);
        self.use_traits.extend(use_traits);
        self.use_impl_defs.extend(use_impl_defs);
        self.use_extern_types.extend(use_extern_types);
        self.use_extern_functions.extend(use_extern_functions);
        self.use_submodules.extend(use_submodules);
        self.use_macro_declarations.extend(use_macro_declarations);
    }
```

maciektr (approved):

extensions/scarb-doc/tests/data/expose_macros/src/hello_world.md
maciektr: Right 🤦

PR #2841, opened 2025-11-12
Update Scarb new template

maciektr:


examples/hello_world/src/lib.cairo
```diff
@@ -1,25 +1 @@
...
-#[cfg(test)]
-mod tests {
-    use super::fib;
-
-    #[test]
-    fn it_works() {
-        assert(fib(16) == 987, 'it works!');
-    }
-}
+mod hello_world;
```
piotmag769, line 1: Why dis?

maciektr:
snfoundry PR https://github.com/foundry-rs/starknet-foundry/pull/3932

scarb/src/ops/new.rs
```diff
@@ -208,6 +166,93 @@ fn mk(
...
+        }
+    }
+
+    fn materialize(&self, source_path: &Utf8Path) -> Result<()> {
+        for (content_path, content) in self.src.iter() {
+            let path = source_path.join(content_path.as_ref());
+            fsx::create_dir_all(path.parent().expect("file path must have a parent"))?;
+
+            match content_path {
+                SourcePath::SkipDuplicated(path) if path.exists() => continue,
```
DelevoXDG, line 246: Shouldn't we check existence of `canonical_path.join(path)`? It seems we're checking the existence of the relative path here 🤔

maciektr: https://github.com/software-mansion/scarb/pull/2878

Issue #2794, opened 2025-10-29
(mvp) adjust the files formatting for .mdx


FroyaTheHen:
expected example md vs mdx: 

> # Array
> 
> A collection of elements of the same type contiguous in memory.
> 
> Fully qualified path: [core](./core.md)::[array](./core-array.md)::[Array](./core-array-Array.md)
> 
> <pre><code class="language-cairo">pub extern type Array&lt;T&gt;;</code></pre>

> 
> ---
> title: "core::array::Array"
> ---
> 
> A collection of elements of the same type contiguous in memory.
> 
> 
> 
> ## Signature
> 
> ```rust
> pub extern type Array;
> ```
> 

ask @FroyaTheHen for more examples if needed, 
Note: this could be easy to improve mdx version with links for example, if so go for it but make sure before that it is correctly interpreted mintify

PR #2843, opened 2025-11-13
generate mdx file content

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2794

- I decided to treat the `.mdx` as a special case of markdown, this aligns with the [mdx idea](https://mdxjs.com/#:~:text=Markdown%20for%20the%0Acomponent%20era) and saves us from writing and maintaining a lot of additional code that would not differ significantly from what's already implemented for markdown output

### DIFF
--- a/extensions/scarb-doc/src/docs_generation/markdown/context.rs
+++ b/extensions/scarb-doc/src/docs_generation/markdown/context.rs
@@ -1,22 +1,113 @@
+use crate::docs_generation::common::{
+    OutputFilesExtension, SummaryIndexMap, get_filename_with_extension,
+};
+use crate::docs_generation::markdown::SUMMARY_FILENAME;
 use crate::docs_generation::markdown::traits::WithPath;
-use crate::docs_generation::markdown::{SUMMARY_FILENAME, get_filename_with_extension};
+use crate::location_links::DocLocationLink;
 use crate::types::crate_type::Crate;
 use cairo_lang_defs::ids::{ImplItemId, LookupItemId, TraitItemId};
 use cairo_lang_doc::documentable_item::DocumentableItemId;
 use cairo_lang_doc::parser::CommentLinkToken;
+use itertools::Itertools;
 use std::collections::HashMap;
 
 pub type IncludedItems<'a, 'db> = HashMap<DocumentableItemId<'db>, &'a dyn WithPath>;
 
 pub struct MarkdownGenerationContext<'a, 'db> {
     included_items: IncludedItems<'a, 'db>,
+    formatting: Box<dyn Formatting>,
+}
+
+pub trait Formatting {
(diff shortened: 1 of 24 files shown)


extensions/scarb-doc/src/docs_generation/markdown/context.rs
```diff
@@ -1,22 +1,113 @@
...
 use std::collections::HashMap;
 
 pub type IncludedItems<'a, 'db> = HashMap<DocumentableItemId<'db>, &'a dyn WithPath>;
 
 pub struct MarkdownGenerationContext<'a, 'db> {
     included_items: IncludedItems<'a, 'db>,
+    formatting: Box<dyn Formatting>,
+}
+
+pub trait Formatting {
```
maciektr, line 21: Won't this grow pretty quickly, as we start to support more and more concepts from both markdowns? 🤔

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -78,13 +77,33 @@ macro_rules! impl_markdown_doc_item {
...
                 summary_index_map: &SummaryIndexMap,
             ) -> Result<String> {
-                generate_markdown_from_item_data(
-                    self,
-                    context,
-                    header_level,
-                    item_suffix,
-                    summary_index_map,
-                )
+                let mut markdown = String::new();
```
maciektr, line 80: Why we cannot use the function here no more? 🤔

extensions/scarb-doc/src/docs_generation/markdown/context.rs
FroyaTheHen: It would but I do not expect to that we will suport many more concepts.

maciektr: 👌

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
FroyaTheHen: For the items that are documented within the parent page the formatting differs. See for example `extensions/scarb-doc/tests/data/hello_world_no_features_mdx/src/hello_world-Circle.mdx
` and how in `.mdx` the parent `Circle` content is different than the struct members within the same page. It applies for all of the items that are documented within the parent page.

FroyaTheHen: *Fiy, I did force update here one thing, though.  `SubPathDocItem` should not be implemented for  for `MacroDeclaration` as it is an item that has it's own page. That fixes the `Fully qualified path` for macro declarations, see changes in `extensions/scarb-doc/tests/data/hello_world_doc_groups_reeksports/src/hello_world-macro_module-macro_definition.md`

maciektr (approved):

PR #2850, opened 2025-11-13
Allow to skip compiler optimizations

piotmag769:


scarb/src/compiler/compilers/starknet_contract/validations.rs
```diff
@@ -120,12 +123,26 @@ pub fn check_allowed_libfuncs<'db>(
...
+            } => {}
+            CompilerOptimizations::Disabled => {
+                ws.config().ui().error(indoc! {r#"
+                    this build runs in `release` profile, but has compiler optimizations turned off
+                    turning off the optimizations may significantly increase gas costs of calling entrypoints of the compiled contract
+                    see https://docs.swmansion.com/scarb/docs/reference/manifest.html#skip-optimizations for more info"#});
+                bail!("improper settings for `release` build of starknet contract target");
+            }
+            _ => {
+                ws.config().ui().warn(indoc! {r#"
```
maciektr, line 139: Can we have a test cases that hits that?

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -1289,18 +1299,43 @@ impl TomlManifest {
...
+        };
+
+        if cairo.skip_optimizations == Some(true)
+            && cairo
+                .inlining_strategy
+                .as_ref()
+                .is_some_and(|inlining_strategy| inlining_strategy != &InliningStrategy::Avoid)
+        {
+            bail!(formatdoc! {r#"
+                inlining-strategy field is set but its effects are overriden by skip-optimizations = true
```
maciektr, line 1314: Can we have a test case to hit that?

Issue #2840, opened 2025-11-12
Fix struct signature


FroyaTheHen:
Repro: 

```
pub struct A {
    pub b: felt252,
    c: felt252,
}
```

will have a signature generated as:
```
pub struct A {
    pub b: felt252,
    c: felt252,
}
``` 
instead of 

```
pub struct A {
    pub b: felt252,
}
```

whereas the members have the visibility checked properly.

Issue #2853, opened 2025-11-18
Fix attribute signature html formatting


FroyaTheHen:
**The problem:** 
let's take
```
            #[doc(group: "this is visible only within the scope")]
            struct Six {}
```
the signature in markdown will be evaluated to: 
```
<pre><code class="language-cairo">#[doc(group: &quot;test group 3&quot;)]
struct Three {}
</code></pre>
```
then the generated page will look like: 

<img width="1908" height="678" alt="Image" src="https://github.com/user-attachments/assets/290343dd-adcf-42be-b80d-711937c8ed8b" />

even though the  attribute  `#[doc(group: "this is visible only within the scope")]` is present it cannot be seen. 

**Solution:**
It should either be removed from the signature or made visible (by for example including an extra `<span>` around the signature - not sure though if that's the best solution).

PR #2854, opened 2025-11-18
hide private struct members from signature

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2840

- I do not like the solution but I cannot think of anything better. The compiler does not know what visibility is requested and I think it's better this way, and the signature that we get already does not contain the relevant semantic data (the scarb doc formatting could otherwise use). We could alternatively build the struct signature in scarb doc but that would be an overkill. If you have a better idea how to handle the issiue please let me know 🙏


extensions/scarb-doc/tests/data/json_doc_groups_reeksports.json
```diff
@@ -135,7 +166,48 @@
...
+                  "full_path": "hello_world::inner_module::TestMembersVisibility",
+                  "group": null
+                }
+              },
+              {
+                "members": [],
+                "item_data": {
+                  "name": "TestMembersVisibility2",
+                  "doc": null,
+                  "signature": "#[doc(group: \"visible\")]\npub struct TestMembersVisibility2 {}",
```
FroyaTheHen: https://github.com/software-mansion/scarb/issues/2853

extensions/scarb-doc/tests/data/json_reexports.json
```diff
@@ -130,7 +130,7 @@
                   "item_data": {
                     "name": "TestStruct",
                     "doc": null,
-                    "signature": "pub struct TestStruct {\n    abc: u32,\n}",
+                    "signature": "pub struct TestStruct {}",
```
maciektr: I see that Rust adds a comment marking omitted private items, like this:
```
pub struct TestStruct {/* private fields */}
```
(e.g. https://docs.rs/cairo-lang-macro/latest/cairo_lang_macro/struct.AuxData.html) 
Should we do smth similar? 🤔

FroyaTheHen: hmm, yes, I think it brings extra value 🙏

extensions/scarb-doc/src/types/other_types.rs
```diff
@@ -796,3 +843,41 @@ impl<'db> MacroDeclaration<'db> {
...
+/// Helper that removes a member signature from the struct signature and updates doc location links accordingly.
+fn adjust_signature_rows_and_links_for_removed_member(
+    lines: &mut Vec<String>,
+    links: &mut Vec<DocLocationLink>,
+    member_sig: &str,
+) {
+    let Some((indices_to_remove, _)) = lines
+        .iter()
+        .enumerate()
+        .find(|(_, line)| line.contains(member_sig))
```
maciektr: Do we assert somewhere, that the HirFormatter can never emit more than one member per line / struct with some member in single line?

Maybe `line.trim() == member_sig` could work too, as a more defensive alternative? 🤔 

Also, what if the member sig is part of some attr? 🤔 

Is it ever possible, that a very long member type declaration would span multiple lines? 🤔

extensions/scarb-doc/src/types/other_types.rs
maciektr, line 1: I feel like this impl could be encapsulated in some new struct that would hold struct signature lines and doc link locations, modify them as seen fit (e.g. with `fn remove(&mut self, member_sig?)`) then just `(item.signature, item.doc_link_locations) = impl.build()` at the end of the function 🤔

extensions/scarb-doc/src/types/other_types.rs
FroyaTheHen: > Do we assert somewhere, that the HirFormatter can never emit more than one member per line / struct with some member in single line?

We do not. The assumptions are based on the compiler code. I do not see how we could assert that, except having tests that would break in case the formatting changes. 
At this stage of signature formatting with syntax data only, the assumptions seem unavoidable. 

> Maybe line.trim() == member_sig could work too, as a more defensive alternative? 🤔

There are also potentially: proceeding whitespaces, newlines, commas to take into account when checking `==`.  I like the `contains` here because it simplifies the logic and we cannot possibly have a line that contains another-member line. 
Every member must have a type so there must be a colon and a whitespace in member definition and there is no way to define a member that would have two pairs of name-characters+whitespace+colon. Or please correct me if i'm wrong but i can't think of such syntax. And we're sure about the name-characters+space+colon because the signature was already formatted by compiler. 

> Also, what if the member sig is part of some attr? 🤔

If for example there is a macro that refers to the member then scarb should not care because the signature was already formatted with semantic data including the potential changes made to the field. 
Please elaborate if I misunderstood 🙏 

> Is it ever possible, that a very long member type declaration would span multiple lines? 🤔

No, with the current state of [state of members formatting logic](https://github.com/starkware-libs/cairo/blob/8a7e6a4f8330e6d198e1a4bc02a5bfce4b31f241/crates/cairo-lang-doc/src/documentable_formatter.rs#L343) it is not going to happen. 
I will add case to the test to have the changes detectable in scarb.

maciektr: >  I do not see how we could assert that, except having tests that would break in case the formatting changes.

So can we have such test cases in the compiler where it's implemented? 🤔 With a comment stating why it's important to not change this? 

>  I like the contains

Makes sense 👍 

> the signature was already formatted with semantic data

👍 

> I will add case to the test to have the changes detectable in scarb.

Can this be checked in the compiler as well? 🤔

PR #2862, opened 2025-11-20
Fix quote var separation

### LINKED ISSUE #2537: Quote may not handle complex types correctly
Repro:
```
quote! { let #name #name_type = snforge_std::fuzzable::Fuzzable::generate(); }
```
expands to:
```
let name:<TYPE>=snforge_std::fuzzable::Fuzzable::generate();
```
However, this may expand to a generic: 
```
let name:Span<MyStruct>=snforge_std::fuzzable::Fuzzable::generate();
```
then missing separator between > and = causes a syntax error.

DelevoXDG:
Fix #2537 

Adds whitespace separator between interpolated `#var` variables in `quote!` / `{}` args in `quote_format!` macros and the surrounding tokens

___
### Note

**Not ready for review**, might not account for some other cases:
- `#obj.#field`
- `let _#name`

### DIFF
--- a/plugins/cairo-lang-quote/src/lib.rs
+++ b/plugins/cairo-lang-quote/src/lib.rs
@@ -263,9 +263,12 @@ pub fn quote_format(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
                     }
                 };
                 output_token_stream.extend(rust_quote! {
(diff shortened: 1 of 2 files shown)


DelevoXDG:
This shall be addressed on Cairo side, so we'll not add any workarounds for now

PR #2863, opened 2025-11-20
Properly delete Scarb managed cache in tests

maciektr:
Temp directory created by `TempDir` is removed from disk when the associated `TempDir` is dropped. 
 When running Scarb in tests, new TempDir for Scarb cache is created and passed via env to the Scarb process. 
However, if the TempDir is dropped before calling Scarb, Scarb will re-create the cache directory on run, never actually deleting it.

### DIFF
--- a/utils/scarb-test-support/src/command.rs
+++ b/utils/scarb-test-support/src/command.rs
@@ -12,6 +12,9 @@ use std::path::{Path, PathBuf};
 use std::process::Command as StdCommand;
 use std::sync::LazyLock;
 
+#[cfg(feature = "scarb-config")]
+use camino::Utf8Path;
+
 pub struct Scarb {
     cache: EnvPath,
     config: EnvPath,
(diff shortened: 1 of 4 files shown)


maciektr:

- #2863 ◀
    - #2867

utils/scarb-test-support/src/command.rs
```diff
@@ -43,21 +45,26 @@ impl Scarb {
...
     }
 
-    pub fn snapbox(self) -> SnapboxCommand {
-        SnapboxCommand::from_std(self.std())
+    pub fn snapbox(self) -> ScarbCommand {
+        let inner = SnapboxCommand::from_std(self.std());
+        let mut state: Vec<Box<dyn Drop>> = vec![Box::new(self.cache), Box::new(self.config)];
+        if let Some(target) = self.target {
+            state.push(Box::new(target));
+        }
```
DelevoXDG: Wouldn't it be simpler and slightly more clear to add a struct that would hold these e.g.
```rs
#[allow(dead_code)]
struct ScarbState {
    cache: EnvPath,
    config: EnvPath,
    target: Option<EnvPath>,
}
```
Using concrete types should allow us to avoid using `#![allow(dyn_drop)]` 🤔

utils/scarb-test-support/src/command.rs
```diff
@@ -43,21 +45,26 @@ impl Scarb {
...
+        ScarbCommand { inner, state }
     }
 
-    pub fn std(self) -> StdCommand {
-        let mut cmd = StdCommand::new(self.scarb_bin);
-        cmd.env("SCARB_LOG", self.log);
+    pub fn std(&self) -> StdCommand {
+        let mut cmd = StdCommand::new(self.scarb_bin.clone());
+        cmd.env("SCARB_LOG", self.log.clone());
```
DelevoXDG, line 74: Correct me if I'm wrong, but it would seem that for `std()` the issue would persist in some cases?

This can be for example observed in `can_control_scarb_log_with_cli`:
```rs
let mut cmd = Scarb::new().std();
```
In this case, `Scarb` instance is dropped in the end of the statement, thus `drop()` is called for `EnvPath`s right afterwards, before we actually run:
```rs
    SnapboxCommand::from_std(cmd)
        ...
        .success();
```

`&self` would only prevent the issue if the `Scarb` instance remained in the scope:
```rs
    let scarb = Scarb::new();
    let mut cmd = scarb.std();
```
which doesn't seem to be the case at any of the places where it's used

maciektr: Yes, this is a good point!

However, we rarely need to use `.std()` (less than 10 test cases afaik) so it's not a big deal. 

Although, to make it more explicit, I suggest we enforce the cache and config dirs to be managed by the user, in case `.std()` is used (with assert) - made the change here. 

Note that technically it's still possible to leak the temp dir (let's say you create it in some helper function, pass to cache function and then return built command from the helper), but it shouldn't really bother us.

utils/scarb-test-support/src/command.rs
maciektr: I think that both are fine, but I like `impl Drop`, as it directly suggest the point of keeping it - only to drop later.

Issue #2659, opened 2025-09-23
Excluding packages from doc


FroyaTheHen:
• Excluding specific packages from documentation generation

Some command-line proposals:

Package Exclusion

> scarb doc --workspace --exclude macros,macros_tests
> scarb doc --workspace --include math,ascii,utils

PR #2864, opened 2025-11-20
add --workspace --exclude

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2659

### DIFF
--- a/utils/scarb-ui/src/args/packages_filter.rs
+++ b/utils/scarb-ui/src/args/packages_filter.rs
@@ -37,6 +37,15 @@ pub struct PackagesFilter {
     /// Run for all packages in the workspace.
     #[arg(short, long, conflicts_with = "package")]
     workspace: bool,
+    /// Exclude packages from the command.
+    #[arg(
+        long,
+        value_delimiter = PACKAGES_FILTER_DELIMITER,
+        value_name = "SPEC",
+        requires = "workspace",
+        conflicts_with = "package"
+        )]
+    exclude: Vec<String>,
 }
 
 /// [`clap`] structured arguments that provide package selection.
@@ -67,6 +76,15 @@ pub struct PackagesFilterLong {
     /// Run for all packages in the workspace.
     #[arg(long, conflicts_with = "package")]
     workspace: bool,
+    /// Exclude packages from the command.
+    #[arg(
+        long,
+        value_delimiter = PACKAGES_FILTER_DELIMITER,
+        value_name = "SPEC",
+        requires = "workspace",
+        conflicts_with = "package"
+    )]
+    exclude: Vec<String>,
 }
 
 impl PackagesFilter {
@@ -96,13 +114,17 @@ impl PackagesFilter {
                 help: use the `--package` option to specify the package"
             );
         }
-
+        if !self.exclude.is_empty() {
+            bail!(
+                "`--exclude` is not supported in a context that requires a single package selection"
+            );
+        }
         let specs_filter: String = specs
             .iter()
             .map(|s| s.to_string())
             .collect::<Vec<_>>()
             .join(PACKAGES_FILTER_DELIMITER.to_string().as_str());
-        let found = Self::do_match_all::<S>(specs, self.workspace, members)?;
+        let found = Self::do_match_all::<S>(specs, vec![], self.workspace, members)?;
 
         ensure!(
             found.len() <= 1,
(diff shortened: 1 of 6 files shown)


website/docs/extensions/documentation-generation.md
```diff
@@ -1,6 +1,15 @@
 # Generating documentation
 
-`scarb doc` is a tool that generates documentation based on the code comments. Generation supports different output formats. The result is being placed inside the `/target/doc`.
+`scarb doc` is a tool that generates documentation based on the code comments. Generation supports different output formats. The result is being placed inside the `/target/doc` directory.
+
+### Generating workspace documentation
+You can run `scarb doc --workspace` to generate documentation for all packages in the workspace. 
+
+Use `--exclude` to omit the workspace packages in the documentation. Must be used with the `--workspace` flag.
```
FroyaTheHen: Not sure if we want to advertise `--exclude` somewhere else in web documentation, wdyt?

utils/scarb-ui/src/args/packages_filter.rs
```diff
@@ -67,6 +75,14 @@ pub struct PackagesFilterLong {
...
     #[arg(long, conflicts_with = "package")]
     workspace: bool,
+    /// Exclude packages from the command.
+    #[arg(
+        long,
+        value_delimiter = PACKAGES_FILTER_DELIMITER,
+        value_name = "SPEC",
+        requires = "workspace"
+    )]
+    exclude: Vec<String>,
```
maciektr, line 87: I guess we should only allow using this with `workspace` argument, right? 🤔 Can we have a test case for that?

utils/scarb-ui/src/args/packages_filter.rs
```diff
@@ -257,9 +300,15 @@ enum Spec<'a> {
     All,
     One(&'a str),
     Glob(&'a str),
+    Exclude(&'a str),
```
maciektr: This may be personal, but I feel like this smears the responsibility a bit between Spec and PackagesFilter. For me, the Spec should only be concerned with asserting whether some package name *matches* some package name spec (e.g. a single `--package <name>` arg). 

Maybe we should just `Spec::parse` both the `package` and `exclude` specs, then keep their kind externally (e.g. by wrapping in an enum, or just keeping them in two vecs)? 

I feel like this would should even simplify the logic in this file a bit 🤔 
We would also get globs support in exclude for free (not like this matters really though 😅 ) 

lmkwyt

website/docs/extensions/documentation-generation.md
maciektr: I think it's fine. It will pop up in --help anyway.

utils/scarb-ui/src/args/packages_filter.rs
FroyaTheHen: Sure!

utils/scarb-ui/src/args/packages_filter.rs
FroyaTheHen: Good idea! Definitely the implementation can be simplified this way and even that argument single puts my opinion in favour of the changes.

utils/scarb-ui/src/args/packages_filter.rs
```diff
@@ -102,7 +120,7 @@ impl PackagesFilter {
             .map(|s| s.to_string())
             .collect::<Vec<_>>()
             .join(PACKAGES_FILTER_DELIMITER.to_string().as_str());
-        let found = Self::do_match_all::<S>(specs, self.workspace, members)?;
+        let found = Self::do_match_all::<S>(specs, vec![], self.workspace, members)?;
```
maciektr, line 127: Why not pass exclude here as well? 🤔

FroyaTheHen: If any exclude spec exists this line is close to unreachable because `--exclude` requires `--workspace` and if `-w` was passed then  
the above ☝️ 
```
       if (self.workspace || specs.iter().any(|spec| matches!(spec, Spec::All)))
            && members.len() > 1
        {
            bail!(...
```
either fails or nither `-w` nor `--exclude` passed to the command make sense imo: because at this point we know for sure that there is one workspace package, then there are two ways: 1. exclude the very package - I see no reason to ever do that (??). 
2. exclude "random characters" - again no reason (??). 

From that i drive the conclusion that it doesn't make sense to consider exclude at this point. 
As alternative i'd consider changing the conditions for ```bail!( "could not determine which package to work on...``` or banning the workspace for fn match_one and I'd reason that it's not reasonable to specify a package by specifying a workspace and excluding everything else from the workspace. 
However I see clearly that the current might be confusing if someone plays with the commands and the code is not there to address reasonable only. 
Let me know what you think 🙏

maciektr: @FroyaTheHen Seems like we should bail with some message, e.g. "cannot use `exclude` in a context that expects a single package to be specified" there, so it's not silently ignored

FroyaTheHen: ok, added `bail` and a test case in `scarb/tests/expand.rs` 🫡

PR #2869, opened 2025-11-24
Save `scarb execute` outputs

DelevoXDG:
## Stack
- #2886 
- #2869 
- #2870
- #2876
- #2893
---

Save `scarb execute` output to files so we can embed the outputs in `scarb doc` generated docs

### DIFF
--- a/utils/scarb-extensions-cli/src/execute/unchecked.rs
+++ b/utils/scarb-extensions-cli/src/execute/unchecked.rs
@@ -124,6 +124,14 @@ pub struct RunArgs {
     /// Whether to save cairo-profiler trace data.
     #[arg(long, default_value_t = false)]
     pub save_profiler_trace_data: bool,
+
+    /// Whether to save program output to a file.
+    #[arg(long, default_value_t = false)]
+    pub save_program_output: bool,
+
+    /// Whether to save stdout output to a file.
+    #[arg(long, default_value_t = false)]
+    pub save_print_output: bool,
 }
 
 impl ToArgs for RunArgs {
@@ -135,6 +143,8 @@ impl ToArgs for RunArgs {
             print_program_output,
             print_resource_usage,
             save_profiler_trace_data,
+            save_program_output,
+            save_print_output,
         } = self;
         let mut args = arguments.to_args();
         if let Some(output) = output {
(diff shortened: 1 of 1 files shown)


extensions/scarb-execute/src/lib.rs
```diff
@@ -318,6 +333,18 @@ pub fn execute(
...
+    {
+        let program_output_path = execution_output_dir.join(EXECUTE_PROGRAM_OUTPUT_FILENAME);
+        fs::write(program_output_path, output.as_str())?;
+    }
+
+    if args.run.save_print_output {
+        let print_output_path = execution_output_dir.join(EXECUTE_PRINT_OUTPUT_FILENAME);
+        fs::write(print_output_path, &captured_print_output)?;
+    }
+
```
maciektr, line 347: I wonder how useful is it, to save this to files, instead of printing to stdout? Can't we just parse the stdout wherever we need to use this? 🤔

DelevoXDG: 1. I based this off of the decision we made in the past regarding `scarb prove`: https://github.com/software-mansion/scarb/pull/1900#discussion_r1925479291

   Although I'm not sure that's super problematic since we have control over the logic of both extensions, I suppose if extracting `N` from `Saving output to [..]N` is unstable, then extracting multiline outputs could be considered even less stable 🤔
That's why we use `SCARB_EXECUTION_ID`, and it seemed reasonable to me not to rethink this here

2. Another thing that concerned me - verbosity. 
   - If we pass verbosity from `doc` to `execute`, and that verbosity is `--quiet`, the output is lost
   - If we force a normal verbosity for `execute`, we'd have to capture `execute` stdout, and then somehow figure out based on `doc` verbosity what to print and not to print. Since we'd have only raw text at our disposal, there would be no way to filter out messages based on their intended level and `doc` verbosity
Maybe there's some elegant solution to this, but I just didn't think it's worth going there since the files should just work 😄

maciektr: I see your point, but I don't really think these two are comparable. 

For starters, I don't think that using HintProcessor to capture print is reliably what we want. It seems it can be bypassed (see running with bootloader), but also it runs on the assumption that println! it's the only possible way to access stdout (which I am not sure is true). 

It shouldn't require any unpredictable parsing, as we can expect the program output to be the last message when run with `--program-output` flag. We can use json output format in ui, to output it as easily parsable json. We can also do what we do in `scarb metadata`, that is, run the command with `--quite`, but set the output to always be printed regardless of verbosity (imho makes sense, if you are explicitly for it to be printed).

PR #2870, opened 2025-11-24
Unify `incremental_create_output_dir`

DelevoXDG:
## Stack
- #2886 
- #2869 
- #2870
- #2876
- #2893
---

### DIFF
--- a/utils/scarb-fs-utils/src/lib.rs
+++ b/utils/scarb-fs-utils/src/lib.rs
@@ -1,4 +1,4 @@
-//! Helper utilities shared by Scarb execute-related commands.
+//! Helper utilities for filesystem operations shared by Scarb commands.
 
 use anyhow::{Result, bail};
 use camino::{Utf8Path, Utf8PathBuf};
@@ -10,13 +10,13 @@ const MAX_ITERATION_COUNT: usize = 10000;
 pub const EXECUTE_PROGRAM_OUTPUT_FILENAME: &str = "program_output.txt";
 pub const EXECUTE_PRINT_OUTPUT_FILENAME: &str = "stdout_output.txt";
 
-/// Creates an incremental execution output dir inside the given path.
-/// The dir name is `execution{N}` with the lowest `N` without existing dir.
+/// Creates an incremental directory inside the given path.
+/// The dir name is `{name}{N}` with the lowest `N` without existing dir.
 ///
 /// Returns the path to the created directory and corresponding `N`.
-pub fn incremental_create_execution_output_dir(path: &Utf8Path) -> Result<(Utf8PathBuf, usize)> {
+pub fn incremental_create_dir_unique(path: &Utf8Path, name: &str) -> Result<(Utf8PathBuf, usize)> {
     for i in 1..=MAX_ITERATION_COUNT {
-        let filepath = path.join(format!("execution{i}"));
+        let filepath = path.join(format!("{name}{i}"));
         let result = fs::create_dir(&filepath);
         return match result {
             Err(e) => {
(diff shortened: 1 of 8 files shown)


utils/create-output-dir/src/lib.rs
maciektr, line 1: I feel like this crate solves a very specific problem, and I do not like the idea of moving other utils here.

DelevoXDG: Moved to a separate crate

utils/scarb-execute-utils/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "scarb-execute-utils"
```
maciektr: Nit: Maybe just `scarb-fs-utils`? 🤔 Seems general enough to me.

utils/scarb-fs-utils/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use anyhow::{Result, bail};
+use camino::{Utf8Path, Utf8PathBuf};
+use std::fs;
+use std::io;
+
+const MAX_ITERATION_COUNT: usize = 10000;
+
+pub const EXECUTE_PROGRAM_OUTPUT_FILENAME: &str = "program_output.txt";
+pub const EXECUTE_PRINT_OUTPUT_FILENAME: &str = "stdout_output.txt";
```
maciektr, line 11: I'd just keep them where they are used.

utils/scarb-execute-utils/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+const MAX_ITERATION_COUNT: usize = 10000;
+
+pub const EXECUTE_PROGRAM_OUTPUT_FILENAME: &str = "program_output.txt";
+pub const EXECUTE_PRINT_OUTPUT_FILENAME: &str = "stdout_output.txt";
+
+/// Creates an incremental execution output dir inside the given path.
+/// The dir name is `execution{N}` with the lowest `N` without existing dir.
+///
+/// Returns the path to the created directory and corresponding `N`.
+pub fn incremental_create_execution_output_dir(path: &Utf8Path) -> Result<(Utf8PathBuf, usize)> {
```
maciektr: Let's just take `execution` as param.

```suggestion
pub fn incremental_create_dir_unique(path: &Utf8Path, name: &str) -> Result<(Utf8PathBuf, usize)> {
```

utils/scarb-fs-utils/src/lib.rs
DelevoXDG: I moved them here so we could re-use these both in `scarb execute` and `scarb doc` in #2876.

Can define these in both places, but not sure why?

IMO it's reasonable to keep this crate as `scarb execute` utils rather then general. If we do, it makes sense to keep those here.

utils/scarb-execute-utils/src/lib.rs
DelevoXDG: I'm not sure we expect `incremental_create_dir_unique` to be used in any other contexts? 🤔 

If we don't, it seems reasonable to me to keep this crate `scarb execute`-focused.

maciektr: >  it seems reasonable to me to keep this crate scarb execute-focused

I'd look at this from the other side, that is, it makes sense to me to define some crate for how Scarb interacts with the filesystem. E.g. canonicalization helpers from internal/fsx.rs could go there in the future - this is not Scarb specific, and would make sense in the extensions as well.

Issue #2299, opened 2025-06-06
Variant and Member should not have groups


FroyaTheHen:
```
pub struct ItemData {
    ...
    pub group: Option<String>,
}
```
is relevant for Variant and Members and there is no need for those to have groups, this generates unnecessary json output summary

PR #2874, opened 2025-11-24
omit group in variant and member serialization

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2299

### DIFF
--- a/extensions/scarb-doc/src/types/other_types.rs
+++ b/extensions/scarb-doc/src/types/other_types.rs
@@ -1,21 +1,17 @@
-use anyhow::Result;
-
-use crate::attributes::find_groups_from_attributes;
 use crate::db::ScarbDocDatabase;
-use crate::location_links::DocLocationLink;
+use crate::docs_generation::markdown::context::IncludedItems;
+use crate::docs_generation::markdown::traits::WithPath;
+use crate::types::item_data::{ItemData, SubItemData};
 use crate::types::module_type::is_doc_hidden_attr;
 use cairo_lang_defs::ids::{
     ConstantId, EnumId, ExternFunctionId, ExternTypeId, FreeFunctionId, ImplAliasId,
     ImplConstantDefId, ImplDefId, ImplFunctionId, ImplItemId, ImplTypeDefId, LanguageElementId,
     LookupItemId, MacroDeclarationId, MemberId, ModuleId, ModuleItemId, ModuleTypeAliasId,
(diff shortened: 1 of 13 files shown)


extensions/scarb-doc/src/types/other_types.rs
```diff
@@ -122,21 +123,26 @@ impl<'db> ItemData<'db> {
...
 }
 
+/// Unifies serialization logic for `doc` filed of `ItemData` in `documentation_serializer` and `serialize_subitem_item_data`
+fn docs_to_single_string(docs: &Option<Vec<DocumentationCommentToken>>) -> Option<String> {
+    docs.as_ref().map(|doc_vec| {
+        doc_vec
+            .iter()
+            .map(|dct| dct.to_string())
+            .collect::<Vec<String>>()
+            .join("")
```
maciektr: NIT
```suggestion
            .collect::<String>()
```

extensions/scarb-doc/src/types/other_types.rs
```diff
@@ -796,3 +802,17 @@ impl<'db> MacroDeclaration<'db> {
...
+    S: Serializer,
+{
+    let mut state = serializer.serialize_struct("ItemData", 4)?;
+    state.serialize_field("name", &item.name)?;
+    let doc_string: Option<String> = docs_to_single_string(&item.doc);
+    state.serialize_field("doc", &doc_string)?;
+    state.serialize_field("signature", &item.signature)?;
+    state.serialize_field("full_path", &item.full_path)?;
+    state.end()
+}
```
maciektr: I don't like this approach, as it would be super easy to miss updating this function, e.g. when a new field is added to `ItemData`. I'd prefere to rewrite this structs to some dto, and serialize through this types with `serde(from)` and `serde(into)` https://serde.rs/container-attrs.html#from

For example: https://github.com/software-mansion/scarb/blob/3b6bf5117f9fe0ca5533034ee80149f0880526cd/scarb/src/core/lockfile.rs#L38-L38

FroyaTheHen: I didn't use the `serde(from/into)`  because the way to use it afaic would be to implement on the parent type (Struct/Enum) and this way I would need a dto for the types: Struct, Enum, ItemData. I believe this would make the implementation more complicated that the current proposition. Wdyt?

maciektr (approved):

PR #2876, opened 2025-11-25
Support runnable code blocks in docs

DelevoXDG:
## Stack
- #2886 
- #2869 
- #2870
- #2876
- #2893
---
- Support `cairo, runnable` code blocks in `scarb doc`
- Run said code blocks via `scarb execute`
- Embed print output and program output under the code block


extensions/scarb-doc/src/doc_test/code_blocks.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+    }
+}
+
+/// Represents code block extracted from doc comments.
+#[derive(Debug, Clone, PartialEq)]
+pub struct CodeBlock {
+    pub id: CodeBlockId,
+    pub content: String,
+    pub attributes: Vec<CodeBlockAttribute>,
```
FroyaTheHen: This is to have more than one attr? can we have such a test case?

extensions/scarb-doc/tests/code/code_12.cairo
```diff
@@ -0,0 +1,33 @@
...
+  /// ```cairo
+  ///     bar();
+  /// ```
+  pub fn bar() {
+      println!("bar");
+  }
+
+  /// Function that calls both foo and bar functions.
+  /// Can invoke it like that:
+  /// ```cairo,runnable
```
FroyaTheHen: what if we do: 
` /// ```rust,runnable` ??

extensions/scarb-doc/tests/runnable_examples.rs
```diff
@@ -0,0 +1,169 @@
...
+            error: Panicked with "Runtime error occurred".
+            test hello_world::foo ... FAILED
+
+            failures:
+                hello_world::foo
+
+            test result: FAILED. 0 passed; 1 failed; 0 ignored
+            error: doc tests failed
+        "#});
+}
```
FroyaTheHen, line 169: please add cases for `--output-format`: `mdx` and maybe `json`. 
Are you sure we want to generate the output for runnables in `mdx`??

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -0,0 +1,245 @@
...
+        let build_result = ScarbCommand::new()
+            .arg("build")
+            .current_dir(ws.root())
+            .env("SCARB_TARGET_DIR", target_dir.as_str())
+            .env("SCARB_UI_VERBOSITY", self.ui.verbosity().to_string())
+            .env("SCARB_MANIFEST_PATH", ws.manifest_path().as_str())
+            .env("SCARB_ALL_FEATURES", "true")
+            .run();
+
+        if build_result.is_err() {
```
FroyaTheHen, line 206: The build fails when `target.starknet-contract` present, what is the plan to address that?

PR #2880, opened 2025-11-26
Bump Stwo 

maciektr:
- **Bump Stwo**
- **Bump nightly version**

### DIFF
--- a/extensions/scarb-execute/tests/build.rs
+++ b/extensions/scarb-execute/tests/build.rs
@@ -4,35 +4,11 @@ use assert_fs::prelude::PathAssert;
 use indoc::{formatdoc, indoc};
 use predicates::prelude::predicate;
 use scarb_test_support::command::Scarb;
+use scarb_test_support::fixtures::{build_executable_project, executable_project_builder};
 use scarb_test_support::fsx::ChildPathEx;
 use scarb_test_support::project_builder::ProjectBuilder;
 use snapbox::{Assert, Data, Redactions};
 use test_case::test_case;
-
-fn executable_project_builder() -> ProjectBuilder {
-    ProjectBuilder::start()
-        .name("hello")
-        .version("0.1.0")
-        .dep_cairo_execute()
-        .manifest_extra(indoc! {r#"
-            [executable]
-
-            [cairo]
-            enable-gas = false
-        "#})
-        .lib_cairo(indoc! {r#"
-            #[executable]
-            fn main() -> felt252 {
-                42
-            }
-        "#})
-}
-
-fn build_executable_project() -> TempDir {
-    let t = TempDir::new().unwrap();
-    executable_project_builder().build(&t);
-    t
-}
 
 #[test_case("standalone")]
 #[test_case("bootloader")]
--- a/extensions/scarb-prove/tests/heavy.rs
+++ b/extensions/scarb-prove/tests/heavy.rs
@@ -1,31 +1,8 @@
-use assert_fs::TempDir;
 use assert_fs::assert::PathAssert;
 use assert_fs::prelude::PathChild;
 use indoc::indoc;
 use scarb_test_support::command::Scarb;
-use scarb_test_support::project_builder::ProjectBuilder;
-
-fn build_executable_project() -> TempDir {
-    let t = TempDir::new().unwrap();
-    ProjectBuilder::start()
-        .name("hello")
-        .version("0.1.0")
-        .dep_cairo_execute()
-        .manifest_extra(indoc! {r#"
-                [executable]
-
-                [cairo]
-                enable-gas = false
-            "#})
-        .lib_cairo(indoc! {r#"
-            #[executable]
-            fn main() -> felt252 {
-                42
-            }
-        "#})
-        .build(&t);
-    t
-}
+use scarb_test_support::fixtures::build_executable_project;
 
 #[test]
 #[cfg(not(windows))]
--- a/extensions/scarb-prove/tests/build.rs
+++ b/extensions/scarb-prove/tests/build.rs
@@ -1,31 +1,8 @@
-use assert_fs::TempDir;
 use assert_fs::assert::PathAssert;
 use assert_fs::fixture::PathChild;
 use indoc::indoc;
 use scarb_test_support::command::Scarb;
-use scarb_test_support::project_builder::ProjectBuilder;
-
-fn build_executable_project() -> TempDir {
-    let t = TempDir::new().unwrap();
-    ProjectBuilder::start()
-        .name("hello")
-        .version("0.1.0")
-        .dep_cairo_execute()
-        .manifest_extra(indoc! {r#"
-                [executable]
-
-                [cairo]
-                enable-gas = false
-            "#})
-        .lib_cairo(indoc! {r#"
-            #[executable]
-            fn main() -> felt252 {
-                42
-            }
-        "#})
-        .build(&t);
-    t
-}
+use scarb_test_support::fixtures::build_executable_project;
 
 #[test]
 #[cfg(not(windows))]
--- a/extensions/scarb-verify/tests/build.rs
+++ b/extensions/scarb-verify/tests/build.rs
@@ -1,29 +1,6 @@
-use assert_fs::TempDir;
 use indoc::indoc;
 use scarb_test_support::command::Scarb;
-use scarb_test_support::project_builder::ProjectBuilder;
-
-fn build_executable_project() -> TempDir {
-    let t = TempDir::new().unwrap();
-    ProjectBuilder::start()
-        .name("hello")
-        .version("0.1.0")
-        .dep_cairo_execute()
-        .manifest_extra(indoc! {r#"
-                [executable]
-
-                [cairo]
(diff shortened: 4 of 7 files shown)


extensions/scarb-execute/src/lib.rs
```diff
@@ -48,7 +60,7 @@ pub fn main_inner(args: Args, ui: Ui) -> Result<()> {
     execute(&metadata, &package, &args.execution, &ui)
 }
 
-fn read_arguments(arguments: ProgramArguments) -> Result<Vec<Arg>> {
+fn read_arguments_as_felt(arguments: ProgramArguments) -> Result<Vec<Arg>> {
```
DelevoXDG: nit: 

```suggestion
fn read_arguments_as_felts(arguments: ProgramArguments) -> Result<Vec<Arg>> {
```

extensions/scarb-execute/src/lib.rs
```diff
@@ -62,42 +74,80 @@ fn read_arguments(arguments: ProgramArguments) -> Result<Vec<Arg>> {
...
-
-    if !args.no_build {
-        let filter = PackagesFilter::generate_for::<Metadata>([package.clone()].iter());
-        ScarbCommand::new()
-            .arg("build")
-            .env("SCARB_PACKAGES_FILTER", filter.to_env())
-            .env("SCARB_UI_VERBOSITY", ui.verbosity().to_string())
-            .envs(args.features.clone().to_env_vars())
-            .run()?;
+fn read_arguments_as_value(arguments: ProgramArguments) -> Result<Vec<Value>> {
```
DelevoXDG: nit:
```suggestion
fn read_arguments_as_values(arguments: ProgramArguments) -> Result<Vec<Value>> {
```

extensions/scarb-execute/src/lib.rs
```diff
@@ -163,35 +193,122 @@ pub fn execute(
...
+    )?;
+
+    let cairo_run_config = build_cairo_run_config(&output, args)?;
+    let executable = load_prebuilt_executable(&executable_path)?;
+
+    let output_dir = scarb_target_dir.join("execute").join(&package.name);
+    create_output_dir(output_dir.as_std_path())?;
+
+    let execution_output_dir = get_or_create_output_dir(&output_dir)?;
+
```
DelevoXDG:
```suggestion
    let executable = load_prebuilt_executable(&executable_path)?;

    let output_dir = scarb_target_dir.join("execute").join(&package.name);
    create_output_dir(output_dir.as_std_path())?;

    let execution_output_dir = get_or_create_output_dir(&output_dir)?;
    let cairo_run_config = build_cairo_run_config(&output, args)?;
```

extensions/scarb-execute/src/output.rs
```diff
@@ -196,6 +201,51 @@ impl Message for ExecutionResources {
...
+        all_used_resources.basic_resources += &used_resources;
+
+        while !self.subtask_cairo1_hint_processor_stack.is_empty() {
+            let Some(cairo_hint_processor) =
+                self.subtask_cairo1_hint_processor_stack.pop().flatten()
+            else {
+                continue;
+            };
+            all_used_resources += cairo_hint_processor.syscalls_used_resources;
+        }
```
DelevoXDG, line 239: Can you please provide some context on what are we doing here? Both in this thread, and maybe in a form of comment for future reference 🤔

Cargo.toml
```diff
@@ -128,14 +130,14 @@ serde_json = { version = "1", features = ["raw_value"] }
 serde_repr = "0.1"
 serde_test = "1"
 sha2 = "0.10"
-shell-words = "1.1.1"
+shell-words = "1.1.0"
```
DelevoXDG: Why?

.github/workflows/ci.yml
```diff
@@ -109,7 +109,12 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
       - name: run tests
-        run: cargo test --doc
+        run: cargo test --doc --workspace --exclude scarb-prove --exclude scarb-verify --exclude scarb-execute
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_NIGHTLY_TOOLCHAIN }}
+      - name: run nightly tests
+        run: cargo test --doc -p scarb-prove -p scarb-verify -p scarb-execute
```
DelevoXDG, line 117: I'm a bit confused, do we or do we not currently use nightly for `scarb-execute`? 🤔

I'm asking because here it seems we do, but in other places (jobs in ci.yml, release workflows) we still treat only `verify` and `prove` as nightly

utils/scarb-extensions-cli/src/execute/mod.rs
```diff
@@ -1,9 +1,9 @@
 #[cfg(feature = "execute")]
-mod checked;
+mod for_execution;
 
 #[cfg(feature = "execute")]
-pub use checked::*;
+pub use for_execution::*;
 
 // TODO(maciektr): remove when stwo can use the same starknet-types-core version as Cairo.
```
DelevoXDG: The version of `starknet-types-core` used by stwo appears to be the same

So should we make the merge now, or do we want to leave it as is for now in case it diverges again? Just so we can clarify at which point can this change be made 🤔

utils/scarb-extensions-cli/src/execute/for_proving.rs
```diff
@@ -106,11 +106,11 @@ pub struct RunArgs {
     pub arguments: ProgramArguments,
 
     /// Desired execution output, either default Standard or CairoPie
-    #[arg(long)]
-    pub output: Option<OutputFormat>,
+    #[arg(long, default_value_t = OutputFormat::Standard)]
+    pub output: OutputFormat,
```
DelevoXDG: Default seems to be specified at `OutputFormat` level already - should we specify it just in one place?

utils/scarb-extensions-cli/src/execute/mod.rs
DelevoXDG: The main reason I'm asking is because it gets harder to keep track of differences between `for_execution` and `for_proving` variants

maciektr: Those two have different defaults - for execution, you run with standalone target and no output by default ; for proving, with bootloader target and standard output.

Cargo.toml
maciektr: Just a rebase going off, thanks for spotting!

utils/scarb-extensions-cli/src/execute/mod.rs
DelevoXDG: I think we should be able to address it without duplicating everything else - by making common `RunArgsBase` and flattening it in `RunArgsExecute` and `RunArgsProve`, which would have their own output format and target fields

utils/scarb-extensions-cli/src/execute/for_proving.rs
maciektr: Refactored.

utils/scarb-extensions-cli/src/execute/mod.rs
maciektr: Yep, refactored.

.github/workflows/ci.yml
maciektr: The execute logic does not rely on any unstable rust code. It's built with the stwo nightly code hidden behind a feature flag. But the doc tests seem to be built with the flags enabled, thus the change.

extensions/scarb-execute/tests/build.rs
```diff
@@ -200,14 +193,15 @@ fn can_print_panic_reason() {
         .arg("execute")
         .arg("--print-program-output")
         .arg("--print-resource-usage")
```
FroyaTheHen:
```suggestion

```

extensions/scarb-prove/Cargo.toml
```diff
@@ -10,22 +10,27 @@ readme.workspace = true
...
-stwo_cairo_prover.workspace = true
-stwo-cairo-adapter.workspace = true
+stwo-cairo-adapter = {workspace = true}
+stwo_cairo_prover = {workspace = true }
 
 [dev-dependencies]
 assert_fs.workspace = true
+predicates.workspace = true
 scarb-test-support = { path = "../../utils/scarb-test-support" }
 snapbox.workspace = true
```
FroyaTheHen:
```suggestion
```

extensions/scarb-prove/tests/heavy.rs
```diff
@@ -0,0 +1,90 @@
...
+            "#})
+        .lib_cairo(indoc! {r#"
+            #[executable]
+            fn main() -> felt252 {
+                42
+            }
+        "#})
+        .build(&t);
+    t
+}
```
FroyaTheHen: the code is repeated couple of times, maybe it can be moved to some common file and reused instead

maciektr: Moved to test-support crate

PR #2892, opened 2025-12-03
Support proc macro crate fingerprint in PMS

Draggu:
https://github.com/software-mansion/cairols/issues/775

### DIFF
--- a/scarb/src/compiler/plugin/collection.rs
+++ b/scarb/src/compiler/plugin/collection.rs
@@ -10,6 +10,7 @@ use smol_str::SmolStr;
 use std::vec::IntoIter;
 
 use super::proc_macro::{DeclaredProcMacroInstances, ProcMacroHostPlugin, ProcMacroInstance};
+use crate::compiler::incremental::PluginFingerprint;
 use crate::compiler::plugin::CairoPlugin;
 use crate::core::PackageId;
 use crate::{
@@ -27,12 +28,17 @@ impl PluginsForComponents {
     pub fn collect(workspace: &Workspace<'_>, unit: &CairoCompilationUnit) -> Result<Self> {
         let mut plugins = collect_builtin_plugins(workspace, unit)?;
 
-        let proc_macros = collect_proc_macros(workspace, unit)?
+        let proc_macros = collect_proc_macros(workspace, unit, false)?
             .into_iter()
             .map(|(component_id, instances)| {
                 Ok((
                     component_id,
-                    ComponentProcMacroHost::try_from_instances(instances)?,
+                    ComponentProcMacroHost::try_from_instances(
+                        instances
+                            .into_iter()
+                            .map(|(instance, _hash)| instance)
+                            .collect(),
+                    )?,
                 ))
             })
             .collect::<Result<HashMap<_, _>>>()?;
@@ -65,6 +71,8 @@ pub struct WorkspaceProcMacros {
     /// each mapped to a [`ProcMacroHostPlugin`] which contains
     /// **all proc macro dependencies of the package** collected from **all compilation units it appears in**.
     pub macros_for_components: HashMap<CompilationUnitComponent, Arc<Vec<ProcMacroHostPlugin>>>,
+    /// Mapping from package id to fingerprint hash.
+    pub instance_to_hash: HashMap<PackageId, u64>,
 }
 
 impl WorkspaceProcMacros {
@@ -76,7 +84,7 @@ impl WorkspaceProcMacros {
         let mut macros_for_components = HashMap::<_, Vec<_>>::new();
 
         for &unit in compilation_units {
-            for (component_id, mut macro_instances) in collect_proc_macros(workspace, unit)? {
+            for (component_id, mut macro_instances) in collect_proc_macros(workspace, unit, true)? {
                 let component: CompilationUnitComponent = unit
                     .components
                     .iter()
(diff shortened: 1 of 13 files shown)


utils/scarb-proc-macro-server-types/src/methods/mod.rs
```diff
@@ -24,6 +24,8 @@ pub struct ProcMacroResult {
     pub diagnostics: Vec<DiagnosticV2>,
     /// A list of code mappings produced by the macro
     pub code_mappings: Option<Vec<CodeMapping>>,
+    /// A proc macro fingerprint
+    pub fingerprint: Option<u64>,
```
piotmag769: Can't be `None` in real use cases. `Default` is only for tests I believe, no?

scarb/src/compiler/plugin/collection.rs
```diff
@@ -65,6 +72,7 @@ pub struct WorkspaceProcMacros {
     /// each mapped to a [`ProcMacroHostPlugin`] which contains
     /// **all proc macro dependencies of the package** collected from **all compilation units it appears in**.
     pub macros_for_components: HashMap<CompilationUnitComponent, Arc<Vec<ProcMacroHostPlugin>>>,
+    pub instance_to_hash: HashMap<Arc<ProcMacroInstance>, u64>,
```
piotmag769: Doc pls

scarb/src/compiler/plugin/collection.rs
```diff
@@ -189,12 +208,15 @@ impl PluginSuiteAssembler {
...
 
+type InstanceAndHash = (Arc<ProcMacroInstance>, Option<u64>);
+
 /// Collects [`ProcMacroInstances`]s for each component of the [`CairoCompilationUnit`],
 /// according to the dependencies on procedural macros.
 fn collect_proc_macros(
     workspace: &Workspace<'_>,
     unit: &CairoCompilationUnit,
-) -> Result<HashMap<CompilationUnitComponentId, Vec<Arc<ProcMacroInstance>>>> {
+    rt: Option<&Runtime>,
```
piotmag769: This method is called in chain from `get_workspace_macros` which has `Config` arg which has `tokio_handle` method

scarb/src/compiler/plugin/collection.rs
```diff
@@ -215,7 +237,15 @@ fn collect_proc_macros(
...
             }
 
-            component_proc_macro_instances.push(plugin.instantiate(workspace.config())?);
+            let hash = match rt {
+                Some(rt) => Some(
+                    rt.block_on(PluginFingerprint::try_from_plugin(plugin, workspace))?
+                        .digest_u64(),
+                ),
+                None => None,
+            };
```
piotmag769: This is ugly and creates a mess on the caller side, just create a wrapper method around `collect_proc_macros` at this point

scarb/src/compiler/plugin/proc_macro/instance.rs
```diff
@@ -30,6 +31,18 @@ pub struct ProcMacroInstance {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.package_id == other.package_id
+    }
+}
+impl Eq for ProcMacroInstance {}
+impl Hash for ProcMacroInstance {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.package_id.hash(state);
+    }
+}
```
maciektr: Why not just index `instance_to_hash` via `PackageId` instead? 🤔

scarb/src/compiler/plugin/collection.rs
```diff
@@ -74,9 +82,12 @@ impl WorkspaceProcMacros {
         compilation_units: &[&CairoCompilationUnit],
     ) -> Result<Self> {
         let mut macros_for_components = HashMap::<_, Vec<_>>::new();
+        let rt = Runtime::new()?;
```
maciektr: use `config.tokio_handle()`

scarb/src/compiler/plugin/collection.rs
maciektr: You can just do `workspace.config().tokio_handle()`

scarb/src/ops/proc_macro_server/methods/expand_derive.rs
```diff
@@ -60,21 +59,25 @@ impl Handler for ExpandDerive {
...
+                    &proc_macro_instance,
+                    hash,
                     current_width,
                     expansion,
                     call_site.clone(),
                     item.clone(),
                 ),
             }?;
 
+            result.fingerprint.hash(&mut hasher);
```
maciektr, line 83: What's the point of this? 🤔

utils/scarb-proc-macro-server-types/src/methods/mod.rs
Draggu: New LS + old Scarb requires this to work, also useful for tests and default

scarb/src/compiler/plugin/collection.rs
Draggu: simplified

scarb/src/compiler/plugin/proc_macro/instance.rs
Draggu: good call

scarb/src/ops/proc_macro_server/methods/expand_derive.rs
Draggu: We are expanding all derives at once here, so we need to combine hashes of all plugins used here

utils/scarb-proc-macro-server-types/src/methods/mod.rs
maciektr: Why would this happen though - LS is tied with Scarb so we can make breaking changes here, right?

scarb/src/compiler/plugin/proc_macro/instance.rs
```diff
@@ -30,6 +31,12 @@ pub struct ProcMacroInstance {
     expansions: Vec<Expansion>,
 }
 
+impl PartialEq for ProcMacroInstance {
+    fn eq(&self, other: &Self) -> bool {
+        self.package_id == other.package_id
+    }
+}
+
```
maciektr: Do you still need this? 🤔

Draggu: artefact, removed

PR #2895, opened 2025-12-05
Bump mdbook to 0.5

maciektr:
- **fix lock**
- **Bump mdbook**
- **Remove deprecated book.multilingual**
- **Update mdbook test expectations**


extensions/scarb-doc/src/docs_generation/markdown/book_toml.rs
```diff
@@ -8,7 +8,6 @@ pub fn generate_book_toml_content(package_metadata: &AdditionalMetadata) -> Stri
             [book]
             authors = {:?}
             language = "en"
-            multilingual = false
```
FroyaTheHen, line 11: why was it here in the first place?

maciektr: Not sure, seems it has been added as future proofing and never used. https://github.com/rust-lang/mdBook/pull/2775

PR #2903, opened 2025-12-11
Add functions debug info

piotmag769:


maciektr (approved):

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -499,6 +499,10 @@ pub struct TomlCairo {
     /// Used by [cairo-coverage](https://github.com/software-mansion/cairo-coverage).
     /// This feature is unstable and is subject to change.
     pub unstable_add_statements_code_locations_debug_info: Option<bool>,
+    /// Add additional information about sierra functions to debug info.
+    /// Used by [cairo-debugger](https://github.com/software-mansion-labs/cairo-debugger).
+    /// This feature is unstable and is subject to change.
+    pub add_functions_debug_info: Option<bool>,
```
maciektr, line 505: Wait, shouldn't this have the `unstable_` prefix? 
```suggestion
    pub unstable_add_functions_debug_info: Option<bool>,
```

piotmag769: Idk, if I add `unstable` we are left with `unstable` for the rest of our lives

maciektr: It's easy to migrate - just add an alias :) 

Should we drop unstable from the other too as well? 🤔

piotmag769: Yep, but then there is a deprecation period and all that. I think we should, they are quite stable

maciektr: I mean, you can accept both `unstable_add_functions` and `add_functions` and never care about deprecating the former, while the latter is added.

piotmag769: I understood, but idk, I don't like the long name xD

maciektr: I'm fine with leaving this without `unstable_`.

PR #2904, opened 2025-12-12
Require no path in top-level inline macros

maciektr:
Add failing test case

Require no path in top-level inline macros

### DIFF
--- a/scarb/src/compiler/plugin/proc_macro/v2/host/mod.rs
+++ b/scarb/src/compiler/plugin/proc_macro/v2/host/mod.rs
@@ -214,6 +214,12 @@ impl MacroPlugin for ProcMacroHostPlugin {
 
         // Expand module-level inline macro.
         if let ast::ModuleItem::InlineMacro(inline_macro) = &item_ast
+            && inline_macro
+                .path(db)
+                .segments(db)
+                .elements(db)
+                .exactly_one()
+                .is_ok()
             && let Some(result) =
                 expand_module_level_inline_macro(self, db, inline_macro, &stream_metadata)
         {


maciektr:

- #2904 ◀

scarb/src/compiler/plugin/proc_macro/v2/host/mod.rs
```diff
@@ -214,6 +214,12 @@ impl MacroPlugin for ProcMacroHostPlugin {
 
         // Expand module-level inline macro.
         if let ast::ModuleItem::InlineMacro(inline_macro) = &item_ast
+            && inline_macro
+                .path(db)
+                .segments(db)
+                .elements(db)
+                .exactly_one()
+                .is_ok()
```
DelevoXDG, line 222: Rather than skipping it, should we perhaps handle this explicitly, and return error similar to one that Cairo emits?
```
error: Inline macro `not::a::path::some` not found.
```
or could modify it slightly, to inform that module path macros are not supported

piotmag769: There could very well exist a valid declarative macro under that path - we must not error on this

maciektr: @DelevoXDG as Piotrek said, we don't intend to limit the possible semantics that could be introduced by the compiler - I'd rather let them throw the error instead. From our pov, we only expand macros that we see and macro with malformed path is just something we don't know how to expand on.

DelevoXDG (approved):

PR #2905, opened 2025-12-12
Add test case for derive with invalid path

maciektr:


scarb/tests/proc_macro_v2_expand_derive.rs
```diff
@@ -609,3 +609,97 @@ fn can_use_two_derive_macros() {
...
+                let token_stream = TokenStream::new(vec![TokenTree::Ident(Token::new(
+                  code.clone(),
+                    TextSpan {
+                        start: 0,
+                        end: code.len() as u32,
+                    },
+                ))]);
+
+                ProcMacroResult::new(token_stream)
+            }
```
DelevoXDG, line 624: Nit:
since we're not actually using the macro, maybe we can pick the simplest and shortest macro possible?

maciektr: True, changed!

PR #2907, opened 2025-12-15
add arg remote-base-repo to scarb doc command

FroyaTheHen:


utils/scarb-extensions-cli/src/doc.rs
```diff
@@ -52,4 +52,8 @@ pub struct Args {
     /// Logging verbosity.
     #[command(flatten)]
     pub verbose: VerbositySpec,
+
+    /// Base URL of a remote repository. Used for generating links to source code.
+    #[arg(long, env = "REMOTE_BASE_URL")]
+    pub remote_base_url: Option<String>,
```
maciektr, line 58: Should we auto-derive the base url from manifests `repository` field and some vcs info? 🤔 

I would see three options here: 

1) If `repository` url in the manifest file is declared, use it as base. Read commit hash similar to https://github.com/software-mansion/scarb/blob/b5536486a7e830b973aad76d0fe4ca1ea51225ef/scarb/src/ops/package.rs#L364-L364 and use it. Would require some best-effort approach for different domains (github vs gitlab vs bitbucket etc). 

2) If user specifies override via SCARB_REMOTE_BASE_URL or --remote-base-url, use this instead. 

3) Add some option to skip generating link entirely, e.g. by --remote-url-base=none or some manifest specification 🤔 

(can be done in a separate pr)

LMKWYT

FroyaTheHen: I do think that this would be a nice improvement. However, I foresee issues. Such as for example to build the full url for an item we must make assumptions such as (for example) that the current branch is the one that is desired to be included in the url with no regards for its stability.  


> Add some option to skip generating link entirely, e.g. by --remote-url-base=none or some manifest specification 🤔

I'm just thinking that maybe by default the linking should be off even if manifest specifies the repo url. Otherwise this would generate potentially unwanted changes for existing packages. But I'm not sure since that sounds a bit like an extra complication. Any opinion?   

Also, this was already partially discussed see[slack thread](https://swmansion.slack.com/archives/C06UPCDPZD3/p1765437204421309). 

Then, created a [task](https://github.com/software-mansion/scarb/issues/2929) and I'll dig into this more :)

maciektr: > the current branch is the one that is desired to be included in the url

I guess it's the other way round. It must be the current branch, or otherwise you risk generating invalid links, as the code you are generating docs for (current branch) may be completely different than what urls target (remote base url branch) 🤔 

> Then, created a task and I'll dig into this more :)

Sounds like a good idea!

PR #2906, opened 2025-12-15
add source code link in generated documentation

FroyaTheHen:

PR #2909, opened 2025-12-15
document scarb doc --remote-base-url

FroyaTheHen:
**Stack**:
- #2907
- #2906

### DIFF
--- a/website/docs/extensions/documentation-generation.md
+++ b/website/docs/extensions/documentation-generation.md
@@ -34,6 +34,24 @@ Currently, we support only those types of links:
 
 - `[ItemName]` and ``[`ItemName`]`` (where `ItemName` is a valid path to an item).
 
+## Linking to the source code GitHub repository
+
+Use flag `--remote-base-url` or set `SCARB_DOC_REMOTE_BASE_URL` environmental variable to link to the source code of the package in the GitHub repository. Each item subpage will be linked to the corresponding source code file.
+For example:
+
+```shell
+scarb doc --remote-base-url=https://github.com/ExampleRepoOwner/ExampleRepoProject/blob/example_branch/
+```
+
+Scarb constructs urls combining: the remote base url, the relative path to the file from the package root, and line anchors when relevant. It does not check whether the links are valid or if they were resolved correctly. It is up to the user to provide a correct base url and verify the results.
+Example url for an item in a workspace package named `hello_world` with a relative path `src/lib.cairo`, line anchor start `10`, end `15`, and `SCARB_DOC_REMOTE_BASE_URL` like above, will be:
+
+```
+https://github.com/ExampleRepoOwner/ExampleRepoProject/blob/example_branch/hello_world/src/lib.cairo#L10-L15
+```
+
+Note, it's relevant for Markdown output only.
+
 ## mdBook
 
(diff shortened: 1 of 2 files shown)


website/docs/extensions/documentation-generation.md
```diff
@@ -34,6 +34,24 @@ Currently, we support only those types of links:
...
+```
+
+Scarb constructs urls combining: the remote base url, the relative path to the file from the package root, and line anchors when relevant. It does not check whether the links are valid or if they were resolved correctly. It is up to the user to provide a correct base url and verify the results.
+Example url for an item in a workspace package named `hello_world` with a relative path `src/lib.cairo`, line anchor start `10`, end `15`, and `REMOTE_BASE_URL` like above, will be:
+
+```
+https://github.com/ExampleRepoOwner/ExampleRepoProject/blob/example_branch/hello_world/src/lib.cairo#L10-L15
+```
+
+Note, it's relevant for Markdown output only.
```
maciektr, line 53: I feel like we should bail with an error if both output_format=json and remove_base_url are defined 🤔 
WDYT?

website/docs/extensions/documentation-generation.md
```diff
@@ -34,6 +34,24 @@ Currently, we support only those types of links:
...
+
+Use flag `--remote-base-url` or set `REMOTE_BASE_URL` environmental variable to link to the source code of the package in the GitHub repository. Each item subpage will be linked to the corresponding source code file.
+For example:
+
+```shell
+scarb doc --remote-base-url=https://github.com/ExampleRepoOwner/ExampleRepoProject/blob/example_branch/
+```
+
+Scarb constructs urls combining: the remote base url, the relative path to the file from the package root, and line anchors when relevant. It does not check whether the links are valid or if they were resolved correctly. It is up to the user to provide a correct base url and verify the results.
+Example url for an item in a workspace package named `hello_world` with a relative path `src/lib.cairo`, line anchor start `10`, end `15`, and `REMOTE_BASE_URL` like above, will be:
```
maciektr:
```suggestion
Example url for an item in a workspace package named `hello_world` with a relative path `src/lib.cairo`, line anchor start `10`, end `15`, and `SCARB_DOC_REMOTE_BASE_URL` like above, will be:
```

website/docs/extensions/documentation-generation.md
```diff
@@ -34,6 +34,24 @@ Currently, we support only those types of links:
 
 - `[ItemName]` and ``[`ItemName`]`` (where `ItemName` is a valid path to an item).
 
+## Linking to the source code GitHub repository
+
+Use flag `--remote-base-url` or set `REMOTE_BASE_URL` environmental variable to link to the source code of the package in the GitHub repository. Each item subpage will be linked to the corresponding source code file.
```
maciektr:
```suggestion
Use flag `--remote-base-url` or set `SCARB_DOC_REMOTE_BASE_URL` environmental variable to link to the source code of the package in the GitHub repository. Each item subpage will be linked to the corresponding source code file.
```

website/docs/extensions/documentation-generation.md
FroyaTheHen: hmm, yeah, that makes sense 👌 will do

PR #2911, opened 2025-12-15
Drop unstable_ prefix from mapping keys in manifest, keep aliases

maciektr:


scarb/tests/metadata.rs
```diff
@@ -1877,8 +1877,57 @@ fn compiler_config_collected_properly() {
...
+fn compiler_config_aliases_collected_properly() {
+    let t = TempDir::new().unwrap();
+    ProjectBuilder::start()
+        .name("hello")
+        .version("0.1.0")
+        .manifest_extra(indoc! {r#"
+         [profile.dev.cairo]
+         sierra-replace-ids = false
+         unstable-add-statements-code-locations-debug-info = false
+         add-statements-code-locations-debug-info = true
```
piotmag769, line 1901: This doesn't test unstables in the vacuum which is the usual use case

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -1346,17 +1348,29 @@ impl TomlManifest {
             if let Some(enable_gas) = cairo.enable_gas {
                 compiler_config.enable_gas = enable_gas;
             }
-            if let Some(unstable_add_statements_functions_debug_info) =
+            if let Some(add_statements_functions_debug_info) =
```
piotmag769, line 1356: Separate blocks of `x + unstable_{x}` with newlines please for readability

scarb/src/core/manifest/toml_manifest.rs
```diff
@@ -492,12 +492,14 @@ pub struct TomlCairo {
     /// statement to be generated and all functions that were inlined or generated along the way.
     /// Used by [cairo-profiler](https://github.com/software-mansion/cairo-profiler).
     /// This feature is unstable and is subject to change.
+    pub add_statements_functions_debug_info: Option<bool>,
```
piotmag769, line 495: Doc what takes precedence

scarb/tests/metadata.rs
piotmag769, line 1: We should migrate all the other tests to not use `unstable` anymore

scarb/tests/metadata.rs
maciektr: It does, the `add-statements-functions-debug-info` is only defined via `unstable-`.

PR #2936, opened 2026-01-12
Add a test case for registry dep with transient dev dep

maciektr:


scarb/tests/http_registry.rs
```diff
@@ -406,6 +406,50 @@ fn caching() {
     expected.assert_eq(&registry.logs());
 }
 
+#[test]
+fn transitive_dev_dependency_not_pulled() {
```
maciektr, line 410: This test should fail, as we do not recognize the dependency kind of dependencies of packages from the registry, but it is passing instead. 
This is caused by the fact, that all uploads to the registry that happens in Scarb tests go through `LocalRegistryClient`, even if we test the `HttpRegistryClient`. The logic for uploads diverged between the two, as local client handle creating registry records for the new package through Scarb (see https://github.com/software-mansion/scarb/blob/8300b96c471c7da3ac4a4b84d167b44b046089f5/scarb/src/core/registry/client/local.rs#L177) and http client handles it on the backend (see https://github.com/software-mansion/scarb/blob/8300b96c471c7da3ac4a4b84d167b44b046089f5/scarb/src/core/registry/client/http.rs#L180). 
The test passes, as local client does not include dev dependencies in the publish, while real registry does. This is fixed in a subsequent PR (see https://github.com/software-mansion/scarb/pull/2937/changes/6de0ba1e42f11d51674c0ffb3d727cb25f520ff4).

integraledelebesgue (approved):

PR #2942, opened 2026-01-13
add documentation linking if package repository given in manifest

FroyaTheHen:


extensions/scarb-doc/src/linking.rs
```diff
@@ -0,0 +1,84 @@
...
+                commit_hash,
+                repository_url,
+            } => {
+                let relative_path_str = file_path.strip_prefix(repo_root).ok()?.to_str()?;
+
+                Some(format!(
+                    "<a href='{repository_url}/blob/{commit_hash}/{relative_path_str}{postfix}'> [source code] </a>"
+                ))
+            }
+            RemoteDocLinkingData::Disabled => None, // TODO: add option to disable explicitly
```
wawel37: Track the TODO with issue plz and link it here

FroyaTheHen: this _todo_ is done and the comment removed in the stack ([here](https://github.com/software-mansion/scarb/pull/2944/changes#diff-64af2fa7c1504facfeb1a41e73701336032262b46fe07e5fabf00b99c2007726:~:text=//%20TODO%3A%20add%20option%20to%20disable%20explicitly))

extensions/scarb-doc/tests/source_code_linking.rs
```diff
@@ -215,3 +216,77 @@ fn special_characters_in_base_url_are_not_malformed() {
...
+    );
+
+    // assert links are correct
+    let pairs = vec![
+        ("hello_world.md", "hello_world/src/lib.cairo"),
+        ("hello_world-main.md", "hello_world/src/lib.cairo#L1-L3"),
+    ];
+
+    let expected_base_url =
+        format!("https://github.com/ExampleRepoOwner/ExampleRepoProject/blob/{commit_hash}/");
```
Arcticae, line 279: The implementation doesn't check if it's a pushed commit, so the reference can be invalid if the scenario would be like this:
1. I generate the docs 
2. I commit 
3. I expected the docs to refer to the commited ref, but it's the previous one.

extensions/scarb-doc/src/linking.rs
```diff
@@ -0,0 +1,94 @@
...
+        repo_root: String,
+        commit_hash: String,
+        repository_url: String,
+    },
+    /// Result of a lacking, faulty specification or (not yet implemented) explicitly disabled.
+    Disabled,
+}
+
+impl RemoteDocLinkingData {
+    pub fn from(
```
Arcticae: This is weird, i'd rather have 3 different constructors for each variant separate one

extensions/scarb-doc/src/linking.rs
```diff
@@ -0,0 +1,94 @@
...
+        remote_base_url: Option<String>,
+        repo_root: Option<String>,
+        workspace_root: Option<String>,
+        commit_hash: Option<String>,
+        repository_url: Option<String>,
+    ) -> Self {
+        if remote_base_url.is_some()
```
Arcticae: you can use let chain here

Arcticae: or actually a pattern line you did below

Arcticae: (not needed if you implement 3 different constructors)

extensions/scarb-doc/src/linking.rs
```diff
@@ -0,0 +1,94 @@
...
+                repository_url,
+            }
+        } else {
+            RemoteDocLinkingData::Disabled
+        }
+    }
+
+    pub fn get_formatted_url(
+        &self,
+        offset: FileLinkDataLocationOffset,
```
Arcticae: This can be just a standard rust `Range`

extensions/scarb-doc/src/linking.rs
```diff
@@ -0,0 +1,94 @@
...
+    Disabled,
+}
+
+impl RemoteDocLinkingData {
+    pub fn from(
+        remote_base_url: Option<String>,
+        repo_root: Option<String>,
+        workspace_root: Option<String>,
+        commit_hash: Option<String>,
+        repository_url: Option<String>,
```
Arcticae: Why convert those to strings? I would rather have `Path`s here (you convert them before passing them here)

extensions/scarb-doc/src/main.rs
```diff
@@ -39,19 +40,36 @@ fn main_inner(args: Args, ui: Ui) -> Result<()> {
...
+                .map(str::to_owned);
 
-    if args.packages_filter.get_workspace() & !matches!(args.output_format, OutputFormat::Json) {
-        let mut builder = WorkspaceMarkdownBuilder::new(
-            args.output_format.into(),
-            workspace_root.clone(),
-            remote_base_url,
-        );
+            let commit_hash = repo.rev_parse_single("HEAD")?.to_string();
+            (repo_root, Some(commit_hash))
```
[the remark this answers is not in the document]
FroyaTheHen: already addressed further in the stack  [here](https://github.com/software-mansion/scarb/pull/2943/changes#:~:text=%2B-,let%20repo_root%20%3D%20repo,.map(str%3A%3Ato_owned)%3B,-50)

extensions/scarb-doc/tests/source_code_linking.rs
FroyaTheHen: If the last commit contains only docs changes then all the links will point to the correct snippets and that should be the case. If it's not, then the problem is as valid as a problem of providing a wrong link - this is not something that Scarb can be blamed for as it is the users responsibility to provide the correct data. (I do understand that the manifest driven base-url is less user engaging than explicit flag.)
However, what I see suitable here is exposing those details in documentation.

extensions/scarb-doc/src/linking.rs
FroyaTheHen: so `repo.workdir()` is of the type `&Path`, without changing the type we need to face lifetime resolving problem that can be done but results in rearranging the logic in `main.rs/ fn main_inner ` to something like: 
```
...
    let remote_base_url = args.remote_base_url.clone();

    match gix::discover(&workspace_root) {
        Ok(repo) => {
            let repo_root = repo
                .workdir();
            let commit_hash = repo.rev_parse_single("HEAD")?.to_string();
            some_helper_function(args, metadata_for_packages, metadata, ui, remote_base_url, repo_root, workspace_root, Some(commit_hash), output_dir,)
        }
        Err(_) => {
            some_helper_function(args, metadata_for_packages, metadata, ui, remote_base_url, None, workspace_root, None, output_dir)
        },
    }
} 
...
```
That just seems more messy to me than changing the type, but let me know if you don't agree. 
Other than that, this is code can probably still be improved so I'll update.

extensions/scarb-doc/tests/source_code_linking.rs
Arcticae: > If the last commit contains only docs changes

This isn't guaranteed though. We take the reference of the latest commit and we're not sure about user flow. Ideally user does it in CI but that's also not guaranteed.

We could detect dirty index in git, and then warn the user about this possibility. This warn can also be implemented later, but i can see a case where we generate the wrong links and user is frustrated when he gets the wrong docs (the deployer of the docs won't know until somebody complains and there is an actual discrepancy).

extensions/scarb-doc/src/linking.rs
Arcticae: We can just copy it if lifetime is a problem, we do it anyway 
Path's are easier to manipulate in general, but it's not crucial for this PR, just a nitpick

Arcticae: Also you can use or pass `PathBuf`

extensions/scarb-doc/src/linking.rs
```diff
@@ -0,0 +1,106 @@
...
+    repository_url: Option<String>,
+) -> RemoteDocLinkingData {
+    match (remote_base_url, repo_root, commit_hash, repository_url) {
+        (Some(remote_base_url), ..) => {
+            RemoteDocLinkingData::new_explicit(remote_base_url, workspace_root)
+        }
+        (None, Some(repo_root), Some(commit_hash), Some(repository_url)) => {
+            RemoteDocLinkingData::new_manifest(repo_root, commit_hash, repository_url)
+        }
+        _ => RemoteDocLinkingData::new_disabled(),
```
Arcticae: If commit hash is not present, we disable linking. (i.e. uncommited changes, before any commit happens in the repo). 

Is this desired behavior or just coincidental?

extensions/scarb-doc/tests/source_code_linking.rs
FroyaTheHen: https://github.com/software-mansion/scarb/issues/2957

extensions/scarb-doc/src/linking.rs
```diff
@@ -0,0 +1,127 @@
...
+                repo_root.clone(),
+                commit_hash.clone(),
+                repo_url.clone(),
+            ))
+        }
+        (_, _, None, Some(base_url)) => Ok(RemoteDocLinkingData::new_explicit(
+            base_url.clone(),
+            workspace_root.clone(),
+        )),
+        _ => Ok(RemoteDocLinkingData::new_disabled()),
```
FroyaTheHen, line 125: This is changed to fail in next PR in stack. In this PR the _disable_ is not yet implemented, so if this causes an error all tests that use json format would either fail or (if updated) cause a lot of unnecessary noise

extensions/scarb-doc/src/main.rs
```diff
@@ -67,25 +71,35 @@ fn main_inner(args: Args, ui: Ui) -> Result<()> {
         let mut output = match args.output_format {
             OutputFormat::Json => {
                 ensure!(
```
Arcticae, line 73: Shouldnt' we just warn here?

FroyaTheHen: We could, yet there is an analogical case above lines 28-31 where we do throw an error.

Issue #2929, opened 2026-01-07
Auto dervie repo url for linking in scarb doc


FroyaTheHen:
https://github.com/software-mansion/scarb/pull/2907#discussion_r2640040358

PR #2915, opened 2025-12-17
link macro call items source code

FroyaTheHen:
**Stack**:
- #2909
- #2907
- #2906

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

closes https://github.com/software-mansion/scarb/issues/2137

PR #2943, opened 2026-01-13
add option to explicitly disable linking documentation to remote repo

FroyaTheHen:
closes https://github.com/software-mansion/scarb/issues/2929

**Stack**:
- #2942
- #2915

### DIFF
--- a/utils/scarb-extensions-cli/src/doc.rs
+++ b/utils/scarb-extensions-cli/src/doc.rs
@@ -4,6 +4,7 @@
 
 use clap::Parser;
 use scarb_ui::args::{FeaturesSpec, PackagesFilter, VerbositySpec};
+use std::str::FromStr;
 
 /// CLI command name.
 pub const COMMAND_NAME: &str = "doc";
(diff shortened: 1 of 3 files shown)


wawel37 (commented):
imho it's a little bit confusing having this arg as a input param and the true/false at the same time. Maybe we could do something like this?:

- introduce `--disable-remote-linking` (or smth like this), and force scarb doc to not include the links
- make the `--remote-base-url` take only urls, not the `enabled` and `disabled` values. If user doesnt specify the remote url and the manifest doesnt have the repository link, we could just put a warn, that the link wont be generated as the user didn't specify the remote url in arg neither in the manifest

utils/scarb-extensions-cli/src/doc.rs
```diff
@@ -25,6 +26,27 @@ pub enum OutputFormat {
...
+    /// Enable linking documentation to source code.
+    Enabled(Option<String>),
+    /// Disable linking documentation to source code.
+    Disabled,
+}
+
+impl FromStr for RemoteBaseUrl {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
Arcticae, line 41: This doesn't error, does it?

FroyaTheHen: well, i can use `anyhow::Result` so its idiomatic, good call

PR #2954, opened 2026-01-20
scarb doc: enable/disable remote linking

FroyaTheHen:


extensions/scarb-doc/src/lib.rs
```diff
@@ -120,6 +120,7 @@ pub fn generate_package_information<'a>(
     workspace_root: &Utf8PathBuf,
     repo_root: &Option<PathBuf>,
     commit_hash: &Option<String>,
+    disable: bool,
```
Arcticae:
```suggestion
    disable_linking: bool,
```

extensions/scarb-doc/src/linking.rs
```diff
@@ -103,9 +105,22 @@ pub fn resolve_remote_linking_data(
     workspace_root: &Utf8PathBuf,
     repo_root: &Option<PathBuf>,
     commit_hash: &Option<String>,
+    disable: bool,
```
Arcticae:
```suggestion
    disable_linking: bool,
```

PR #2956, opened 2026-01-21
dupdate doc base repo linking

FroyaTheHen:
**Stack**:
- #2954
- #2942

### DIFF
--- a/website/docs/extensions/documentation-generation.md
+++ b/website/docs/extensions/documentation-generation.md
@@ -25,7 +25,7 @@ As for now, we support those types of comments:
 - `///` documentation for the following item.
 - `//!` documentation for enclosing item (also works with file modules).
 
-the `///` and `//!` comment prefixes are supported.
+The `///` and `//!` comment prefixes are supported.
 
 ## Item linkage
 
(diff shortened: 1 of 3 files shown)


website/docs/extensions/documentation-generation.md
```diff
@@ -34,23 +34,49 @@ Currently, we support only those types of links:
...
+
+Output format constraints:
+
+- Remote linking is supported only for Markdown output only.
+
+Requirements:
+
+- Linking requires either a manifest `repository` or an explicit `--remote-base-url`. If neither is configured and linking is not disabled, Scarb will error.
+
+Officially supported VCS providers:
```
Arcticae, line 76: We can elaborate on **why** they are supported here - we may support other ones but we're just not sure if it works ¯\_(ツ)_/¯ 
Also we don't detect which one you use, which can be pointed out here

website/docs/extensions/documentation-generation.md
```diff
@@ -34,23 +34,49 @@ Currently, we support only those types of links:
...
-## Linking to the source code GitHub repository
+## Linking to the source code VCS repository
 
-Use flag `--remote-base-url` or set `SCARB_DOC_REMOTE_BASE_URL` environmental variable to link to the source code of the package in the GitHub repository. Each item subpage will be linked to the corresponding source code file.
-For example:
+You can add "View source" links in generated docs that point to your VCS repository. There are two ways to enable linking:
+
+#### 1. Use the package manifest [`repository`](../reference/manifest.md#repository) field
+
+If your package manifest `Scarb.toml` specifies a [`repository`](../reference/manifest.md#repository) that points to a VCS project, Scarb can compose links automatically. This requires running `scarb doc` command in a VCS repository that Scarb can discover. If Git repository discovery fails, remote linking is disabled for this package.
```
Arcticae:
```suggestion
If your package manifest `Scarb.toml` specifies a [`repository`](../reference/manifest.md#repository) that points to a VCS project website, Scarb can compose links automatically. This requires running `scarb doc` command in a git repository that Scarb can discover. If Git repository discovery fails, remote linking is disabled for this package.
```

website/docs/extensions/documentation-generation.md
```diff
@@ -34,23 +34,49 @@ Currently, we support only those types of links:
...
+
+#### 2. Explicit base URL (CLI/env)
+
+Provide a VCS browsing URL via flag `--remote-base-url` or `SCARB_DOC_REMOTE_BASE_URL` environmental variable. Scarb will append the file path (relative to the workspace root) and, when available, a VCS line range anchor.
```
Arcticae:
```suggestion
Provide a VCS project website via flag `--remote-base-url` or `SCARB_DOC_REMOTE_BASE_URL` environmental variable. Scarb will append the file path (relative to the workspace root) and, when available, a VCS line range anchor.
```

website/docs/extensions/documentation-generation.md
```diff
@@ -34,23 +34,49 @@ Currently, we support only those types of links:
...
 ```
 https://github.com/ExampleRepoOwner/ExampleRepoProject/blob/example_branch/hello_world/src/lib.cairo#L10-L15
 ```
 
-Note, it's relevant for Markdown output only.
+Note that Scarb does not validate that the link targets exist. You must provide a correct base URL and verify the results yourself.
+
+Precedence and disabling:
+
+- If both a manifest `repository` and `--remote-base-url` are provided, the manifest `repository` takes precedence for link generation.
```
Arcticae: Isn't usually command line input more important? What's the convention in scarb?

wawel37: I also don't know what's the standard in Scarb for these cases, but for me it seems counterintuitive for manifest `repository` taking priority over explicite specified url in the arg

FroyaTheHen: True! That is indeed the case in Scarb and Cargo as well. And I agree. It makes much more sense that the flag takes precedence over manifest. I'll update accordingly.

website/docs/extensions/documentation-generation.md
FroyaTheHen: see updated 👇

PR #2960, opened 2026-01-23
scarb doc: dispose struct FileLinkData

FroyaTheHen:


extensions/scarb-doc/src/types/item_data.rs
```diff
@@ -185,50 +188,28 @@ where
     }
 }
 
-fn get_file_link_data<'db>(
+pub fn get_file_and_location_in_file_data<'db>(
```
integraledelebesgue: "Data" in the name isn't adequate anymore

```suggestion
pub fn get_file_and_location<'db>(
```

PR #2980, opened 2026-02-05
Test links

maciektr:
test


.lycheeignore
```diff
@@ -0,0 +1,6 @@
+file://
+https://127\.0\.0\.1/
+https://github\.com/ExampleRepoOwner/ExampleRepoProject*
+https://github\.com/software-mansion/scarbs\.xyz
+https://stackoverflow.com/
+https://this-registry-does-not-exist/
```
wawel37, line 6: nit: maybe we can just ignore all `tests` directories?

maciektr: This come from the docs though, tests are ignored with `--exclude-path` cli args.

PR #2983, opened 2026-02-05
Support runnable code blocks in docs

### LINKED ISSUE #2473: Open scarb doc examples in cairo playground
In doc, you can add code snippets to docstrings, like this: 
```cairo
/// Function that prints "test" to stdout with endline.
/// Can invoke it like that:
/// ```cairo
///     fn main() {
///         test();
///     }
/// ```
pub fn test() {
    println!("test");
}
```
Which are renedered in the generated mdbook 

Image

As of this task, we should enable users to execute such snippets and show the outputs. 

Hint: research possibility to use [cairo playground](https://www.cairo-lang.org/cairovm/)

### LINKED ISSUE #2821: Define interface for distinguishing doc tests / runnable examples
Also, define interface for skipping doc tests etc

maciektr:
Support runnable code blocks in docs

Add test case with starknet-contract target

Add test cases for mdx and json outputs

Add support for should_panic and compile_fail doc-test attributes

### DIFF
--- a/extensions/scarb-doc/src/doc_test/code_blocks.rs
+++ b/extensions/scarb-doc/src/doc_test/code_blocks.rs
@@ -1,0 +1,278 @@
+use crate::doc_test::runner::RunStrategy;
+use crate::types::crate_type::Crate;
+use crate::types::module_type::Module;
+use cairo_lang_doc::parser::DocumentationCommentToken;
+use itertools::Itertools;
+use std::collections::HashMap;
+use std::str::from_utf8;
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
+pub struct CodeBlockId {
+    pub item_full_path: String,
+    pub close_token_idx: usize,
+    /// Index of this block in the item's documentation.
+    pub block_index: usize,
+}
+
+impl CodeBlockId {
+    pub fn new(item_full_path: String, block_index: usize, close_token_idx: usize) -> Self {
+        Self {
+            item_full_path,
+            block_index,
+            close_token_idx,
+        }
+    }
+
+    // TODO: (#2888): Display exact code block location when running doc-tests
+    pub fn display_name(&self, total_blocks_in_item: usize) -> String {
+        if total_blocks_in_item <= 1 {
+            self.item_full_path.clone()
+        } else {
+            format!("{} (example {})", self.item_full_path, self.block_index + 1)
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum CodeBlockAttribute {
+    Cairo,
+    Runnable,
+    Ignore,
+    NoRun,
+    CompileFail,
+    ShouldPanic,
+    Other(String),
+}
+
+impl From<&str> for CodeBlockAttribute {
+    fn from(string: &str) -> Self {
+        match string.to_lowercase().as_str() {
+            "cairo" => CodeBlockAttribute::Cairo,
+            "runnable" => CodeBlockAttribute::Runnable,
+            "ignore" => CodeBlockAttribute::Ignore,
+            "no_run" => CodeBlockAttribute::NoRun,
+            "should_panic" => CodeBlockAttribute::ShouldPanic,
+            "compile_fail" => CodeBlockAttribute::CompileFail,
+            _ => CodeBlockAttribute::Other(string.to_string()),
+        }
+    }
+}
+
+/// Represents code block extracted from doc comments.
+#[derive(Debug, Clone, PartialEq)]
+pub struct CodeBlock {
+    pub id: CodeBlockId,
+    pub content: String,
+    pub attributes: Vec<CodeBlockAttribute>,
+}
+
+impl CodeBlock {
+    pub fn new(id: CodeBlockId, content: String, info_string: &str) -> Self {
+        let attributes = Self::parse_attributes(info_string);
+        Self {
+            id,
+            content,
+            attributes,
+        }
+    }
+
+    fn parse_attributes(info_string: &str) -> Vec<CodeBlockAttribute> {
+        info_string
+            .split(',')
+            .map(|attr| attr.trim())
+            .filter(|attr| !attr.is_empty())
+            .dedup()
+            .map(Into::into)
+            .collect()
+    }
+
+    // TODO(#3096): default to Cairo unless specified otherwise
+    fn is_cairo(&self) -> bool {
+        self.attributes.contains(&CodeBlockAttribute::Cairo)
+    }
+
+    pub fn run_strategy(&self) -> RunStrategy {
+        if self.attributes.contains(&CodeBlockAttribute::Ignore) {
+            return RunStrategy::Ignore;
+        }
+        // TODO: drop the `runnable` attribute requirement; default to runnable for Cairo blocks
+        if !self.is_cairo() || !self.attributes.contains(&CodeBlockAttribute::Runnable) {
+            return RunStrategy::Ignore;
+        }
+        if self.attributes.contains(&CodeBlockAttribute::NoRun)
+            || self.attributes.contains(&CodeBlockAttribute::CompileFail)
+        {
+            return RunStrategy::Build;
+        }
+        RunStrategy::Execute
+    }
+}
+
+pub fn collect_code_blocks(crate_: &Crate<'_>) -> Vec<CodeBlock> {
+    let mut runnable_code_blocks = Vec::new();
+    collect_from_module(&crate_.root_module, &mut runnable_code_blocks);
+    for module in &crate_.foreign_crates {
+        collect_from_module(module, &mut runnable_code_blocks);
+    }
+    runnable_code_blocks.sort_by_key(|block| block.id.clone());
+    runnable_code_blocks.dedup_by_key(|block| block.id.clone());
+    runnable_code_blocks
+}
+
+/// Counts the number of code blocks per documented item. Used to generate display names
+/// for code blocks, allowing to distinguish between multiple code blocks in the same item.
+///
+/// Returns the mapping from `item_full_path` to the number of code blocks in that item.
+pub fn count_blocks_per_item(code_blocks: &[CodeBlock]) -> HashMap<String, usize> {
+    let mut counts = HashMap::new();
+    for block in code_blocks {
+        *counts.entry(block.id.item_full_path.clone()).or_insert(0) += 1;
+    }
+    counts
+}
+
+fn collect_from_module(module: &Module<'_>, runnable_code_blocks: &mut Vec<CodeBlock>) {
+    for &item_data in module.get_all_item_ids().values() {
+        for block in &item_data.code_blocks() {
+            runnable_code_blocks.push(block.clone());
+        }
+    }
+    for &item_data in module.pub_uses.get_all_item_ids().values() {
+        for block in &item_data.code_blocks() {
+            runnable_code_blocks.push(block.clone());
+        }
+    }
+}
+
+pub fn collect_code_blocks_from_tokens(
+    doc_tokens: &Option<Vec<DocumentationCommentToken>>,
+    full_path: &str,
+) -> Vec<CodeBlock> {
+    let Some(tokens) = doc_tokens else {
+        return Vec::new();
+    };
+
+    #[derive(Debug)]
+    struct CodeFence {
+        token_idx: usize,
+        char: u8,
+        len: usize,
+        info_string: String,
+    }
+
+    let mut code_blocks = Vec::new();
+    let mut current_fence: Option<CodeFence> = None;
+    let mut block_index: usize = 0;
+
+    for (idx, token) in tokens.iter().enumerate() {
+        let content = match token {
+            DocumentationCommentToken::Content(content) => content.trim(),
+            DocumentationCommentToken::Link(_) => continue,
+        };
+        if content.is_empty() {
+            continue;
+        }
+        match current_fence {
+            // Handle potential closing fence.
+            Some(ref opening) => {
+                if is_matching_closing_fence(content, opening.char, opening.len) {
+                    let end_idx = idx;
+                    let body = get_block_body(tokens, opening.token_idx + 1, end_idx);
+
+                    // Skip empty code blocks.
+                    if !body.is_empty() {
+                        let id = CodeBlockId::new(full_path.to_string(), block_index, end_idx);
+                        code_blocks.push(CodeBlock::new(
+                            id,
+                            body.to_string(),
+                            &opening.info_string,
+                        ));
+                        block_index += 1;
+                    }
+                    current_fence = None;
+                }
+            }
+            // Handle potential opening fence.
+            None => {
+                if let Some((len, char)) = scan_code_fence(content.as_bytes()) {
+                    let bytes = content.as_bytes();
+                    let after = &bytes[len..];
+                    let info_string = from_utf8(after).unwrap_or("").trim().to_string();
+
+                    current_fence = Some(CodeFence {
+                        token_idx: idx,
+                        char,
+                        len,
+                        info_string,
+                    });
+                }
+            }
+        }
+    }
+    // There may be an unterminated fence at this point, but this is allowed from the spec perspective, so we ignore it.
+    code_blocks
+}
+
+fn get_block_body(
+    tokens: &[DocumentationCommentToken],
+    start_idx: usize,
+    end_idx: usize,
+) -> String {
+    tokens[start_idx..end_idx]
+        .iter()
+        .filter_map(|token| match token {
+            DocumentationCommentToken::Content(content) => Some(content.as_str()),
+            DocumentationCommentToken::Link(_) => None,
+        })
+        .collect::<Vec<_>>()
+        .join("")
+        .trim()
+        .to_string()
+}
+
+/// Checks if the given `content` is a closing fence matching the given opening fence.
+fn is_matching_closing_fence(content: &str, opening_char: u8, opening_len: usize) -> bool {
+    let bytes = content.as_bytes();
+    let Some((len, ch)) = scan_code_fence(bytes) else {
+        return false;
+    };
+    ch == opening_char
+        && len >= opening_len
+        && bytes[len..]
+            .iter()
+            .all(|&b| matches!(b, b' ' | b'\t' | b'\r' | b'\n'))
+}
+
+/// Copied from `pulldown-cmark`:
+/// https://github.com/pulldown-cmark/pulldown-cmark/blob/a574ea8a5e6fda7bc26542a612130a2b458a68a7/pulldown-cmark/src/scanners.rs#L744
+fn scan_code_fence(data: &[u8]) -> Option<(usize, u8)> {
+    let c = *data.first()?;
+    if !(c == b'`' || c == b'~') {
+        return None;
+    }
+    let i = 1 + scan_ch_repeat(&data[1..], c);
+    if i >= 3 {
+        if c == b'`' {
+            let suffix = &data[i..];
+            let next_line = i + scan_nextline(suffix);
+            // FIXME: make sure this is correct
+            if suffix[..(next_line - i)].contains(&b'`') {
+                return None;
+            }
+        }
+        Some((i, c))
+    } else {
+        None
+    }
+}
+
+fn scan_ch_repeat(data: &[u8], c: u8) -> usize {
+    data.iter().take_while(|&&b| b == c).count()
+}
+
+fn scan_nextline(bytes: &[u8]) -> usize {
+    bytes
+        .iter()
+        .position(|&b| b == b'\n')
+        .map_or(bytes.len(), |x| x + 1)
+}
--- a/extensions/scarb-doc/src/doc_test/runner.rs
+++ b/extensions/scarb-doc/src/doc_test/runner.rs
@@ -1,0 +1,380 @@
+use crate::AdditionalMetadata;
+use crate::doc_test::code_blocks::{
+    CodeBlock, CodeBlockAttribute, CodeBlockId, count_blocks_per_item,
+};
+use crate::doc_test::ui::TestResultStatus;
+use crate::doc_test::workspace::DocTestWorkspace;
+use anyhow::{Context, Result};
+use itertools::Itertools;
+use scarb_metadata::ScarbCommand;
+use scarb_ui::Ui;
+use scarb_ui::components::{NewLine, Status};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::process::{Command, Stdio};
+use tempfile::TempDir;
+use tempfile::tempdir;
+
+pub type ExecutionResults = HashMap<CodeBlockId, ExecutionResult>;
+
+#[derive(Debug, Clone)]
+pub struct ExecutionResult {
+    pub status: TestStatus,
+    pub print_output: String,
+    pub program_output: String,
+    pub outcome: ExecutionOutcome,
+}
+
+impl ExecutionResult {
+    pub fn as_markdown(&self) -> String {
+        let mut output = String::new();
+        if !self.print_output.is_empty() {
+            output.push_str("\nOutput:\n```\n");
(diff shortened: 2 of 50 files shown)


maciektr:

- #2983 ◀
    - #3107
        - #3031
            - #3108
                - #3110

extensions/scarb-doc/src/docs_generation/markdown/traits.rs
```diff
@@ -789,7 +810,7 @@ fn generate_markdown_for_subitems<T: MarkdownDocItem + SubPathDocItem>(
             writeln!(
                 &mut markdown,
                 "{}",
-                item.generate_markdown(context, header_level + 2, postfix, summary_index_map)?
+                item.generate_markdown(context, header_level + 2, postfix, summary_index_map,)?
```
wawel37: ?

extensions/scarb-doc/tests/runnable_examples.rs
```diff
@@ -0,0 +1,388 @@
...
+            [..] Running 1 doc examples for `hello_world`
+            [..] Compiling hello_world_example_1 v0.1.0 ([..])
+            [..]  Finished `dev` profile target(s) in [..]
+            test hello_world::foo ... FAILED
+
+            failures:
+                hello_world::foo
+
+            test result: FAILED. 0 passed; 1 failed; 0 ignored
+            error: doc tests failed
```
wawel37, line 235: I'm thinking if we could display diagnostics/logs from failed tests here, so the users have a clue what's wrong with the code. Wdyt, could we do that?

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -0,0 +1,278 @@
...
+                        Err(e) => {
+                            summary.failed += 1;
+                            self.ui.print(TestResult::failed(&display_name));
+                            failed_names.push(display_name);
+                            self.ui.error(format!("Error running example: {:#}", e));
+                        }
+                    }
+                }
+            }
+        }
```
wawel37, line 169: I don't know about the performance of this. Could this be parallelized? I mean, we don't need to do that right now, but i would test running it for some bigger repository and see if the performance stays right

extensions/scarb-doc/src/doc_test/workspace.rs
```diff
@@ -0,0 +1,135 @@
...
+            edition = "{edition}"
+
+            [dependencies]
+            {dep} = {{ path = "{dep_path}" }}
+            cairo_execute = "{CAIRO_VERSION}"
+
+            [cairo]
+            enable-gas = false
+
+            [executable]
```
wawel37, line 87: If i get it correctly, this will be added even if the strategy is `no_run`. Will this be a problem?

extensions/scarb-doc/src/doc_test/code_blocks.rs
```diff
@@ -0,0 +1,303 @@
...
+            close_token_idx,
+        }
+    }
+
+    // TODO: (#2888): Display exact code block location when running doc-tests
+    pub fn display_name(&self, total_blocks_in_item: usize) -> String {
+        if total_blocks_in_item <= 1 {
+            self.item_full_path.clone()
+        } else {
+            format!("{} (example {})", self.item_full_path, self.block_index)
```
wawel37:
```suggestion
            format!("{} (example {})", self.item_full_path, self.block_index + 1)
```
So the users see indexing starting from 1? Wdyt?

extensions/scarb-doc/src/doc_test/workspace.rs
maciektr: Note, that this structure holds a temporary directory in which we write all the changes. This does not affect users code in any way. 
Also, with --no-run, we will never call this code (see  https://github.com/software-mansion/scarb/blob/b977424d107e419bcd241a167bef0181bc24271a/extensions/scarb-doc/src/main.rs#L133-L133 ; https://github.com/software-mansion/scarb/blob/b977424d107e419bcd241a167bef0181bc24271a/extensions/scarb-doc/src/doc_test/runner.rs#L172-L172)

maciektr: Renamed the structure to make it more clear.

extensions/scarb-doc/src/doc_test/code_blocks.rs
maciektr: Good idea!

extensions/scarb-doc/tests/runnable_examples.rs
maciektr: Good point, done!

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: Will be a follow up: https://github.com/software-mansion/scarb/pull/3031

wawel37 (approved):

extensions/scarb-doc/src/doc_test/code_blocks.rs
```diff
@@ -0,0 +1,304 @@
...
+            return true;
+        }
+        false
+    }
+
+    pub fn run_strategy(&self) -> RunStrategy {
+        if self.attributes.contains(&CodeBlockAttribute::Ignore) {
+            return RunStrategy::Ignore;
+        }
+        // TODO: drop the `runnable` attribute requirement; default to runnable for Cairo blocks
```
integraledelebesgue: TODO

extensions/scarb-doc/src/types/other_types.rs
```diff
@@ -89,6 +92,94 @@ impl<'db> FreeFunction<'db> {
...
+pub struct Struct<'db> {
+    #[serde(skip)]
+    pub id: StructId<'db>,
+    #[serde(skip)]
+    pub node: ast::ItemStructPtr<'db>,
+
+    pub members: Vec<Member<'db>>,
+
+    pub item_data: ItemData<'db>,
+}
```
wawel37: how does it differ from `struct_types::Struct`? And where is it used, because i don't see it being used anywhere? And if it's actually this, maybe we should have some more specific naming for this, because on the first glance, i can't think about a difference between `struct_types::Struct` and `other_types::Struct`. Same goes for `Member`

maciektr: Good find, removed dead code

extensions/scarb-doc/src/doc_test/code_blocks.rs
Arcticae, line 98: I think we should do this in the initial version

extensions/scarb-doc/src/doc_test/code_blocks.rs
```diff
@@ -0,0 +1,304 @@
...
+    pub fn new(id: CodeBlockId, content: String, info_string: &str) -> Self {
+        let attributes = Self::parse_attributes(info_string);
+        Self {
+            id,
+            content,
+            attributes,
+        }
+    }
+
+    // TODO: default to Cairo unless specified otherwise
```
Arcticae: Please make an issue if leaving a TODO

extensions/scarb-doc/src/doc_test/code_blocks.rs
```diff
@@ -0,0 +1,304 @@
...
+        if self.attributes.contains(&CodeBlockAttribute::ShouldPanic) {
+            return ExecutionOutcome::RuntimeError;
+        }
+        if self.attributes.contains(&CodeBlockAttribute::NoRun) {
+            return ExecutionOutcome::BuildSuccess;
+        }
+        ExecutionOutcome::RunSuccess
+    }
+
+    fn parse_attributes(info_string: &str) -> Vec<CodeBlockAttribute> {
```
Arcticae: [nit]: Please place higher up, closer to usage

extensions/scarb-doc/src/doc_test/code_blocks.rs
```diff
@@ -0,0 +1,304 @@
...
+        if !self.is_cairo() || !self.attributes.contains(&CodeBlockAttribute::Runnable) {
+            return RunStrategy::Ignore;
+        }
+        match self.expected_outcome() {
+            ExecutionOutcome::None => RunStrategy::Ignore,
+            ExecutionOutcome::BuildSuccess => RunStrategy::Build,
+            ExecutionOutcome::RunSuccess => RunStrategy::Execute,
+            ExecutionOutcome::CompileError => RunStrategy::Build,
+            ExecutionOutcome::RuntimeError => RunStrategy::Execute,
+        }
```
Arcticae: I don't get why we need `ExecutionOutcome` layer - the `RunStrategy` can easily be inferred directly from attribute (and we can probably add a `into` there), and we would need less conversions

extensions/scarb-doc/src/doc_test/code_blocks.rs
```diff
@@ -0,0 +1,304 @@
...
+    for module in &crate_.foreign_crates {
+        collect_from_module(module, &mut runnable_code_blocks);
+    }
+    runnable_code_blocks.sort_by_key(|block| block.id.clone());
+    runnable_code_blocks.dedup_by_key(|block| block.id.clone());
+    runnable_code_blocks
+}
+
+/// Counts the number of code blocks per documented item. Used to generate display names
+/// for code blocks, allowing to distinguish between multiple code blocks in the same item.
```
Arcticae, line 123: Shouldn't the `CodeBlockId`  be already sufficient for this?

extensions/scarb-doc/src/doc_test/code_blocks.rs
```diff
@@ -0,0 +1,304 @@
...
+fn collect_from_module(module: &Module<'_>, runnable_code_blocks: &mut Vec<CodeBlock>) {
+    for &item_data in module.get_all_item_ids().values() {
+        collect_from_item_data(item_data, runnable_code_blocks);
+    }
+    for &item_data in module.pub_uses.get_all_item_ids().values() {
+        collect_from_item_data(item_data, runnable_code_blocks);
+    }
+}
+
+fn collect_from_item_data(
```
Arcticae: [nit]: just inline it imo

extensions/scarb-doc/src/doc_test/workspace.rs
```diff
@@ -0,0 +1,135 @@
...
+use cairo_lang_parser::utils::SimpleParserDatabase;
+use camino::{Utf8Path, Utf8PathBuf};
+use indoc::formatdoc;
+use scarb_build_metadata::CAIRO_VERSION;
+use scarb_ui::Ui;
+use std::fmt::Write;
+use std::fs;
+use tempfile::{TempDir, tempdir};
+
+pub(crate) struct DocTestWorkspace {
```
Arcticae:
```suggestion
pub struct DocTestWorkspace {
```

extensions/scarb-doc/src/doc_test/workspace.rs
```diff
@@ -0,0 +1,135 @@
...
+    _temp_dir: TempDir,
```
Arcticae, line 15: I don't know if we're suggesting visibility or not - if yes we should use getter pattern instead and omit the one we don't want to access 

```suggestion
    temp_dir: TempDir,
```

extensions/scarb-doc/src/doc_test/workspace.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn write_manifest(&self, metadata: &AdditionalMetadata) -> Result<()> {
+        let package_dir = metadata
+            .manifest_path
+            .parent()
+            .context("package manifest path has no parent directory")?;
+
+        let dep = &metadata.name;
+        let dep_path = format!("{}", package_dir);
+        let name = &self.package_name;
+        let edition = edition_variant(Edition::latest());
```
Arcticae: What if the code is written on earlier edition? They are supposed to introduce breaking changes, no?

extensions/scarb-doc/src/doc_test/ui.rs
```diff
@@ -0,0 +1,94 @@
...
+            "test result: {}. {} passed; {} failed; {} ignored",
+            style.apply_to(result),
+            self.passed,
+            self.failed,
+            self.ignored
+        )
+    }
+
+    fn structured<S: Serializer>(self, ser: S) -> anyhow::Result<S::Ok, S::Error> {
+        self.serialize(ser)
```
Arcticae, line 24: 🧀

extensions/scarb-doc/src/doc_test/ui.rs
```diff
@@ -0,0 +1,94 @@
...
+use scarb_ui::Message;
+use serde::{Serialize, Serializer};
+
+impl Message for TestSummary {
+    fn text(self) -> String {
+        let (result, style) = if self.is_ok() {
+            ("ok", Style::new().green())
+        } else {
+            ("FAILED", Style::new().red())
+        };
```
Arcticae: isn't this duplicating logic from below?

extensions/scarb-doc/src/doc_test/ui.rs
```diff
@@ -0,0 +1,94 @@
...
+        Self::new(name, TestResultStatus::Ok)
+    }
+
+    pub fn failed(name: &'a str) -> Self {
+        Self::new(name, TestResultStatus::Failed)
+    }
+
+    pub fn ignored(name: &'a str) -> Self {
+        Self::new(name, TestResultStatus::Ignored)
+    }
```
Arcticae: why not just have the string inside the enum value?

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -0,0 +1,345 @@
...
+    Build,
+    Execute,
+}
+
+/// A runner for executing ([`CodeBlock`]) examples found in documentation.
+/// Uses the target package as a dependency and runs each code block in an isolated temporary workspace.
+/// Relies on `scarb build` and `scarb execute` commands to build and run the examples,
+/// based on the requested [`RunStrategy`] for the given code block.
+///
+/// Note: it is expected examples (`code_blocks`) that this runner executes only depend on the target package and standard libraries.
```
Arcticae, line 92: This isn't stated in the docs

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -0,0 +1,345 @@
...
+pub struct TestRunner<'a> {
+    /// Metadata of the target package whose documentation is being tested.
+    metadata: &'a AdditionalMetadata,
+    has_lib_target: bool,
+    ui: Ui,
+    /// Target directory shared between all doc test runs.
```
Arcticae:
```suggestion
    /// Target directory shared between all doc test runs within one package.
```

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -0,0 +1,345 @@
...
+    Execute,
+}
+
+/// A runner for executing ([`CodeBlock`]) examples found in documentation.
+/// Uses the target package as a dependency and runs each code block in an isolated temporary workspace.
+/// Relies on `scarb build` and `scarb execute` commands to build and run the examples,
+/// based on the requested [`RunStrategy`] for the given code block.
+///
+/// Note: it is expected examples (`code_blocks`) that this runner executes only depend on the target package and standard libraries.
+pub struct TestRunner<'a> {
```
Arcticae, line 93: We should probably state if it's meant to be used for a single package/workspace

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -0,0 +1,345 @@
...
+                    summary.ignored += 1;
+                    self.ui.print(TestResult::ignored(&display_name));
+                }
+                _ => {
+                    idx += 1;
+                    match self.run_single(block, strategy, idx) {
+                        Ok(res) => match res.status {
+                            TestStatus::Passed => {
+                                summary.passed += 1;
+                                self.ui.print(TestResult::ok(&display_name));
```
Arcticae: This doesn't print program output or stdout?

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -0,0 +1,345 @@
...
+                            summary.failed += 1;
+                            self.ui.print(TestResult::failed(&display_name));
+                            failed_names.push(display_name);
+                            self.ui.error(format!("Error running example: {:#}", e));
+                        }
+                    }
+                }
+            }
+        }
+        // TODO: add struct with `impl Message` to display this
```
Arcticae: Please add an issue #

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -0,0 +1,345 @@
...
+        if !failed_names.is_empty() {
+            self.ui.print("\nfailures:");
+            for display_name in &failed_names {
+                self.ui.print(format!("    {}", display_name));
+            }
+        }
+        self.ui.print(NewLine::new());
```
Arcticae, line 178: why is it printed here like this and `\n <text>` above?

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    fn run_single_inner(
+        &self,
+        ws: &DocTestWorkspace,
+        strategy: RunStrategy,
+    ) -> Result<(ExecutionOutcome, String, String)> {
+        if strategy == RunStrategy::Ignore {
+            unreachable!("the code block should be filtered out before reaching here");
+        }
```
Arcticae, line 262: shouldn't this be `debug_assert!()` ?

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -0,0 +1,345 @@
...
+            .env("SCARB_TARGET_DIR", self.target_dir.path())
+            .env("SCARB_UI_VERBOSITY", self.ui.verbosity().to_string())
+            .env("SCARB_MANIFEST_PATH", ws.manifest_path().as_str())
+            .env("SCARB_ALL_FEATURES", "true")
+            .run();
+
+        if build_result.is_err() {
+            return Ok((ExecutionOutcome::CompileError, String::new(), String::new()));
+        } else if strategy == RunStrategy::Build {
+            return Ok((ExecutionOutcome::BuildSuccess, String::new(), String::new()));
```
Arcticae, line 275: Don't we have outputs to pass here from the build?

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -0,0 +1,345 @@
...
+
+#[derive(Deserialize)]
+struct ExecutionErrorLine {
+    #[serde(rename = "type")]
+    message_type: String,
+    message: String,
+}
+
+#[derive(Deserialize)]
+#[allow(dead_code)]
```
Arcticae: ?

extensions/scarb-doc/tests/runnable_examples.rs
Arcticae, line 1: We could use an example with a workspace and multiple packages inside, each with it's own doc tests

extensions/scarb-doc/tests/runnable_examples.rs
```diff
@@ -0,0 +1,422 @@
+use assert_fs::TempDir;
+use indoc::{formatdoc, indoc};
+use scarb_test_support::command::Scarb;
+// use scarb_test_support::filesystem::dump_temp;
```
Arcticae: ?

Arcticae (changes requested):

extensions/scarb-doc/src/doc_test/workspace.rs
maciektr: We hold the temp_dir here, because TempDir has some action associated with it's drop (removal of the underlying directory from the fs), but the temp_dir var itself is never accessed. So, it's marked with the `_` prefix as unused value. Otherwise, we would need `#[allow(dead_code)]` to get rid of clippy warning.

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: Wouldn't this be to noisy? 🤔

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: If this is truly unreachable, then I think that panicking is the right behaviour - no reason for this path to be hit at runtime 🤔

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: Same as above - wouldn't this be too noisy?

extensions/scarb-doc/src/doc_test/ui.rs
maciektr: 🧀

extensions/scarb-doc/src/doc_test/code_blocks.rs
maciektr: I mean, you may have multiple code block per item (thus multiple CodeBlockIds), no? :thinking:

extensions/scarb-doc/src/doc_test/runner.rs
Arcticae: I don't know, but at least some option to enable this would be nice, in order to debug?

extensions/scarb-doc/src/doc_test/runner.rs
Arcticae: this is checking if a developer hasn't made an error composing the functions, which shouldn't be done really (it's always true in current code setup). now it's checking this if every time. although this is not really a location in which the performance is crucial, so i'll let you choose which you prefer

extensions/scarb-doc/src/doc_test/runner.rs
Arcticae: vide supra

extensions/scarb-doc/src/doc_test/workspace.rs
Arcticae: I think this is worth mentioning in the comment

extensions/scarb-doc/src/doc_test/code_blocks.rs
Arcticae: I mean CodeBlockId should already be a sufficient way to identify a code block, even within the same item. It does contain `end_index` which should be unique enough to establish identity right?

maciektr: But we want to collect all code blocks for specific item, so Id is too precise for us

Arcticae: Yeah but we can already do that, since we have the item name in the ID (that's what we're using below) 

And creating a display name could be done, appending the `end_index` 

Thus, i don't get why this function even exists if all we want can already be done with the ID

maciektr: The function does use the id though? 🤔 
```
        *counts.entry(block.id.item_full_path.clone()).or_insert(0) += 1;
```

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: Ahh, I get it now, you want to get rid of the branching, 👍

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: Done in https://github.com/software-mansion/scarb/pull/3107

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: Done in https://github.com/software-mansion/scarb/pull/3107

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: Does it really matter? :thinking:

Arcticae: for cleanliness, yes?

Arcticae (approved):

PR #2986, opened 2026-02-09
Add informative error message on missing target dir env var in extensions

maciektr:


extensions/scarb-prove/src/main.rs
```diff
@@ -62,10 +61,7 @@ fn main_inner(args: Args, ui: Ui) -> Result<()> {
         Some(id) => id,
         None => {
             assert!(args.execute);
-            let scarb_target_dir = Utf8PathBuf::from(
-                env::var("SCARB_TARGET_DIR")
-                    .context("`SCARB_TARGET_DIR` env var must be defined")?,
-            );
+            let scarb_target_dir = scarb_target_dir_from_env()?;
```
wawel37: do we need it here too? isnt it better to clone the previously defined one?

extensions/scarb-execute/src/lib.rs
```diff
@@ -430,6 +424,22 @@ pub fn execute(
...
+            if manifest_path.exists() {
+                bail!("`SCARB_TARGET_DIR` env var must be defined")
+            } else {
+                bail!(
+                    "no {MANIFEST_FILE_NAME} found, this command must be run inside a Scarb project"
+                )
+            }
+        }
+    }
+}
```
wawel37, line 446: This repeats throughout the extensions. I know it's a little piece of code, but is there any more pieces of code that repeats between extensions? Just asking, cuz if there's a lot of it, maybe we should think about moving all of this code to shared packaged, like extension utils? wdyt?

maciektr: Yep, I see what you mean, though I wouldn't say it's a lot - mostly in main.rs files and they are short too. Maybe later on if it grows too much.

extensions/scarb-prove/src/main.rs
maciektr: Fixed!

wawel37 (approved):

PR #2995, opened 2026-02-12
bump stwo and cairo

Arcticae:


Cargo.toml
```diff
@@ -167,7 +167,7 @@ which = "8"
 windows-sys = { version = "0.61", features = ["Win32_Foundation", "Win32_System", "Win32_System_Console"] }
 xshell = "0.2"
 xxhash-rust = { version = "0.8", features = ["xxh3"] }
-zip = { version = "7.4", default-features = false, features = ["deflate"] }
+zip = { version = "7.2", default-features = false, features = ["deflate"] }
```
maciektr: nit
```suggestion
zip = { version = "7", default-features = false, features = ["deflate"] }
```

Cargo.toml
```diff
@@ -91,7 +91,7 @@ dialoguer = "0.12.0"
 directories = "6"
 dunce = "1"
 expect-test = "1.5"
-flate2 = { version = "1.1.9", default-features = false, features = ["zlib"] }
+flate2 = { version = "1.1.8", default-features = false, features = ["zlib"] }
```
maciektr: nit 
```suggestion
flate2 = { version = "1", default-features = false, features = ["zlib"] }
```

extensions/scarb-verify/src/main.rs
maciektr, line 1: Broken merge - please keep target dir handling from main.

.github/workflows/_check-release.yml
maciektr, line 1: Why?

Arcticae: `--all-features` will enable `prover` in `stwo-cairo` which will pull `stwo`

maciektr: Doesn't it only enable features in workspace packages, not in dependency crates?

Arcticae: Adapter has to be compiled with enabled feature "prover" in order to work for `scarb-prove`
It has to be compiled without it for `scarb-execute`
So i have to pull it with features and not enable them if i want to have 1 dependency source only (workspace).

maciektr (approved):

Issue #2992, opened 2026-02-12
Separate crate to serve Manifest fields documentation


FroyaTheHen:
- Separate logic that will extract and expose the TomlMetadata fields docstrings. 
- Can use [schemars](https://crates.io/crates/schemars). 
- The purpose is to use it in LS, so can be done similarly to [scarb-metadata](https://github.com/software-mansion/cairols/blob/e7c2f3a42e246d295c744f3b0f87787f25f6947f/Cargo.toml#L72)
- Explain the above to the future devs in README or similiar

PR #2996, opened 2026-02-13
scarb-manifest-schema

FroyaTheHen:
Note:  tests under `utils/scarb-manifest-schema/tests/traverse.rs` make more sense on the next in stack [PR](https://github.com/software-mansion/scarb/pull/2997) where the documentation update actually influences the results. 

closes https://github.com/software-mansion/scarb/issues/2992

### DIFF
--- a/utils/scarb-manifest-schema/Cargo.toml
+++ b/utils/scarb-manifest-schema/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "scarb-manifest-schema"
+version = "1.0.0"
+description = "Allows access to the Scarb manifest fields documentation"
+readme = "README.md"
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[dependencies]
+anyhow.workspace = true
+scarb = { path = "../../scarb" }
+schemars.workspace = true
+serde_json.workspace = true
--- a/utils/scarb-manifest-schema/src/lib.rs
+++ b/utils/scarb-manifest-schema/src/lib.rs
@@ -1,0 +1,109 @@
+//! Dynamic JSON Schema for Scarb manifest.
+//!
+//! Provides a way to generate and traverse the Scarb manifest JSON Schema and retrieve the
+//! definition for a specific TOML path, such as `package.dependencies`.
+
+use anyhow::{Result, anyhow};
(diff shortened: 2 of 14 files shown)


utils/scarb-manifest-schema/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+readme = "README.md"
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[dependencies]
+anyhow.workspace = true
+scarb = { path = "../../scarb" }
```
maciektr, line 14: FYI you will not be able to upload it to crates.io with this dep. 
You can either define the toml manifest here and import in Scarb, create another crate for it, or play around with `#[path = ""] mod` tricks (somewhat like this https://github.com/software-mansion/scarb/blob/2a5820adf636a09827a4b2eaa1189e27382b0b17/utils/scarb-test-support/src/project_builder.rs#L18-L18)

FroyaTheHen: Ok, appreciate the heads up

FroyaTheHen: resolved in https://github.com/software-mansion/scarb/pull/3007

utils/scarb-manifest-schema/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+//! Provides a way to generate and traverse the Scarb manifest JSON Schema and retrieve the
+//! definition for a specific TOML path, such as `package.dependencies`.
+
+use anyhow::{Result, anyhow};
+use scarb::core::TomlManifest;
+use schemars::schema_for;
+use serde_json::Value;
+use std::sync::OnceLock;
+
+static GLOBAL_TRAVERSER: OnceLock<SchemaTraverser> = OnceLock::new();
```
Arcticae, line 12: Why is it made global?

FroyaTheHen: because it is a read only value that we only want to get once, have live during entire program duration, and reuse. There is a doc comment right below for `pub fn get_shared_traverser(` and I did place the comment there because it is a part of public api whereas the `GLOBAL_TRAVERSER` isn't.

Arcticae (approved):

PR #3003, opened 2026-02-18
scarb-manifest-schema: changelog and readme

FroyaTheHen:
**Stack**:
- #2997
- #2996

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #2997, opened 2026-02-13
update docstrings

FroyaTheHen:
- I will update `utils/scarb-manifest-schema/CHANGELOG.md` and `utils/scarb-manifest-schema/README.md` in following PR, just this is shippable pice so I wanted to publish it already as I assume (and hope) that the docstrings review will cause a lot of noise.
- The docstring were partially AI generated, I did read and edit all of them but still consider it  might be relevant to let you know 🫡 
- Also, originally we discussed having the dosctrings one-to-one copy of what's on the [offcicial doc web](https://docs.swmansion.com/scarb/docs/reference/manifest.html) but I did shift from that idea as they are tailored for the web - let me know if you do not agree and would rather see them as a copy of what's on the web 🙏 

closes https://github.com/software-mansion/scarb/issues/2991

PR #3007, opened 2026-02-20
scarb-manifest-schema-gen

FroyaTheHen:
**Stack**:
- #3003
- #2997
- #2996

### DIFF
--- a/utils/scarb-manifest-schema-gen/README.md
+++ b/utils/scarb-manifest-schema-gen/README.md
@@ -1,0 +1,30 @@
+# scarb-manifest-schema-gen
+
+An internal workspace utility to synchronise the scarb-manifest-schema JSON snapshot with the source of truth in the scarb core crate.
+
+## Purpose
+The `scarb-manifest-schema` crate provides a JSON Schema for Scarb's `Scarb.toml` manifest. 
+To avoid making the core scarb crate a mandatory runtime dependency for the `scarb-manifest-schema`, we use a snapshotting strategy:
+
+- Source: The `TomlManifest` struct lives in the scarb `scarb::core::manifest` crate.
+
+- This Tool: Generates a `schema.json` file from the Rust structs.
+
+- Library: The `scarb-manifest-schema` exposes the schema to use outside the Scarb workspace.
+
+## Usage
+You should run this tool whenever you modify the `TomlManifest` struct in `/scarb/src/core/manifest/toml_manifest.rs`
+
+To update run: 
+```bash
+cargo run -p scarb-manifest-schema-gen
+```
+
+## Consistency Check
+The `test_schema_is_up_to_date` test in `scarb-manifest-schema` ensures that the snapshot is up to date.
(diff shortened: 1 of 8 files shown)


utils/scarb-manifest-schema-gen/README.md
```diff
@@ -0,0 +1,30 @@
...
+To avoid making the core scarb crate a mandatory runtime dependency for the `scarb-manifest-schema`, we use a snapshotting strategy:
+
+- Source: The `TomlManifest` struct lives in the scarb `scarb::core::manifest` crate.
+
+- This Tool: Generates a `schema.json` file from the Rust structs.
+
+- Library: The `scarb-manifest-schema` exposes the schema to use outside the Scarb workspace.
+
+## Usage
+You should run this tool whenever you modify the `TomlManifest` struct in `/scarb/src/core/manifest/toml_manifest.rs`
```
maciektr, line 16: Maybe we can go a step further and just make scarb-manfiest-schema/build.rs generate the file each time the package is built? 🤔

FroyaTheHen: That would be perfect, however I'm not sure how to implement that. Modifying `scarb-manifest-schema/build.rs` to use `scarb-manifest-schema-gen` deadlocks the build. 
I only see that possible moving the publishing to an orchestrating event outside of the cargo scope such as github action for example or a publishing script where the tests and generation can be run one after another. Any thoughts?

utils/scarb-manifest-schema-gen/src/main.rs
```diff
@@ -0,0 +1,55 @@
+use scarb::core::TomlManifest;
+use std::fs;
+use std::path::PathBuf;
+
+fn main() {
+    let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
```
Arcticae, line 6: Why not CWD?

utils/scarb-manifest-schema-gen/README.md
maciektr: Ah, yeah, I am hallucinating, sorry 😓 We would need to distribute scarb-manifest-schema-gen along the scarb-manifest-schema, which we cannot do 😅.

maciektr: Can we have a CI step to check if it's up to date then?

FroyaTheHen: Sure! This is happening, already. There is a test that will fail if schema is outdated. 
[see example run](https://github.com/software-mansion/scarb/actions/runs/22291334681/job/64480282213)

FroyaTheHen: Do you think it's better to implement it as a separate step?

utils/scarb-manifest-schema-gen/src/main.rs
FroyaTheHen: the env always sets the output dir to the expected one, while with cwd the output dir would depend on the directory where the command happened to be run from

utils/scarb-manifest-schema-gen/README.md
maciektr: Great, thanks!

maciektr (approved):

Issue #2977, opened 2026-02-04
Improve macro API docs


wawel37:
Especially "span matching". Check this code for a reference https://github.com/OpenZeppelin/cairo-contracts/blob/924e9d019fa943f53a6a314445d777809f621948/packages/macros/src/attribute/with_components/definition.rs#L12.

PR #3014, opened 2026-02-24
Refactor and extend proc macro docs

wawel37:
resolves https://github.com/software-mansion/scarb/issues/2977
The proc macro docs where split into appropriate sections. 
Please note that the Example number 4 is completely new, and refer the problem of people from OZ.

### DIFF
--- a/scarb/tests/examples.rs
+++ b/scarb/tests/examples.rs
@@ -5,6 +5,7 @@ use assert_fs::prelude::*;
 use snapbox::cmd::Command;
 
 use scarb_test_support::cargo::cargo_bin;
+use scarb_test_support::command::Scarb;
 use test_for_each_example::test_for_each_example;
 
 #[test_for_each_example(ignore = "dependencies,procedural_macros")]
@@ -52,13 +53,7 @@ fn build_procedural_macro_examples() {
 
             let cairo_repo = entry.path().join("cairo-repository");
             if cairo_repo.is_dir() {
-                Command::new(cargo_bin("scarb"))
-                    .arg("clean")
-                    .current_dir(&cairo_repo)
-                    .assert()
-                    .success();
-
-                Command::new(cargo_bin("scarb"))
+                Scarb::quick_command()
                     .arg("build")
                     .current_dir(&cairo_repo)
                     .assert()
--- a/examples/procedural_macros/example2/macro/Cargo.lock
+++ b/examples/procedural_macros/example2/macro/Cargo.lock
@@ -68,6 +68,7 @@ name = "example2_macro"
 version = "0.1.0"
 dependencies = [
  "cairo-lang-macro",
+ "cairo-lang-primitive-token",
 ]
 
 [[package]]


website/docs/procedural-macros/index.md
```diff
@@ -0,0 +1,36 @@
...
+
+> [!WARNING]
+> Procedural macros, by design, introduce a lot of overhead during the compilation.
+> They may also be harder to maintain.
+> Prefer the declarative inline macros written directly in Cairo, unless you have a specific reason to use procedural macros.
+> Please see the [declarative macros chapter in Cairo Book](https://www.starknet.io/cairo-book/ch12-05-macros.html#declarative-inline-macros-for-general-metaprogramming) for more information.
+
+> [!INFO]
+> To use procedural macros, you need to have Rust toolchain (Cargo) installed on your machine.
+> Please see [Rust installation guide](https://www.rust-lang.org/tools/install) for more information.
```
maciektr, line 18: Maybe it would make sense to keep those in all sub pages as well? 🤔 Wdyt?

website/docs/procedural-macros/examples.md
```diff
@@ -0,0 +1,468 @@
+<script setup>
+import { data as rel } from "../../github.data";
+</script>
+
+# Examples
+
+## Example 1: returning a value
```
maciektr, line 26: I am not sure if we should move it here. The parsing function is very primitive, it's panicking, the errors are not handled at all, etc.  It served as a way to demonstrate what you can do with very limited knowledge you have after reading "writing proc macros" seciton, not as an example of how good macro looks. 

It seems sensible to me, that we should keep such demonstrations as a comment to guides we write in subpages, and only promote good macros in the example section - maybe even community written ones?

website/docs/procedural-macros/examples.md
```diff
@@ -0,0 +1,468 @@
...
+
+fn parse_function_name<'db>(db: &'db SimpleParserDatabase, node: SyntaxNode<'db>) -> String {
+    assert_eq!(node.kind(db), SyntaxKind::SyntaxFile);
+    let file = ast::SyntaxFile::from_syntax_node(db, node);
+    let items = file.items(db).elements_vec(db);
+    assert_eq!(items.len(), 1);
+    let func = items.into_iter().next().unwrap();
+    assert!(matches!(func, ModuleItem::FreeFunction(_)));
+    let ModuleItem::FreeFunction(func) = func else {
+        panic!("not a function");
```
maciektr: vide supra

website/docs/procedural-macros/examples.md
```diff
@@ -0,0 +1,468 @@
...
+Collected 1 test(s) from hello_world package
+Running 1 test(s) from src/
+[PASS] hello_world::tests::it_works (l1_gas: ~0, l1_data_gas: ~0, l2_gas: ~80000)
+Tests: 1 passed, 0 failed, 0 ignored, 0 filtered out
+```
+
+## Example 4: modifying an already existing function
+
+Using quote macro we can as well modify the body of the function that the attribute macro is being applied to.
+In this example, with the use of an attribute macro, we will define completely new variable inside the function, which later will be used in a user code. We also make all the diagnostics from user code correctly mapped to the origin code.
```
maciektr:
```suggestion
Using the quote macro, we can also modify the body of the function that the attribute macro is being applied to.
In this example, with the use of an attribute macro, we will define a completely new variable inside the function, which later will be used in a user code. We also make all the diagnostics from the user code correctly mapped to the original code.
```

website/docs/procedural-macros/examples.md
```diff
@@ -0,0 +1,468 @@
...
+
+```rust
+// src/lib.rs
+use cairo_lang_macro::{attribute_macro, quote, ProcMacroResult, TokenStream};
+use cairo_lang_parser::{printer::print_tree, utils::SimpleParserDatabase};
+use cairo_lang_syntax::node::{ast, with_db::SyntaxNodeWithDb, TypedSyntaxNode};
+
+#[attribute_macro]
+fn my_macro(_attr: TokenStream, code: TokenStream) -> ProcMacroResult {
```
maciektr: NIT: just to keep consistent with other examples here 
```suggestion
fn my_macro(_args: TokenStream, body: TokenStream) -> ProcMacroResult {
```

website/docs/procedural-macros/examples.md
```diff
@@ -0,0 +1,468 @@
...
+    // Initialize parser and parse the incoming token stream.
+    let db = SimpleParserDatabase::default();
+    // Parse incoming token stream.
+    let (node, _diagnostics) = db.parse_token_stream(&code);
+
+    // This section is used only for macro debugging purposes.
+    // This way, we can see the exact syntax structure of the item we want to modify.
+    let node_tree = print_tree(&db, &node, false, false);
+    println!("node tree: \n{}", node_tree);
```
maciektr: Maybe we should add a "debugging macros" section or smth, with tips like that?

website/docs/procedural-macros/examples.md
```diff
@@ -0,0 +1,468 @@
...
+
+    // Locate the function item this attribute macro is applied to.
+    let module_item_list = node
+        .get_children(&db)
+        .get(0)
+        .expect("This attribute macro should be only used for a function");
```
maciektr: Let's return proper errors like we would do in a real macro (see https://github.com/maciektr/starknet-foundry/blob/e0b114347aa1d4cce298e97297b655084c402f4f/crates/snforge-scarb-plugin/src/common.rs#L30)

website/docs/procedural-macros/examples.md
wawel37: Tbh i would leave them here. I've changed them so they have proper handling of errors. They might not be some complicated ones, but such macros are in the examples below. In my opinion it's also good to keep really simple examples present here, so the users can start with something rather simple

website/docs/procedural-macros/examples.md
wawel37: https://github.com/software-mansion/scarb/pull/3014#discussion_r2858900796

website/docs/procedural-macros/using.md
```diff
@@ -0,0 +1,83 @@
...
+> Procedural macros, by design, introduce a lot of overhead during the compilation.
+> They may also be harder to maintain.
+> Prefer the declarative inline macros written directly in Cairo, unless you have a specific reason to use procedural macros.
+> Please see the [declarative macros chapter in Cairo Book](https://www.starknet.io/cairo-book/ch12-05-macros.html#declarative-inline-macros-for-general-metaprogramming) for more information.
+
+> [!INFO]
+> To use procedural macros, you need to have Rust toolchain (Cargo) installed on your machine.
+> Please see [Rust installation guide](https://www.rust-lang.org/tools/install) for more information.
+
+## Procedural macro user perspective
```
Arcticae, line 20: I would provide a diagram of how actually the architecture works (i.e there's scarb who loads a dylib file, the dylib is self-contained, receives outputs and inputs from scarb, etc.)

Visual aids are essential in more complicated topics and we underuse them

(This should be a separate task, its' not a concern to be resolved here but when we're at it let's make as best effort as we can)

website/docs/procedural-macros/using.md
```diff
@@ -0,0 +1,83 @@
...
+To use a procedural macro, a Cairo programmer needs to:
+
+- Declare a dependency on a package, that implements the procedural macro, by adding it to the `dependencies` section in
+  the Scarb manifest file.
+- Use the procedural macro in Cairo code, by calling it, or adding an attribute or derive macro to a Cairo item.
+
+Since Scarb procedural macros are, in fact, Rust functions that are usually distributed as source code and compiled into
+shared libraries (see [writing a procedural macro](./writing) for more details) on the user side,
+users are **required to have Rust toolchain installed** on their machine.
+This limitation can be omitted by distributing procedural macros as precompiled shared libraries, see
```
Arcticae, line 31: This is underdocumented - it should be more recommended to do this, since it makes the user compilation not use rust dep at all

Arcticae: Also can be a topic for separate issue btw

website/docs/procedural-macros/using.md
```diff
@@ -0,0 +1,83 @@
...
+- Scarb packages can simply declare dependency relationships on other packages that implement Cairo procedural macros.
+- The semantics of Scarb package resolution guarantee that only one instance of a given
+  procedural macro package exists in the resolved package set.
+  - In other words, Scarb will out of the box verify that there is no simultaneous dependency on `proc-macro 1.0.0`
+    and `proc-macro 2.0.0` or `proc-macro 1.0.1`.
+- Procedural macros will end up being actual Scarb compilation unit components, though, because they will have to be
+  treated differently from regular components, they will not be listed under `components` fields, but rather in a new
+  one: `plugins`.
+
+### Procedural macro must be called from Cairo code
```
Arcticae, line 38: I would place it first on the page tbh, and then explain the dependency concept

website/docs/procedural-macros/using.md
```diff
@@ -0,0 +1,83 @@
...
+#[derive(ToValue)]
+struct Input {
+    value: felt252,
+}
+
+#[instrument]
+fn main() -> felt252 {
+    let a = Input { value: 1 };
+    let b = Input { value: 2 };
+    add!(a.to_value(), b.to_value());
```
Arcticae, line 78: We should annotate this with comments (where is the macro?)

website/docs/procedural-macros/writing.md
```diff
@@ -0,0 +1,151 @@
+<script setup>
+import { data as rel } from "../../github.data";
+</script>
+
+# Writing Procedural Macros
+
+> [!WARNING]
+> Procedural macros, by design, introduce a lot of overhead during the compilation.
+> They may also be harder to maintain.
+> Prefer the declarative inline macros written directly in Cairo, unless you have a specific reason to use procedural macros.
```
Arcticae, line 10: We don't say what the reason is

website/docs/procedural-macros/writing.md
```diff
@@ -0,0 +1,151 @@
...
+- This function accepts a `TokenStream` as an input and returns a `ProcMacroResult` as an output, both defined in the
+  helper library.
+- The result struct contains the transformed `TokenStream`. Three kinds of results are possible:
+  - If the `TokenStream` is the same as the input, the AST is not modified.
+  - If the `TokenStream` is different from the input, the input is replaced with the generated code.
+  - If the `TokenStream` is empty, the input is removed.
+- Alongside the new TokenStream, a procedural macro can emit compiler diagnostics, auxiliary data and full path
+  identifiers, described in detail in advanced macro usage section.
+
+We define `TokenStream` as some encapsulation of code represented in plain Cairo.
```
Arcticae: This is not enough of an explanation of what it is

website/docs/procedural-macros/writing.md
```diff
@@ -0,0 +1,151 @@
...
+A procedural macro is implemented as a Rust library which defines functions that implement these transformations
+(later called macro _expansions_).
+This Rust code is then compiled into a shared library (shared object) and loaded into Scarb process memory during the
+Cairo project compilation.
+Scarb will call expansions from the loaded shared library, thus allowing you to inject custom logic to the Cairo
+compilation process.
+
+## Procedural macro author perspective
+
+To implement a procedural macro, a programmer has to:
```
Arcticae, line 36: Flowchart would fit really nice into here

website/docs/procedural-macros/debugging.md
```diff
@@ -0,0 +1,20 @@
+# Debugging
+
+## Inspecting the Syntax Tree
+
+Debugging procedural macros can be challenging. A practical approach is to inspect the syntax tree of the input using `cairo_lang_parser::printer::print_tree`. This function generates a readable representation of the parsed syntax structure, which you can print and examine to understand the token stream you're working with.
```
Arcticae, line 22: Also LS has a feature that allows you to view it. We should mention it here.

website/docs/procedural-macros/examples.md
```diff
@@ -0,0 +1,524 @@
...
+```
+[ap + 0] = 987, ap++;
+ret;
+```
+
+> [!INFO]
+> To see a real life example of a procedural macro that offloads some work into compile time,
+> you can take a look at the [`alexandria` project](https://github.com/keep-starknet-strange/alexandria/tree/6b98da52c819aeb86697b787b4bcf4abe94bc788/packages/macros).
+
+## Example 2: building token stream with `quote!` macro
```
Arcticae, line 157: Correct me if i'm wrong but we don't introduce how to use ths earlier (it's hard to get to this if you don't read all of it). It's the recommended way to create code with macros because of mappings and we should recommend it more (people don't use it because they don't know about it - we know it's a fact)

Arcticae (commented):
There are 2 concepts that are missing explanation from the docs 

1. Mappings - what is code mapping, how to properly map code back, some visualisation (really helpful when it comes to mappings)
1. Using cairo parser. I know it's not our responsibility to document it, but we should at least recommend how to explore the APIs, most important concepts which don't change over time, and how it fits in usual macro use case.

That said this PR is a really good improvement for clarity in the docs. They are vast and exhaustive in the topics they cover, but need to provide some visual aids IMO for ease of understanding of this complex topic 

I would merge this and add a milestone with some of my suggestions

website/docs/procedural-macros/examples.md
wawel37: Added additional section for `quote!` macro in `writing.md`

website/docs/procedural-macros/using.md
wawel37: There is a link to section in `advanced` below, and there's special section for explaining it how the precompiled macros can be shared to scarbs.xyz. I wouldn't say that this is underdocumented. Wdyt?

website/docs/procedural-macros/using.md
wawel37: https://github.com/software-mansion/scarb/issues/3028

website/docs/procedural-macros/writing.md
wawel37: https://github.com/software-mansion/scarb/issues/3028

website/docs/procedural-macros/using.md
Arcticae: It is because if you have to go to "advanced" to know about a feature you _should_ be enabling anyway - it's not optimal learning experience. That side is about publishing as well, not using per-se

website/docs/procedural-macros/writing.md
```diff
@@ -0,0 +1,183 @@
...
+ --> ../hello_world/src/lib.cairo:2:5
+    fib(16)
+    ^^^
+
+error: could not compile `hello_world` due to previous error
+```
+
+Maybe it's not the most productive code you wrote, but the function has been removed during the compilation.
+
+## Building token streams with `quote!`
```
Arcticae: This is treated like a second class citizen here - first you use an example where you don't use it. It's not a narrative i would like to push. I would introduce the section on `quote` along with the APIs of the macro definition

website/docs/procedural-macros/using.md
wawel37: https://github.com/software-mansion/scarb/issues/3029

website/docs/procedural-macros/writing.md
wawel37: Made it so the introduction example with the quote macro is the first visible in this doc file

website/docs/procedural-macros/examples.md
```diff
@@ -0,0 +1,531 @@
...
+) -> Result<String, Diagnostics> {
+    if node.kind(db) != SyntaxKind::SyntaxFile {
+        return Err(Diagnostics::new(Vec::new()).error("Expected SyntaxFile"));
+    }
+    let file = ast::SyntaxFile::from_syntax_node(db, node);
+    let items = file.items(db).elements_vec(db);
+    if items.len() != 1 {
+        return Err(Diagnostics::new(Vec::new()).error("Expected exactly one item"));
+    }
+    let func = items.into_iter().next().unwrap();
```
maciektr: I don't think we can unwrap in a macro example. Please show a diagnostic instead.

examples/procedural_macros/example3/macro/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+) -> Result<String, Diagnostics> {
+    if node.kind(db) != SyntaxKind::SyntaxFile {
+        return Err(Diagnostics::new(Vec::new()).error("Expected SyntaxFile"));
+    }
+    let file = ast::SyntaxFile::from_syntax_node(db, node);
+    let items = file.items(db).elements_vec(db);
+    if items.len() != 1 {
+        return Err(Diagnostics::new(Vec::new()).error("Expected exactly one item"));
+    }
+    let func = items.into_iter().next().unwrap();
```
maciektr: Vide supra

examples/procedural_macros/example1/cairo-repository/.tool-versions
maciektr, line 1: Please remove all .tool-versions from examples

examples/procedural_macros/example1/cairo-repository/Scarb.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "example1_app"
+version = "0.1.0"
+edition = "2023_11"
```
maciektr: NIT: Why not 2024_07?

examples/procedural_macros/example2/macro/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+cairo-lang-macro = "0.2"
+cairo-lang-primitive-token = "1"
```
maciektr, line 12: Do we need this, If we are not using the parser here? 🤔

scarb/tests/examples.rs
```diff
@@ -7,7 +7,7 @@ use snapbox::cmd::Command;
 use scarb_test_support::cargo::cargo_bin;
 use test_for_each_example::test_for_each_example;
 
-#[test_for_each_example(ignore = "dependencies")]
+#[test_for_each_example(ignore = "dependencies,procedural_macros")]
```
maciektr, line 11: Shouldn't we have a test case that builds macro examples?

examples/procedural_macros/example2/macro/Cargo.toml
wawel37: It's used and needed for quote [here](https://github.com/software-mansion/scarb/blob/2cb1f3c586da26170ffab81d3b619ac48fd0fe04/examples/procedural_macros/example2/macro/src/lib.rs#L15)

scarb/tests/examples.rs
wawel37: Added separate tests for proc macro docs as it's structure is a little bit different than other examples

scarb/tests/examples.rs
```diff
@@ -22,7 +22,53 @@ fn build(example: &Path) {
...
+                    .assert()
+                    .success();
+
+                Command::new(cargo_bin("scarb"))
+                    .arg("build")
+                    .current_dir(&cairo_repo)
+                    .assert()
+                    .success();
+            }
+        }
```
maciektr, line 62: Nit: I think it should be enough to run  ` Scarb::quick_command().arg("build")` in the cairo dir

PR #3022, opened 2026-02-27
Implement manifest validation with diagnostics output for LS

Arcticae:
Adds a layer of publishing manifest diagnostics via ndjson output mechanism

### DIFF
--- a/scarb/src/bin/scarb/commands/metadata.rs
+++ b/scarb/src/bin/scarb/commands/metadata.rs
@@ -61,13 +61,11 @@ pub fn run(args: MetadataArgs, config: &Config) -> Result<()> {
 }
 
 fn emit_manifest_diagnostic(config: &Config, error: &anyhow::Error) {
-    let file = error
-        .chain()
-        .find_map(|cause| {
-            cause
-                .downcast_ref::<ManifestParseError>()
-                .map(|error| error.path().to_string())
-        });
+    let file = error.chain().find_map(|cause| {
+        cause
+            .downcast_ref::<ManifestParseError>()
+            .map(|error| error.path().to_string())
+    });
 
     let span = error
         .chain()


scarb/src/bin/scarb/commands/metadata.rs
```diff
@@ -1,14 +1,45 @@
...
+    end: usize,
+}
+
 #[tracing::instrument(skip_all, level = "info")]
 pub fn run(args: MetadataArgs, config: &Config) -> Result<()> {
-    let ws = ops::read_workspace(config.manifest_path(), config)?;
+    let ws = match ops::read_workspace(config.manifest_path(), config) {
+        Ok(ws) => ws,
+        Err(error) => {
+            emit_manifest_diagnostic_if_json(config, &error);
```
wawel37: nit: i'd check the json output format here rather than naming the function this way only for a single check

scarb/src/bin/scarb/commands/metadata.rs
```diff
@@ -24,3 +55,45 @@ pub fn run(args: MetadataArgs, config: &Config) -> Result<()> {
...
+
+    let message = error
+        .chain()
+        .find(|cause| cause.downcast_ref::<ManifestParseError>().is_none())
+        .map(ToString::to_string)
+        .unwrap_or_else(|| error.to_string());
+
+    config
+        .ui()
+        .force_print(MachineMessage(ManifestDiagnosticMessage {
```
wawel37, line 90: nit: shouldn't we just go with `error()` instead of `force_print`

scarb/src/ops/workspace.rs
```diff
@@ -57,16 +58,15 @@ fn read_workspace_impl<'c>(
...
             this virtual manifest specifies a [dependencies] section, which is not allowed
             help: use [workspace.dependencies] instead
-        "#}))
-        .with_context(|| format!("failed to parse manifest at: {manifest_path}"))
-    } else {
-        Ok(())
+        "#}));
     }
+
+    Ok(())
```
wawel37, line 69: nit: no need for `return` statement

scarb/src/bin/scarb/commands/metadata.rs
```diff
@@ -24,3 +55,45 @@ pub fn run(args: MetadataArgs, config: &Config) -> Result<()> {
...
+    }
+
+    let file = error
+        .chain()
+        .find_map(|cause| {
+            cause
+                .downcast_ref::<ManifestParseError>()
+                .map(|error| error.path().to_string())
+        })
+        .unwrap_or_else(|| config.manifest_path().to_string());
```
maciektr: This may not be true. Config will point to the closest manifest. So if you are in a workspace package, and the error is in workspace manifest, you will show wrong location here.

scarb/src/bin/scarb/commands/metadata.rs
maciektr: Actually, if we want to call it programatically from LS, it needs to be forc e_print. LS runs scarb with --quite to suppress all ui, so if not force_print, you would not get it back.

scarb/tests/metadata.rs
```diff
@@ -111,6 +111,118 @@ fn includes_compilation_units() {
     );
 }
 
+#[test]
+fn emits_manifest_diagnostic_ndjson_for_invalid_manifest_in_json_mode() {
+    let t = TempDir::new().unwrap();
+    t.child("Scarb.toml")
+        .write_str(
+            "[package]
```
maciektr: nit: use indoc! please

maciektr (commented):
I really like this! Clean impl!

scarb/src/ops/workspace.rs
Arcticae: How come?

scarb/src/bin/scarb/commands/metadata.rs
Arcticae: I try to get a specific `ManifestParseError` which contains a path of the unparseable (semantically or syntactically) manifest. This will include child manifest, and fall back to the original one only if it can't find the root cause from the error chain (shouldn't happen but it might be possible)

scarb/tests/metadata.rs
Arcticae: Done

scarb/src/bin/scarb/commands/metadata.rs
Arcticae: Done

scarb/src/bin/scarb/commands/metadata.rs
maciektr: Why fallback at all then? It's just adding the risk of showing wrong path with no benefit.

scarb/src/ops/workspace.rs
maciektr: With a little magic of if-else statement :p

Arcticae: i personally like early-return style more if an `else` does not contain significant logic

Less cognitive load

scarb/src/bin/scarb/commands/metadata.rs
Arcticae: We will be able to display a diagnostic so user would be able to read it, so there is a benefit

scarb/src/ops/workspace.rs
maciektr: Yeah, both is fine. I don't think we have a strong preference in the code.

scarb/src/bin/scarb/commands/metadata.rs
maciektr: You can show the diagnostic. Just don't include a path 🤔

scarb/src/bin/scarb/commands/metadata.rs
```diff
@@ -24,3 +55,45 @@ pub fn run(args: MetadataArgs, config: &Config) -> Result<()> {
 
     Ok(())
 }
+
+fn emit_manifest_diagnostic_if_json(config: &Config, error: &anyhow::Error) {
+    if config.ui().output_format() != OutputFormat::Json {
```
maciektr: I am not sure that's what we want. If we fail to compute metadata, we should still display diagnostics to the user. Otherwise, you type some command and all you see is command exited without printing anything.

scarb/src/ops/workspace.rs
wawel37: Yup, it's a nit, resolve if u want. Im not strongly against this, just wanted to point out in case it was missed

scarb/src/bin/scarb/commands/metadata.rs
Arcticae: I can't show a diagnostic without a path 😬

maciektr: Why? 🤔

Arcticae: because every diagnostic has to be tied to a file. anyways i will add a case that will emit it without the file

maciektr (approved):

PR #3031, opened 2026-03-05
Run doc examples in parallel

maciektr:


maciektr:

- #3031 ◀
    - #3108
        - #3110
            - #3111

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -96,17 +99,22 @@ pub struct TestRunner<'a> {
     /// Target directory shared between all doc test runs.
     /// This allows speeding up doc tests compilation by sharing incremental caches.
     target_dir: TempDir,
+    /// Build profile passed to `scarb build` / `scarb execute` and used to locate the
+    /// incremental cache directory (`target/<profile>/`).
+    profile: String,
```
Arcticae, line 108: Why does it need it?

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -274,6 +359,52 @@ impl<'a> TestRunner<'a> {
...
+        }
+        let dst = profile_to.join(cache_dir);
+        fs::create_dir_all(&dst)
+            .with_context(|| format!("failed to create `{}`", dst.display()))?;
+        copy_dir_contents(&src, &dst)?;
+    }
+    Ok(())
+}
+
+fn copy_dir_contents(from: &Path, to: &Path) -> Result<()> {
```
Arcticae: isn't this a recursive dir copy? why do we reimplement such stuff?

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -117,48 +125,112 @@ impl<'a> TestRunner<'a> {
...
             "Running",
             &format!("{} doc examples for `{pkg_name}`", code_blocks.len()),
         ));
 
-        let mut idx = 0;
-        for block in code_blocks {
-            let strategy = block.run_strategy();
-            let total_in_item = *blocks_per_item.get(&block.id.item_full_path).unwrap_or(&1);
-            let display_name = block.id.display_name(total_in_item);
+        let execution_blocks: Vec<(usize, &CodeBlock)> = indexed_blocks
```
Arcticae: Why isn't this typed as `IndexedBlock` as well?

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -117,48 +125,112 @@ impl<'a> TestRunner<'a> {
...
-                                summary.failed += 1;
-                                self.ui.print(res.print_output.clone());
-                                self.ui.print(res.program_output.clone());
-                                self.ui.print(TestResult::failed(&display_name));
-                                failed_names.push(display_name);
-                            }
-                        },
-                        Err(e) => {
+                _ => match run_results.remove(&order) {
+                    Some(Ok(res)) => match res.status {
```
Arcticae: Three nested matches is a bit much for me

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -117,48 +125,112 @@ impl<'a> TestRunner<'a> {
...
+                block.run_strategy(),
+                order + 1,
+                self.target_dir.path(),
+            );
+            let build_succeeded = matches!(
+                &result,
+                Ok(r) if r.outcome != ExecutionOutcome::CompileError
+            );
+            run_results.insert(*order, result);
+            if build_succeeded {
```
Arcticae, line 175: just inline the condition

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -117,48 +125,112 @@ impl<'a> TestRunner<'a> {
...
+            .map(|(order, block)| (*order, *block))
+            .collect();
 
-            match strategy {
+        let mut run_results = HashMap::new();
+
+        // Run non-compile_fail blocks sequentially until the first successful compilation to warm
+        // the incremental cache. Compile_fail blocks are skipped here and always run in parallel.
+        for (order, block) in &execution_blocks {
+            if block.expected_outcome() == ExecutionOutcome::CompileError {
```
Arcticae: [nit]: Don't we want to filter out those beforehand? I would do it in a functional manner here

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -117,48 +125,112 @@ impl<'a> TestRunner<'a> {
...
+                break;
+            }
+        }
+
+        let parallel_results = execution_blocks
+            .iter()
+            .filter(|(order, _)| !run_results.contains_key(order))
+            .collect_vec()
+            .into_par_iter()
+            .map(|(order, block)| {
```
Arcticae, line 266: This could be extracted into its' own function

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -96,17 +99,22 @@ pub struct TestRunner<'a> {
...
+    /// Build profile passed to `scarb build` / `scarb execute` and used to locate the
+    /// incremental cache directory (`target/<profile>/`).
+    profile: String,
 }
 
 impl<'a> TestRunner<'a> {
     pub fn new(metadata: &'a AdditionalMetadata, has_lib_target: bool, ui: Ui) -> Result<Self> {
         let target_dir =
             tempdir().context("failed to create directory for doc tests target directory")?;
+        let profile = std::env::var("SCARB_PROFILE").unwrap_or_else(|_| "dev".to_string());
```
Arcticae, line 122: Why bother having it as a parameter here if we're getting this from env anyway

Arcticae (approved):
LGTM, with some style suggestions from me

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: It copies incremental cache to speed subsequent builds up - we need profile for it's location.

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: Yep, we can use fs_extra crate too

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: Changed

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: Refactored

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: I'd need to do a clone on result then 😓

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: We don't!
This loop is a trick we do - we compile the first block that compiles without an error, only for it to produce the incremental caches we can copy to all subsequent builds. 
So we cannot filter builds that are expected to fail out, as we still need to compile (and assert they actually fail) later only. We skip them here, as they won't produce incremental caches we won't to grab if they fail.

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: Done

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: I think having a centralized source for this is fine 🤔 Don't see a problem here

extensions/scarb-doc/src/doc_test/runner.rs
```diff
@@ -117,47 +128,85 @@ impl<'a> TestRunner<'a> {
...
+                    self.ui.print(TestResult::failed(&display_name));
+                    failed_names.push(display_name);
+                    self.ui.error(format!("Error running example: {:#}", e));
+                }
+                None => {
+                    summary.failed += 1;
+                    let display_name = Self::display_name(block, &blocks_per_item);
+                    self.ui.print(TestResult::failed(&display_name));
+                    failed_names.push(display_name);
+                    self.ui.error("Missing doc test run result.");
```
wawel37: shouldn't this arm be unreachable? Shouldn't we have the same number of `run_results` as indexed_blocks?

wawel37 (approved):

extensions/scarb-doc/src/doc_test/runner.rs
maciektr: It's not, replaced with call to unreachable instead

PR #3098, opened 2026-04-30
Implement warning about unknown fields

Arcticae:
This flow emits the diagnostics on succesful runs as well, so it will need an adjustment in LS as well to parse diagnostics on successful metadata runs

### DIFF
--- a/scarb/src/ops/workspace.rs
+++ b/scarb/src/ops/workspace.rs
@@ -13,12 +13,18 @@ use crate::MANIFEST_FILE_NAME;
 use crate::core::TomlManifest;
 use crate::core::config::Config;
 use crate::core::errors::{ManifestErrorWithSource, ManifestParseError};
+use crate::core::manifest::{
+    ManifestDiagnosticAnchor, ManifestDiagnosticMessage, ManifestDiagnosticSpan,
+    ManifestMessageKind, resolve_manifest_anchor,
+};
 use crate::core::package::Package;
 use crate::core::source::SourceId;
 use crate::core::workspace::Workspace;
 use crate::ops::find_workspace_manifest_path;
 use scarb_fs_utils as fsx;
 use scarb_fs_utils::{PathBufUtf8Ext, is_hidden};
+use scarb_ui::OutputFormat;
+use scarb_ui::components::MachineMessage;
 
 #[tracing::instrument(level = "debug", skip(config))]
 pub fn read_workspace<'c>(manifest_path: &Utf8Path, config: &'c Config) -> Result<Workspace<'c>> {
@@ -79,6 +85,7 @@ fn read_workspace_root<'c>(
         .with_context(|| format!("failed to read manifest at: {manifest_path}"))?;
     let toml_manifest = TomlManifest::read_from_str(&manifest_source)
         .map_err(|err| ManifestParseError::new(manifest_path, err))?;
+    warn_unknown_manifest_fields(manifest_path, &manifest_source, config);
     let toml_workspace = toml_manifest.get_workspace();
 
     let with_source = |err| {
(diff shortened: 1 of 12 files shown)


scarb/src/ops/workspace.rs
```diff
@@ -131,6 +138,7 @@ fn read_workspace_root<'c>(
             .map(AsRef::as_ref)
             .map(|package_path| {
                 let package_manifest = TomlManifest::read_from_path(package_path)?;
+                warn_unknown_manifest_fields(package_path, &package_manifest.raw_content, config);
```
piotmag769, line 141: This will emit the diagnostic twice for the runtime manifest, since you called it already before. Move the upper `warn_unknown_manifest_fields` to the `else` branch

scarb/src/ops/workspace.rs
```diff
@@ -186,6 +194,60 @@ fn read_workspace_root<'c>(
...
+                .ui()
+                .warn(format!("unknown manifest field `{path_str}`{location}"));
+        }
+    }
+}
+
+/// Converts a byte-offset span to a ` (path:line:col)` locator string.
+fn format_span_location(path: &Utf8Path, source: &str, span: &ManifestDiagnosticSpan) -> String {
+    let (line, col) = byte_offset_to_line_col(source, span.start);
+    format!(" ({path}:{line}:{col})")
```
piotmag769: [nit] weird place to add a space, better do it on the caller side

scarb/tests/manifest_warnings.rs
```diff
@@ -0,0 +1,314 @@
...
+        "#})
+        .unwrap();
+    t.child("src/lib.cairo").write_str("").unwrap();
+
+    Scarb::quick_command()
+        .arg("fetch")
+        .current_dir(&t)
+        .assert()
+        .success()
+        .stdout_eq("warn: unknown manifest field `unknown-section` ([..]/Scarb.toml:[..])\n");
```
piotmag769: I wouldn't skip the line:col part (don't replace it with [..]), this makes you not test spans

scarb/src/ops/workspace.rs
maciektr: I don't see how it's true. This loop iterates over the `workspace.members` list, so it does not include the workspace manifest itself 🤔

piotmag769: Mb then

utils/scarb-manifest-schema/src/lib.rs
```diff
@@ -19,9 +23,87 @@ pub fn get_shared_traverser() -> &'static SchemaTraverser {
...
 }
 
-///  Serializes full JSON Schema for the TomlManifest it into a serde_json::Value.
+/// Serializes full JSON Schema for the TomlManifest into a serde_json::Value.
 pub fn get_manifest_schema() -> Value {
-    serde_json::from_str(SCARB_SCHEMA_JSON).expect("Failed to serialize Manifest schema")
+    serde_json::from_str(SCARB_SCHEMA_JSON).expect("Failed to deserialize Manifest schema")
+}
+
+/// Deserializes the strict JSON Schema for the TomlManifest into a serde_json::Value.
```
maciektr: Nit: use either Serializes or Deserializes for both.

maciektr (approved):
Looks good to me!

scarb/src/ops/workspace.rs
wawel37: But if the workspace is a package at the same time, wouldn't that also duplicate the warnings? Maybe we should write a test case for that just to be sure

wawel37 (approved):
Other than this little doubt about warnings being duplicated when workspace is a package, everything looks fine. LGTM

PR #3115, opened 2026-05-08
Fix race condition in scarb cache clean

maciektr:


scarb/tests/cache.rs
```diff
@@ -47,3 +59,112 @@ fn path_print() {
...
+//
+// Correct behavior: while clean holds the package cache lock, `scarb fetch`
+// must block on lock acquisition and emit `flock.rs`'s "Blocking waiting
+// for file lock" message. The test asserts that, and therefore fails until
+// the lock is kept on a stable inode across `cache clean` (e.g. moved out
+// of the cache dir).
+#[cfg(unix)]
+#[test]
+#[timeout(60_000)]
+fn cache_clean_race_with_concurrent_lock_acquisition() {
```
wawel37, line 87: this test will not run `test_pause_hook` if not specified with `--feature` flag. Doesnt it mean that it will have unpredictable behaviour without this feature flag?

maciektr: The test will just fail with "scarb cache clean never reached the pause sentinel" error then

wawel37 (approved):

PR #3123, opened 2026-05-14
Add component cache fingerprints for check

maciektr:


maciektr:

- #3123 ◀

wawel37 (approved):
Seems good, but how about check -> build test case and check -> build -> check tests?

maciektr:
@wawel37 Added! My problem with this tests is, that it's very hard to assert if cache file has been used or not. But they can sure act as to detect some deserialization errors etc.

wawel37 (approved):

Arcticae (approved):

PR #3125, opened 2026-05-16
feat(scarb-prove): expose --proof-format flag for cairo_serde output

seritalien:
## Summary

Expose `ProofFormat::CairoSerde` on `scarb prove` via a new `--proof-format` flag, mirroring the upstream `stwo-cairo-prover` `prove` binary. Enables proof-composition pipelines (e.g. recursive aggregation in `vauban-zkpay`) where one Cairo program takes an `Array<CairoProof>` as input — currently impossible because `scarb prove` only writes `CairoProofForRustVerifier`-shaped JSON which loses the `ExtendedStarkProof` auxiliary data the in-Cairo verifier needs.

## Motivation

`stwo-cairo-prover` already supports `ProofFormat::CairoSerde` through `create_and_serialize_proof`. `scarb prove` hardcodes `ProofFormat::Json` at `extensions/scarb-prove/src/main.rs:108`. This adds one CLI flag with `default_value_t = ProofFormat::Json` so behavior is preserved for existing callers.

## Use case

Vauban Pay (`vauban-org/vauban-zkpay`) recursive aggregation: a `vauban_zkpay_aggregation` Cairo executable accepts K inner `CairoProof` instances via `--arguments-file`, calls `stwo_cairo_air::verify_cairo` K times, then is itself proven by Stwo to produce a single outer STARK that `VaubanSettlement` verifies on Starknet. Without `cairo_serde` output from `scarb prove`, the K inner proofs cannot be marshaled into the aggregation's args-file without reimplementing the prover offline.

## Behavior

- Default unchanged (`--proof-format json`).
- `--proof-format cairo-serde` writes `["0x...", "0x...", ...]`.
- `scarb verify` already accepts both formats via `cairo_air::utils::ProofFormat` round-trip.

## Compat

Backwards-compatible. No breaking changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7432,7 +7432,6 @@ version = "2.18.0"
 dependencies = [
  "anyhow",
  "assert_fs",
- "cairo-air",
  "camino",
  "clap",
  "create-output-dir 1.0.0",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7295,6 +7295,7 @@ name = "scarb-extensions-cli"
 version = "2.18.0"
 dependencies = [
  "anyhow",
+ "cairo-air",
  "cairo-vm",
  "camino",
  "clap",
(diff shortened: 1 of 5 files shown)


maciektr (approved):

maciektr:
Thank you @seritalien !

PR #3127, opened 2026-05-19
Introduce machine-readable manifest diagnostics

Arcticae:

PR #3128, opened 2026-05-19
Emit structured compiler diagnostics in JSON output

Arcticae:


Arcticae:

- #3127
    - #3128 ◀
        - #3129

scarb/src/compiler/structured_diagnostics/core.rs
```diff
@@ -0,0 +1,246 @@
...
+
+impl StructuredDiagnosticsReporter {
+    pub fn new(ignore_warnings_crate_ids: Vec<CrateInput>, crates: Vec<CrateInput>) -> Self {
+        Self {
+            ignore_warnings_crate_ids,
+            crates,
+        }
+    }
+
+    pub fn check(&mut self, db: &dyn Database, sink: &mut impl StructuredDiagnosticsSink) -> bool {
```
maciektr, line 72: Can this be upstreamed to the compiler?

maciektr (changes requested):

scarb/src/compiler/structured_diagnostics/core.rs
Arcticae: Yes - this whole file is written in such a way to be copied-over directly without modifications once we verify we want to go forward with this approach.

Arcticae: (that's why there's the StructuredDiagnosticsSink abstraction present here)

scarb/src/compiler/structured_diagnostics/core.rs
```diff
@@ -0,0 +1,246 @@
...
+            span: self.span,
+        }
+    }
+}
+
+fn build_structured_diagnostic_message<'db, TEntry: DiagnosticEntry<'db>>(
+    db: &'db dyn Database,
+    entry: &TEntry,
+    file_notes: &PluginFileDiagnosticNotes<'db>,
+) -> Option<StructuredDiagnosticMessage> {
```
wawel37, line 209: Correct me if im wrong, but this function never returns `None` right? If so change the return type

scarb/src/compiler/structured_diagnostics/core.rs
```diff
@@ -0,0 +1,246 @@
...
+            if skip_warnings && entry.severity() == Severity::Warning {
+                continue;
+            }
+
+            if let Some(message) = build_structured_diagnostic_message(db, &entry, file_notes) {
+                sink.emit(message);
+                found |= group.check_error_free().is_err();
+            }
+        }
+        found
```
wawel37, line 163: Doesn't `check_error_free` check the whole group? If so whats the point of running it every time in loop. Once at the end is enough i think

wawel37 (approved):
Other than some nits seems fine

PR #3129, opened 2026-05-19
Parallelize structured compiler diagnostic collection

Arcticae:


Arcticae:

- #3127
    - #3128
        - #3129 ◀

scarb/src/compiler/structured_diagnostics/core.rs
```diff
@@ -69,85 +43,178 @@ impl StructuredDiagnosticsReporter {
...
+            return self.check_sequentially(db, sink);
+        }
+
+        let reports = self
+            .crates
+            .par_iter()
+            .map_with(db.dyn_clone(), |db, crate_input| {
+                self.collect_crate_diagnostics(db.as_ref(), crate_input)
+            })
+            .collect::<Vec<_>>();
```
maciektr: Why should we do this instead of the warmup style that Cairo does?
https://github.com/starkware-libs/cairo/blob/adf04c14281084d7ef7d383c8418888a5d49d992/crates/cairo-lang-compiler/src/lib.rs#L209

maciektr: Also, would it make sense to paralelize on module level like Cairo does?

maciektr (changes requested):

scarb/src/compiler/structured_diagnostics/core.rs
Arcticae: Done

scarb/src/compiler/structured_diagnostics/core.rs
```diff
@@ -68,11 +40,45 @@ impl StructuredDiagnosticsReporter {
...
+}
+
+impl StructuredDiagnosticMessage {
+    fn error(message: String, file: String) -> Self {
+        Self::new(
+            MachineDiagnosticKind::Diagnostic,
+            message,
+            MachineDiagnosticSeverity::Error,
+            file,
+            MachineDiagnosticSpan { start: 0, end: 0 },
```
wawel37, line 52: Same as 2 prs before, should be a default

PR #3132, opened 2026-05-21
v2.19.0-rc.0

wawel37:


Cargo.lock
```diff
@@ -8259,6 +8295,23 @@ dependencies = [
  "zeroize",
 ]
 
+[[package]]
+name = "starknet-types-core"
```
Arcticae, line 8299: There are 2 different versions in the graph now, why?

scarb/tests/build_starknet_contract_allowed_libfuncs.rs
```diff
@@ -142,7 +142,7 @@ fn pass_named_list() {
         .version("0.1.0")
         .manifest_extra(indoc! {r#"
             [[target.starknet-contract]]
-            allowed-libfuncs-list.name = "experimental"
+            allowed-libfuncs-list.name = "all"
```
Arcticae: Why?

scarb/src/compiler/compilers/starknet_contract/validations.rs
```diff
@@ -77,7 +77,7 @@ pub fn check_allowed_libfuncs<'db>(
                 // instructing how to do this. Otherwise, we know that user knows what they
                 // do, so we do not clutter compiler output.
                 if list_selector == Default::default() {
-                    let experimental = BUILTIN_EXPERIMENTAL_LIBFUNCS_LIST;
+                    let experimental = BUILTIN_ALL_LIBFUNCS_LIST;
```
Arcticae: Why change that?

wawel37: Same as above, the `BUILTIN_EXPERIMENTAL_LIBFUNCS_LIST` was removed from compiler.

scarb/tests/build_starknet_contract_allowed_libfuncs.rs
wawel37: The `all` list name was completely removed from compiler. Changed it to `audited` to be more specific and safer this way.

Cargo.lock
wawel37: Cuz `cairo-annotations` uses 1.0.0 `starknet-types-core`, but the compiler uses 0.2.4

scarb/tests/build_starknet_contract_allowed_libfuncs.rs
Arcticae: But this defeats the point of the message i think, does it ever happen again that something is not allowed in default list and the enabling aforementioned list fixes it?

wawel37: changed it back to `all`, as it's an only one list that makes sense to be proposed to a user, as the `audited` is the default fallback

Cargo.lock
Arcticae: How are the types used now? Won't there be any runtime issues with that?

PR #3134, opened 2026-05-26
Fix scarb check behaviour for proc macros

maciektr:


scarb/tests/proc_macro_v2_build.rs
```diff
@@ -101,6 +101,46 @@ fn can_check_cairo_project_with_plugins() {
...
+    WorkspaceBuilder::start()
+        .add_member("uses_macro")
+        .add_member("no_macro")
+        .build(&project);
+    // Check only `no_macro` — the proc macro should be skipped entirely.
+    Scarb::quick_command()
+        .arg("check")
+        .arg("-p")
+        .arg("no_macro")
+        .env("CARGO_TERM_QUIET", "true")
```
wawel37, line 134: Do we need that here?

wawel37 (commented):
correct me if im wrong, but there is no tests for the proc macro being checked when there is no dependency on it

scarb/tests/proc_macro_v2_build.rs
maciektr: Yep, resolver runs cargo fetch for all the macros in the workspace 
```
--- Expected
++++ actual:   stdout
   1      - [..]Checking no_macro v1.0.0 ([..]Scarb.toml)
   2      - [..]Finished checking `dev` profile target(s) in [..]
        1 +     Updating crates.io index
        2 +      Locking 20 packages to latest compatible versions
        3 +       Adding bumpalo v3.17.0 (available: v3.20.3)
        4 +     Checking no_macro v1.0.0 (/private/var/folders/2t/7dt08_ss585ct08q57mvw_q40000gn/T/.tmpGDqFO8/project/no_macro/Scarb.toml)
        5 +     Finished checking `dev` profile target(s) in 12 seconds

```

https://github.com/software-mansion/scarb/blob/203da4d6a36bf9a6d4fcf158f5f56df77fc6ba48/scarb/src/ops/resolve.rs#L361

piotmag769 (approved):

piotmag769 (approved):

wawel37 (approved):

PR #3159, opened 2026-06-26
Pull the latest available version of `snforge_std` from the registry in CI

integraledelebesgue:


Arcticae (changes requested):
This is not really a solution is it? It's an error on the snforge side

PR #3160, opened 2026-06-29
Expose `extra` data to avoid consumer-side Scarb.toml re-parsing logic

Arcticae:


scarb/src/core/manifest/diagnostic.rs
```diff
@@ -18,6 +18,8 @@ pub struct ManifestDiagnosticMessage {
     pub span: Option<ManifestDiagnosticSpan>,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub related: Vec<ManifestRelatedLocation>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub data: Option<ManifestDiagnosticExtraData>,
```
wawel37, line 22: nit:

```suggestion
    pub extra Option<ManifestDiagnosticExtraData>,
```

wawel37 (approved):

maciektr:
~~don't understand what are we achieving here :thinking:~~
Found the ls pr https://github.com/software-mansion/cairols/pull/1346

scarb/src/core/manifest/diagnostic.rs
maciektr: "data" makes sense here, as it maps to "Diagnostic" from lsp-types

scarb/src/core/manifest/diagnostic.rs
```diff
@@ -74,6 +76,40 @@ pub struct ManifestDiagnosticData {
     pub span: Option<ManifestDiagnosticSpan>,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub related: Vec<ManifestRelatedLocation>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub extra: Option<ManifestDiagnosticExtraData>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(untagged)]
```
maciektr, line 84: Maybe we should tag it? Matching on field names seems weak to me 🤔 

Image

e.g. this action will be pushed for any diagnostic with "field_path" in extra, right?

scarb/tests/metadata.rs
```diff
@@ -291,6 +291,266 @@ fn emits_manifest_diagnostic_for_semantic_manifest_error_without_span() {
...
+    );
+    assert_eq!(
+        diagnostic["data"]["skip_optimizations_path"],
+        serde_json::json!(["profile", "dev", "cairo", "skip-optimizations"])
+    );
+}
+
+#[test]
+fn emits_manifest_diagnostic_data_for_dependency_git_ref_without_git() {
+    for (name, dependency, error_code, field, fields) in [
```
maciektr: nit: Don't like this for -> assert pattern

scarb/src/core/manifest/diagnostic.rs
Arcticae: i'm going to leave as is just because of laziness: 
i'd need another dev version, changes all across the board etc. because of that field rename

scarb/src/core/manifest/diagnostic.rs
Arcticae: No - they are matched on diagnostic types

maciektr: Ah, by codes? This is fine then

maciektr (approved):

scarb/tests/metadata.rs
Arcticae: I agree - [Re-write the test to test-case pattern](https://github.com/software-mansion/scarb/pull/3160/commits/c3d5cbf324e25905eaa62f46a4faf9215f259772)

PR #3177, opened 2026-07-15
fix confusing error when registry retrurns a non-JSON body

### LINKED ISSUE #3175: Properly expose 429 errors to users when uploading packages

hakiers:
Fixes #3175

When publishing a package, if the registry's upload endpoint returned an empty body. Scarb tried to parse it as JSON and failed. That parse error hid the real HTTP status, so users saw a confusing message instead of the actual reason:

error: error decoding response body

Caused by:
    expected value at line 1 column 1

## Fix

In publish, I changed `response.json().await?` to `response.json().await.unwrap_or_default()`, so an empty body yields Null instead of aborting. The real status code now reaches the user:

error: upload failed with status code: `429 Too Many Requests`, `missing error field in the registry response`

## Changes

- http.rs: fall back to unwrap_or_default() when parsing the error body.
- simple_http_server.rs: changed HttpPostResponse.message from String to Option<String> so the test mock can send an empty body (None), and added a test for the 429 empty-body case.

### DIFF
--- a/scarb/src/core/registry/client/http.rs
+++ b/scarb/src/core/registry/client/http.rs
@@ -192,7 +192,7 @@ impl RegistryClient for HttpRegistryClient<'_> {
             StatusCode::OK => RegistryUpload::Success,
             status => {
                 let headers = response.headers().clone();
-                let error_body: serde_json::Value = response.json().await?;
+                let error_body: serde_json::Value = response.json().await.unwrap_or_default();
                 let error_message = error_body
                     .get("error")
                     .and_then(|e| e.as_str())
--- a/scarb/tests/registry_publish.rs
+++ b/scarb/tests/registry_publish.rs
@@ -14,7 +14,7 @@ fn publish() {
     // 200 -> StatusCode::OK
     let registry = HttpRegistry::serve(Some(HttpPostResponse {
         code: 200,
-        message: "published".to_string(),
+        message: Some("published".to_string()),
     }));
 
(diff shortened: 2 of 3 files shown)


scarb/src/core/registry/client/http.rs
```diff
@@ -192,7 +192,7 @@ impl RegistryClient for HttpRegistryClient<'_> {
             StatusCode::OK => RegistryUpload::Success,
             status => {
                 let headers = response.headers().clone();
-                let error_body: serde_json::Value = response.json().await?;
+                let error_body: serde_json::Value = response.json().await.unwrap_or_default();
```
piotmag769, line 195: Just making sure we are fine with the `missing error field in the registry response` follow up message. Isn't it better to just do `unwrap_or_default` in line 199 (if it was consulted - leave it as it is)?

hakiers: It covers two different cases:
- **Line 195** handles the body not being valid JSON at all. With the old `response.json().await?` this aborted the whole function when there was no body. That was the bug.
- **Line 199** handles a valid JSON body that just lacks an error field.

I agree with you that `missing error field in the registry response` might be slightly imprecise. I think we can drop that message.

piotmag769 (approved):

Arcticae (approved):

maciektr (approved):
Looks good! Did you manage to test it against 429 errors from real registry? 🤔

PR #3183, opened 2026-07-20
Use unbounded channels for internal communication in proc-macro-server

### LINKED ISSUE #101: Reruns often fail with timeout
### Problem

We have a reccurring issue with our test runs timing out and it probably should be addressed because it can cause frustration during releases that need to be tested fast 

https://github.com/software-mansion/maat/actions/runs/26348754670
https://github.com/software-mansion/maat/actions/runs/26348752827
https://github.com/software-mansion/maat/actions/runs/25616720992

### Proposed Solution

_No response_

### Notes

_No response_

integraledelebesgue:
Related to https://github.com/software-mansion/maat/issues/101

The current flow in proc-macro-server is:
stdin ──▶ reader thread ──bounded(0)──▶ N worker threads ──bounded(0)──▶ writer thread ──▶ stdout

Bounded channels make the whole communication synchronous. When all workers are busy, the reader stops consuming the `stdin`. On the other hand, when `stdout` is full and not being consumed by the LS, workers cannot push their results and free-up. **As a consequence, the throughput of the whole PMS is limited by the rate of consumption on the LS side.**

This paired with some quirks in the LS' event loop results in a deadlock.

A sibling PR in CairoLS https://github.com/software-mansion/cairols/pull/1367 adds another layer of buffering and explains the deadlock.

### DIFF
--- a/scarb/src/ops/proc_macro_server/connection.rs
+++ b/scarb/src/ops/proc_macro_server/connection.rs
@@ -13,8 +13,8 @@ pub struct Connection {
 
 impl Connection {
     pub fn new() -> Self {
-        let (reader_sender, reader_receiver) = crossbeam_channel::bounded(0);
-        let (writer_sender, writer_receiver) = crossbeam_channel::bounded(0);
+        let (reader_sender, reader_receiver) = crossbeam_channel::unbounded();
+        let (writer_sender, writer_receiver) = crossbeam_channel::unbounded();
 
         let reader = std::thread::spawn(move || {
             let stdin = std::io::stdin();


Arcticae (approved):

scarb/src/ops/proc_macro_server/connection.rs
```diff
@@ -13,7 +13,7 @@ pub struct Connection {
 
 impl Connection {
     pub fn new() -> Self {
-        let (reader_sender, reader_receiver) = crossbeam_channel::bounded(0);
+        let (reader_sender, reader_receiver) = crossbeam_channel::unbounded();
         let (writer_sender, writer_receiver) = crossbeam_channel::bounded(0);
```
piotmag769: Do we want to do the same here?

integraledelebesgue: Yes, this will add another stage of buffering, which won't hurt us

piotmag769 (approved):

PR #3211, opened 2026-08-05
ci: bind workflow inputs via env before shell

SashaMIT:
## Summary
- Bind `inputs.is_dev` through `env:` before the nightly version shell step.
- Bind `inputs.full-verify` through `env:` before the archive verify shell step (instead of embedding a GitHub expression in the script).
- Same class as GitHub’s documented script-injection guidance for interpolating workflow inputs into `run:` scripts.

## Test plan
- [ ] Nightly/dev version suffix logic unchanged when `is_dev=true`
- [ ] Verify archive still passes `--full` only when `full-verify` is true

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/_build-release.yml
+++ b/.github/workflows/_build-release.yml
@@ -174,11 +174,16 @@ jobs:
 
       - name: Verify archive
         shell: bash
-        run: |
-          archive=$(find target/verify -name 'scarb-*.zip' -o -name 'scarb-*.tar.gz')
-          cargo xtask verify-archive --archive "$archive" ${{ inputs.full-verify && '--full' || ''}}
         env:
           EXPECTED_VERSION: ${{ inputs.scarb-tag }}
+          FULL_VERIFY: ${{ inputs.full-verify }}
+        run: |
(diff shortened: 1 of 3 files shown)


maciektr (changes requested):
Hi Sasha!
Thanks for interest in contributing!
To make this change more meaningful, how about we tackle this more generally and disallow patterns like this with static analysis?
We have zizmor configured for the repo here: https://github.com/software-mansion/scarb/blob/main/.github/workflows/zizmor.yml

SashaMIT:
Good call — done both halves: kept the env-binding fix for the live instances (zizmor flags but doesn't fix), and widened the zizmor workflow trigger to all of .github/workflows/*.yml so the class is disallowed repo-wide going forward (previously only the three release workflows were analyzed). New commit on this branch.

Arcticae (approved):

maciektr (approved):

SashaMIT:
Thanks Maciej. Glad the env-bind plus the zizmor widen landed.

PR #3237, opened 2026-08-26
Publish plugins fix

### LINKED ISSUE #3232: Publish packages as a separate workflow
[Publish plugins](https://github.com/software-mansion/scarb/blob/main/.github/workflows/publish_plugins.yml) used to be a workflow that would trigger and run only when the draft release was published (as it needs new scarb binary). We've changed that so the release workflow is running now the publish plugins within it, which is not correct, as at the time of publish plugins firing, we only have the draft release (which is causing problems for publish plugins job, as the binaries are not yet visible).

AC:
- Make `Publish plugins` workflow as a separate one which fires on new release being published as it was before the change.
- Make sure that the slack notification about publish plugins failing still shows up on the slack channel.

hakiers:
#3232 
This pull request updates the GitHub Actions workflows related to publishing built-in plugins. The main change is to trigger the `publish_plugins` workflow automatically when a new release is published, instead of invoking it from the `release.yml` workflow. It also adds required permissions to the workflow configuration.

**Workflow automation improvements:**

* The `publish_plugins.yml` workflow is now triggered automatically on GitHub release publication (`release.published`), in addition to manual dispatch, making the plugin publishing process more streamlined.
* The `publish_plugins` job invocation was removed from `release.yml`, so it is no longer triggered as a dependent job after drafting a release.

**Security and configuration:**

* Added explicit `contents: read` permissions to the `publish_plugins.yml` workflow for better security and compliance with GitHub Actions best practices.

### DIFF
--- a/.github/workflows/publish_plugins.yml
+++ b/.github/workflows/publish_plugins.yml
@@ -1,5 +1,8 @@
 name: Publish Built-in Plugins
 on:
+  release:
+    types:
+      - published
   workflow_dispatch:
     inputs:
       version_tag:
@@ -19,6 +22,9 @@ on:
         required: true
       SLACK_PUBLISH_PLUGINS_FAILURE_WEBHOOK_URL:
         required: true
+
+permissions:
+  contents: read
 
 jobs:
   publish:
--- a/.github/workflows/publish_plugins.yml
+++ b/.github/workflows/publish_plugins.yml
@@ -9,19 +9,6 @@ on:
         required: false
         type: string
         description: Scarb version to publish from.
-  workflow_call:
-    inputs:
-      version_tag:
-        required: false
-        type: string
-        description: Scarb version to publish from.
-    secrets:
-      SCARB_REGISTRY_AUTH_TOKEN:
-        required: true
-      SCARB_DEV_REGISTRY_AUTH_TOKEN:
-        required: true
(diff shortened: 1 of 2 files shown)


maciektr (approved):

.github/workflows/publish_plugins.yml
maciektr, line 1: Do we still need the workflow_call?

.github/workflows/release.yml
```diff
@@ -68,16 +68,4 @@ jobs:
...
-    permissions:
-      contents: read
-    uses: ./.github/workflows/publish_plugins.yml
-    with:
-      version_tag: ${{ github.ref_name }}
-    secrets:
-      SCARB_REGISTRY_AUTH_TOKEN: ${{ secrets.SCARB_REGISTRY_AUTH_TOKEN }}
-      SCARB_DEV_REGISTRY_AUTH_TOKEN: ${{ secrets.SCARB_DEV_REGISTRY_AUTH_TOKEN }}
-      SLACK_PUBLISH_PLUGINS_FAILURE_WEBHOOK_URL: ${{ secrets.SLACK_PUBLISH_PLUGINS_FAILURE_WEBHOOK_URL }}
+
```
Arcticae:
```suggestion


```

Issue #3231, opened 2026-08-21
Check `cairo-lang-*` versions during release


wawel37:
During release flow, we want to also check whether the all the `cairo-lang-*` package versions were set to specific version, not the directly git revs. 

AC:
- Add this check to `release.yml`

PR #3245, opened 2026-08-31
Add cairo-lang check to workflow

hakiers:
#3231 
This pull request introduces a new workflow to ensure that all `cairo-lang-*` crates used in the project resolve to a single, registry-sourced version in `Cargo.lock`. It integrates this check into the release process to catch dependency inconsistencies early and prevent accidental use of multiple versions or non-registry sources for these critical crates.

**Dependency consistency enforcement:**

* Added a new workflow file, `.github/workflows/_check-cairo-lang-versions.yml`, which fetches the list of `cairo-lang-*` crates from the appropriate Cairo release and verifies that each resolves to a single version and source in `Cargo.lock`, failing the workflow if any inconsistencies are found.
* Integrated the new `check-cairo-lang-versions` job into the main release workflow (`.github/workflows/release.yml`), ensuring it runs alongside existing checks before a release is drafted. [[1]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R15-R19) [[2]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L24-R29)

### DIFF
--- a/.github/workflows/_check-release.yml
+++ b/.github/workflows/_check-release.yml
@@ -85,3 +85,18 @@ jobs:
         run: |
           cargo test -p scarb --profile=ci --test main proc_macro_v1_prebuilt::
           cargo test -p scarb --profile=ci --test main proc_macro_v2_prebuilt::
+
+  ciaro-lang-check:
+    name: ciaro-lang check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Fail if cairo-lang uses wildcard versions in Cargo.toml
+        run: |
+          ! grep -E 'cairo-lang-.*=\s*"\*"' Cargo.toml 
+          
+      - name: Fail if cairo-lang uses direct git dependencies in Cargo.toml
+        run: |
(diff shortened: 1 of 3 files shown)


wawel37 (approved):

.github/workflows/_check-release.yml
```diff
@@ -85,3 +85,18 @@ jobs:
...
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Fail if cairo-lang uses wildcard versions in Cargo.toml
+        run: |
+          ! grep -E 'cairo-lang-.*=\s*"\*"' Cargo.toml 
+          
+      - name: Fail if cairo-lang uses direct git dependencies in Cargo.toml
+        run: |
+          ! grep -E 'cairo-lang-.*=\s*\{.*git"' Cargo.toml
+
```
maciektr: This should check Cargo.lock instead of Cargo.toml. 

Cargo can pull the same package in a) two incompatible versions or b) coming from two incompatible sources into the same dependency tree.

maciektr: nit: it's not probable, but you left path deps out :)

maciektr (changes requested):

PR #3247, opened 2026-09-01
Run maat experiments after release

hakiers:
This pull request adds a new job to the GitHub Actions workflow to automate triggering an experiment in the `maat` repository after a plugin is published. The main focus is to streamline the release process by integrating it with downstream experiments.

**Workflow automation:**

* Added a new `trigger-maat-experiment` job to `.github/workflows/publish_plugins.yml` that runs after the `publish` job completes. This job extracts the released version from the tag, strips any leading 'v', and uses the GitHub CLI to trigger the `experiment.yml` workflow in the `software-mansion/maat` repository, passing relevant parameters such as the release version and a notification flag.


maciektr (changes requested):
This does not really solve the task, as it asked for experiments starting before release is published: 

> (...) experiments started automatically before our releases, e.g. when a pull request from a release branch is submitted in Scarb (...)

And your approach is retroactive. 

Please let me know if we agreed to loosen the requirements.

hakiers:
@maciektr I discussed this solution with @wawel37. We need to release a new version first to be able to run the experiment.

maciektr:
@hakiers I think the initial idea suggested by the issue was to lift that restriction and start running experiments before we publish the release, so we can make an informed decision when publishing it. 

I guess triggering the experiments automatically after the release is nice to have, though I don't really think it brings a lot to the table - at best it saves two clicks on the gh actions ui. 

When we merge this, I would keep the issue and maybe bring it up on another backlog grooming to see if we want this implemented or not.

hakiers:
@maciektr How can we run an experiment without a release? If I understand correctly, Maat download the version of scarb which we want to run the experiment on, which is only possible because that version has already been released

.github/workflows/publish_plugins.yml
```diff
@@ -131,3 +131,30 @@ jobs:
...
+      - name: Trigger Maat experiment
+        env:
+          GH_TOKEN: ${{ secrets.MAAT_EXPERIMENTS_DISPATCH_TOKEN }}
+        run: |
+          gh workflow run experiment.yml \
+            --repo software-mansion/maat \
+            --ref main \
+            -f workspace="release" \
+            -f scarb="$VERSION" \
+            -f notify_team="true"
```
wawel37: imho we should put it in a separate workflow that observe if the new version was released. putting it here works only on the assumption that the publish plugin run when the release is published

wawel37 (changes requested):

maciektr (approved):

wawel37 (approved):

PR #3250, opened 2026-09-02
fix: validate locked dependency sources

### LINKED ISSUE #3184: Resolver lockfile-based shortcuts may use sources incompatible with manifest spec
Repro - core patches:
1) git clone https://github.com/starkware-libs/cairo.git
2) cd cairo/corelib
3) Add snforge_std to dev deps
4) scarb build  - should get error:
```
error: found dependencies on the same package `core` coming from incompatible sources:
source 1: /Users/franciszekjob/testing/cairo/corelib/Scarb.toml
source 2: std
```
5) Add patch to manifest
```
[patch."scarb:/std"]
core = { path = "." }
```
6) scarb build  -> should pass
7) scarb clean && scarb cache clean and remove patch
8) scarb build  -> still pases, even though we use core from std not the patch anymore

 scarb --version
scarb 2.19.0+dev-2026-07-08 (e78b99292 2026-07-08)
cairo: 2.19.0 (431559058)
sierra: 1.9.0
arch: aarch64-apple-darwin

zakazaka95:
## Summary
- skip the lockfile dependency shortcut when a path dependency has no recorded source
- fetch the selected registry package summary before rebuilding edges from the current patch map
- cover removing a `[patch]` entry after the lockfile has been created

Closes #3184

## Testing
- `cargo fmt --all`
- `cargo test -p scarb --test main patch::patch_core_with_self -- --exact --nocapture`
- `git diff --check`

### DIFF
--- a/scarb/src/resolver/provider.rs
+++ b/scarb/src/resolver/provider.rs
@@ -182,7 +182,10 @@ impl PubGrubDependencyProvider {
     pub fn request_dependency(&self, dependency: ManifestDependency) {
         if (self.require_audits
             || !dependency.source_id.is_registry()
-            || !self.lockfile.locks_dependency(dependency.clone()))
+            || self
+                .locked_registry_package(&dependency)
+                .and_then(|locked| self.complete_locked_dependencies(locked))
+                .is_none())
             && self
                 .state
                 .index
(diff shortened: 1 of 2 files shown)


scarb/src/resolver/provider.rs
```diff
@@ -334,6 +337,48 @@ impl PubGrubDependencyProvider {
...
+        let locked_dependency_names = dependencies
+            .iter()
+            .map(|dependency| &dependency.name)
+            .collect::<HashSet<_>>();
+
+        (locked_dependency_names == dependency_names
+            && dependencies
+                .iter()
+                .all(|dependency| dependency.source.is_some()))
+        .then_some(dependencies)
```
wawel37, line 380: `locked.dependencies` stores only names, so if the lockfile has two entries for the same name (the `SmallVec` comment says this is intentionally supported), both land in the returned `Vec` and map to the same `PubGrubPackage` key in `get_dependencies` — the later one silently overwrites the earlier, pinning the wrong version. Make sure `dependencies.len()` and `dependency_names.len()` are equal maybe

wawel37 (changes requested):

scarb/src/resolver/provider.rs
zakazaka95: Added the length guard in 0a4db94, so ambiguous names fall back to the current package summary. The regression covers both version orders, valid unique entries, unrelated packages, missing entries and source-less entries. It fails without the guard and passes with it; scoped Clippy and formatting pass too.

scarb/src/resolver/provider.rs
```diff
@@ -334,6 +337,49 @@ impl PubGrubDependencyProvider {
...
+    fn complete_locked_dependencies<'a>(
+        &'a self,
+        locked: &PackageLock,
+    ) -> Option<Vec<&'a PackageLock>> {
+        let dependency_names = locked.dependencies.iter().collect::<HashSet<_>>();
+        let dependencies = self
+            .lockfile
+            .packages()
+            .filter(|package| dependency_names.contains(&package.name))
+            .collect_vec();
```
wawel37, line 370: shouldn't we also match by the version, not only the name?

wawel37 (changes requested):

wawel37:
continuing this one in https://github.com/software-mansion/scarb/pull/3255

PR #3252, opened 2026-09-03
Check cairo-lang-* versions during release

### LINKED ISSUE #3231: Check `cairo-lang-*` versions during release

hakiers:
## Description:
Adds a release-gate check that fails if any `cairo-lang-*` crate resolves to more than one version, or to a non-registry (e.g. git) source, in `Cargo.lock` — catching leftover `[patch.crates-io]`/git-pinned dependencies from active Cairo development that should have been reverted before release.

- New reusable workflow `_check-cairo-lang-versions.yml`, wired into `release.yml`.
- Crate names to check are fetched from `crates_list.sh` in the matching (or newest stable) `starkware-libs/cairo` release, so the list doesn't need manual upkeep.
- Uses `cargo metadata`, so it checks the actual resolved graph, not just `Cargo.toml` constraints - correctly ignoring local path crates like `cairo-lang-macro`.
- Removes the old, ineffective grep-based check in `_check-release.yml`.

Closes #3231.

### DIFF
--- a/.github/workflows/_check-cairo-lang-versions.yml
+++ b/.github/workflows/_check-cairo-lang-versions.yml
@@ -18,8 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          persist-credentials: false
 
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
 
       - name: Fetch scripts/crates_list.sh from the matching Cairo release, or the newest one if it doesn't exist
         id: fetch
@@ -77,11 +79,16 @@ jobs:
           # A crate fails the check if it resolved to more than one version, more
           # than one source, or any source that isn't the crates.io registry
           # (e.g. a leftover git dependency).
-          jq -e '
+          bad=$(jq -c '
             map(select(
               (.versions | length) > 1
               or (.sources | length) > 1
               or (.sources[0] | startswith("registry+") | not)
             ))
-            | length == 0
-          ' <<< "$summary"
+          ' <<< "$summary")
+
+          if [ "$bad" != "[]" ]; then
+            echo "Cairo toolchain crates do not resolve to a single, registry-sourced version in Cargo.lock:"
+            jq . <<< "$bad"
+            exit 1
+          fi
--- a/.github/workflows/_check-cairo-lang-versions.yml
+++ b/.github/workflows/_check-cairo-lang-versions.yml
@@ -1,0 +1,87 @@
+name: Check Cairo-lang Versions
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        required: false
+        type: string
+  workflow_call:
+    inputs:
+      ref:
+        required: false
+        type: string
+
+jobs:
+  check:
+    name: cairo-lang versions
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - uses: dtolnay/rust-toolchain@stable
+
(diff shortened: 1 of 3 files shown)


wawel37 (approved):

.github/workflows/_check-cairo-lang-versions.yml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Fetch scripts/crates_list.sh from the matching Cairo release, or the newest one if it doesn't exist
+        id: fetch
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REF: ${{ inputs.ref }}
+        run: |
+          tag=$REF
+
+          if ! gh api repos/starkware-libs/cairo/releases/tags/$tag --jq .tag_name >/dev/null 2>&1; then
+            tag=$(gh api repos/starkware-libs/cairo/releases/latest --jq .tag_name)
```
maciektr: I don't think this is correct. We want to match Cairo tag we use to build Scarb release. 

Say we release a bug fix Scarb release of v2.19.5, using Cairo v2.19.4. 
This would check for v2.19.5 tag in Cairo, fail to fetch one, and use latest instead. 
Latest in Cairo could already be on v2.20.x; while our intention is to check against v2.19.4 crates list.

.github/workflows/_check-cairo-lang-versions.yml
```diff
@@ -0,0 +1,94 @@
...
+          fi
+
+          curl -sSfL "https://raw.githubusercontent.com/starkware-libs/cairo/$tag/scripts/crates_list.sh" -o crates_list.sh
+
+      - name: Parse cairo-lang-* crate names out of it
+        id: parse
+        run: |
+          names=$(grep -oE 'cairo-lang-[a-zA-Z0-9_-]+' crates_list.sh | sort -u)
```
maciektr: We should take CRATES list contents as is, without assuming it starts with cairo-lang-*. 
This is true now, but only by convention, and could change in future.

.github/workflows/_check-cairo-lang-versions.yml
maciektr, line 1: Extension: we can also validate crates that do not come from Cairo repo, but should also be pinned to registry releases, like cairo-lang-language-server, etc.

.github/workflows/_check-cairo-lang-versions.yml
```diff
@@ -0,0 +1,94 @@
...
+          NAMES_JSON: ${{ steps.parse.outputs.names_json }}
+        run: |
+          # The resolved dependency graph exactly as frozen in Cargo.lock (--locked
+          # errors out instead of silently re-resolving if the lock is stale).
+          metadata=$(cargo metadata --locked --format-version=1)
+
+          # Names of local/workspace crates (path dependencies have no `source`).
+          # These are excluded below since they can share a name with an externally
+          # published crate without it meaning anything (e.g. our own cairo-lang-macro).
+          local_names=$(jq -c '[.packages[] | select(.source == null) | .name]' <<< "$metadata")
```
maciektr: Why do we need to exclude those? There should never be a compiler crate called cairo-lang-macro - it would be a problem as well.

maciektr (changes requested):

.github/workflows/_check-cairo-lang-versions.yml
```diff
@@ -0,0 +1,84 @@
...
+        id: parse
+        run: |
+          # Only eval the CRATES array definition itself, not the rest of the
+          # script, which assumes it's running inside a Cairo checkout.
+          array_def=$(sed -n '/^CRATES=(/,/^)/p' crates_list.sh)
+          eval "$array_def"
+
+          # Registry-pinned crates that aren't part of the Cairo repo's own
+          # release list, but should still resolve to a single registry version.
+          extra_names=(cairo-lang-language-server)
```
maciektr: Should we also add cairo-lint?

maciektr (approved):

PR #3263, opened 2026-09-17
Release 2.19.5

hakiers:


wawel37 (changes requested):
Tbh i think we shouldn't be merging that into the release/2.19.4, as this one is a separate version and we dont want to override anything related to 2.19.4

hakiers:
@wawel37 You are right. I created the PR because I wanted to check if all tests passed and I forgot to close it.

